PR #70, opened 2020-11-18
Pooling

billbrod:
A large commit that mixes too many things:

Big chunk of this relates to the pooling used in my pooled ventral stream models:
 - I split out the `PoolingWindows` class into its own file, separate from the various calculation functions in `pooling.py`. 
 - I switched over to creating them all in `pytorch` (rather than `numpy` and converting), which made their creation much more efficient. 
 - Changed how I normalize the windows (to equalize their L1-norms across eccentricity), calculating the required factor analytically (one per windows) and storing it as an attribute so I can undo it when projecting.
 - Removed the ability to parallelize `PoolingWindows` across multiple GPUs, since I wasn't using that (it messed up metamer synthesis, resulting in noticeable boundary artifacts) and we weren't testing it
 - `PoolingWindows` can now create windows smaller than a pixel (though it raises a warning), in order to allow for a gradual transition to when the windows start (otherwise, when comparing across metamers with different scaling values, it's very obvious where the windows start)
 - Adds two new helper methods: `plot_window_checks`, which plots the L1 norm and sum of windows in a given angle slice as a function of eccentricity, in order to ensure that everything is working correctly, and `summarize_window_sizes()`, which returns a dictionary summarizing the size of the windows in pixels and degrees at several different locations.

`PooledVentralStream`, in addition to changes related to the above:
 - normalization works for all `PooledVentralStream` models, not just `PooledV1`
 - adds several methods for summarizing info about the model and its representation (what is its error at different eccentricity / quadrants, what's the window sizes, etc)
 - adds methods for converting back and forth between the 3d tensor output and the dictionary representation
 - updates docstrings throughout
 - removes ability for PooledV1 to have half-octave pyramid, since I wasn't using it (didn't make a difference in metamer synthesis), and residual highpass (which actively made metamers worse)
 - makes use of the fact that the steerable pyramid can have its scales specified (much more efficient than getting all coefficients and then restricting to the ones I need)

Adds sampling.py and Sampling_and_Aliasing notebook, small script and notebook to checks whether you're sampling finely enough to avoid aliasing and visualize it. Not sure it belongs in plenoptic.

New display code:
 - Adds Display notebook to start explaining how to use display code. This needs feedback
 - Synthesis has new plots that can be made, either separately or as part of plot_synthesis_status() and animate(): image_hist (histograms comparing pixel distributions of synthesized and base signal), representation_comparison and signal_comparison (scatter plots or 2d histograms comparing the values of the two representation or signals)
 - Adds ability to create the figure you want to use for animate() or plot_synthesis_status() externally and then pass it in (along with optional dict specifying which plots go in which axes)
 - All of the various sub-plots are now toggle-able for plot_synthesis_status() and animate()

Other Synthesis changes:
 - Adds support for stochastic weight averaging (SWA) on synthesized image
 - adds new attribute, pixel_change, which stores the max pixel change in synthesized_signal between iterations (before any clamping)

Other misc:
 - removes `cone()` function, since I'm not using it and I don't think it's helpful.
 - adds several potentially helpful functions: polar_radius, polar_angle (pure `pytorch` versions of functions that exist in `pyrtools`), rotate_image, crop_image_centered, flip_image
 - updates Metamer notebook to be compliant with changes
 - adds tests for new stuff
 - adds some new potential loss functions: `mse_and_penalize_range` (uses mean-squared error, penalizes values outside user-specified range), `l2_and_log_barrier` (l2-norm of difference, also adds log barrier for values outside user-specified range)
 - updates required pytorch version to 1.4, adds torchcontrib and seaborn as requirements
 - splits up display-related tests into own script, makes sure we're adequately testing animations


plenoptic/synthesize/Synthesis.py
```diff
@@ -479,11 +490,11 @@ def _finalize_stored_progress(self):
 
     @abc.abstractmethod
     def synthesize(self, seed=0, max_iter=100, learning_rate=1, scheduler=True, optimizer='Adam',
-                   clamper=RangeClamper((0, 1)), clamp_each_iter=True, store_progress=False,
+                   optimizer_kwargs={}, swa=False, swa_kwargs={}, clamper=RangeClamper((0, 1)),
```
lyndond, line 493: todo: We had meeting Jan8 talking about in the future super-classing this into a simpler form.

plenoptic/simulate/canonical_computations/pooling_windows.py
```diff
@@ -0,0 +1,1051 @@
+"""contains the PoolingWindows class
```
lyndond, line 1: is all of this truly new?

lyndond (changes requested):
Minor changes requested. There was quite a bit to go through so I didn't get a chance to dig super deep.

In Sampling_and_aliasing.ipynb, can you flesh out the intro to tell the user in what synthesis method this would be relevant? 
I would pull the metamer.ipynb advanced stuff out and make it its own file since there's a lot there.

plenoptic/simulate/canonical_computations/pooling_windows.py
billbrod: The `pooling_windows.py` file contains the `PoolingWindows` class -- that used to be in `pooling.py` but I split that file up because it was getting too big. So unfortunately it all got marked as "new".

plenoptic/synthesize/Synthesis.py
billbrod: Yup, issue #75. Let's leave this stuff for now, but it will be good to simplify all this.

plenoptic/tools/optim.py
```diff
@@ -99,6 +99,45 @@ def penalize_range(synth_img, allowed_range=(0, 1), **kwargs):
     return torch.sum(torch.cat([below_min, above_max]))
 
 
+def log_barrier(synth_img, allowed_range=(0, 1), epsilon=1., **kwargs):
```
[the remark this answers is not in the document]
billbrod: Good idea, removed

plenoptic/simulate/canonical_computations/sampling.py
```diff
@@ -0,0 +1,258 @@
...
+from .pooling import gaussian
+from ...tools.data import to_numpy
+
+
+def check_sampling(val_sampling=.5, pix_sampling=None, func=gaussian, x=torch.linspace(-5, 5, 101),
+                   **func_kwargs):
+    r"""check how sampling relates to interpolation quality
+
+    Given a function, a domain, and how to sample that domain, this
+    function will use linear algebra (``np.linalg.lstsq``) to determine
```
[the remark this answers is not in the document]
billbrod: Psh but this is special linear algebra

plenoptic/tools/signal.py
```diff
@@ -3,6 +3,135 @@
 from pyrtools.pyramids.steer import steer_to_harmonics_mtx
 
 
+def rotate_image(img, angle, X=None, Y=None):
```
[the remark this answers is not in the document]
billbrod: Ah that's a very good point! I'll clone the input image

PR #73, opened 2020-12-28
Multiple eigendistortions: block power iteration and randomized svd 

lyndond:
Changes to algorithms:
1. Removed Lanczos because it is unstable and too much hassle to stabilize.
2. Generalized the power method so now it does top and bottom k eigendistortions
3. Added random svd which approximates the range of the Fisher Info Matrix; also includes approximation error estimate

Changes to tests: 
1. Added tests to accommodate new algos
2. Might add stronger numerical accuracy tests for the new algos too but need to brainstorm some effective tests

Low priority things that need to be changed in the future:

1.  tutorial to include muti-distortion methods
2. more stringent tests


plenoptic/synthesize/eigendistortion.py
```diff
@@ -1,13 +1,14 @@
...
 from .autodiff import jacobian, vector_jacobian_product, jacobian_vector_product
 import numpy as np
 import pyrtools as pt
 import warnings
 from .synthesis import Synthesis
 from tqdm import tqdm
 from matplotlib import pyplot as plt
-import plenoptic as po
+from typing import Tuple
+
```
billbrod, line 11: Neither of these imports are used, so they should be removed.

plenoptic/synthesize/eigendistortion.py
```diff
@@ -45,6 +46,9 @@ def fisher_info_matrix_eigenvalue(y, x, v):
     """
     Fv = fisher_info_matrix_vector_product(y, x, v)
     lmbda = Fv.T @ v
+    if v.shape[1] > 1:
+        lmbda = torch.diag(lmbda)
+
```
billbrod, line 51: Can we get a brief explanation for why this is done? (one sentence comment woudl suffice)

plenoptic/synthesize/eigendistortion.py
```diff
@@ -161,30 +165,24 @@ def to(self, *args, attrs=[], **kwargs):
...
-            last eigendistortions, with maximum number of iterations dictated by n_steps. 'lanczos' uses the Arnoldi
-            iteration algorithm to estimate the _entire_ eigenspectrum and thus more than just two eigendistortions,
-            as opposed to the power method. Note: 'lanczos' method is experimental and may be numerically unstable.
-            We recommend using the power method.
-        e_vecs: iterable, optional
-            Integer list of which eigenvectors to return for ``method='lanczos'``.
+            last eigendistortions, with maximum number of iterations dictated by n_steps. 'svd' uses randomized SVD
+            to approximate the top k eigendistortions and their corresponding eigenvalues.
+        k: int
+            How many vectors to return using block power method or svd.
```
billbrod, line 179: Parameters out of order in docstring. I think `method` should be first in the call signature, no? Since that's the most important one?

plenoptic/synthesize/eigendistortion.py
```diff
@@ -161,30 +165,24 @@ def to(self, *args, attrs=[], **kwargs):
...
         tol: float, optional
             Tolerance for error criterion in power iteration.
-        n_steps: int, optional
-            Total steps to run for ``method='power'`` or ``method='lanczos'`` in eigenvalue computation.
+        max_steps: int, optional
+            Total steps to run for ``method='power'`` in eigenvalue computation.
```
billbrod, line 183: Should probably be "maximum number of steps", to match the docstring from `_synthesize_power`

plenoptic/synthesize/eigendistortion.py
```diff
@@ -374,170 +386,51 @@ def _synthesize_power(self, l=0, tol=1e-10, n_steps=1000):
...
-            \begin{align}
-            A &\approx Q_kT_kQ_k^T\\
-              &= Q_kV\Lambda V^T Q_k^T
-            \end{align}
-
-        Thus, :math:`Q_kV` are our estimates of :math:`k` eigenvectors of :math:`A` and :math:`\text{diag}(\Lambda )` are
-        their associated eigenvalues.
-
+    def _synthesize_randomized_svd(self, k, p=0, q=0):
+        r"""  Synthesize eigendistortions using randomized truncated SVD.
```
billbrod, line 390: Right now, user cannot set `q` or `p`, and they're hard coded to `p=5`, `q=2`. Are these parameters that users would ever want to change? E.g., are they hyperparameters that will need to be tweaked to find a good solution?

plenoptic/synthesize/eigendistortion.py
```diff
@@ -374,170 +386,51 @@ def _synthesize_power(self, l=0, tol=1e-10, n_steps=1000):
...
-            torch.Size(len(e_vecs),) if e_vecs is not None. eigenvalues corresponding to the eigenvectors at that index.
-        eig_vecs: toch.Tensor
-            Tensor of n_steps eigenvectors with torch.Size(n, len(e_vecs)) . If ``return_evecs=False``,
-            then returned  eig_vecs is an empty tensor.
-        eig_vecs_ind: torch.Tensor
-            Indices of each returned eigenvector
+        S: torch.Tensor
+            Eigenvalues, Size((n, )).
+        V: torch.Tensor
+            Eigendistortions, Size((n, k)).
```
billbrod, line 405: Also returns `error_approx`, should be documented

tests/test_eigendistortion.py
```diff
@@ -104,69 +103,44 @@ def test_method_power(self, color):
...
+        ed.synthesize(k=k, method='power', max_steps=10)
 
-        assert (e_vals[0]-ed.synthesized_eigenvalues[0]) < 1e-2
+        assert ed.synthesized_signal.shape == (k*2, n_chans, n, n)
+        assert ed.synthesized_eigenindex.allclose(torch.cat((torch.arange(k), torch.arange(n**2 - k, n**2))))
+        assert len(ed.synthesized_eigenvalues) == 2*k
 
-    def test_method_equivalence(self):
+    def test_method_svd(self):
+        n, k = 30, 10
```
billbrod, line 117: Should we test finding different number of eigenvectors (different values of `k`)? Is it guaranteed that the first eigenvector will always be the same (for different values of `k`) or should that be explicitly checked?

tests/test_eigendistortion.py
```diff
@@ -104,69 +103,44 @@ def test_method_power(self, color):
...
-        fig_min = e_pow.plot_loss(-1)
-        fig_min.show()
+        assert e_pow.synthesized_eigenvalues[-1].isclose(e_jac.synthesized_eigenvalues[-1], atol=1e-2)
+        assert e_svd.synthesized_eigenvalues[0].isclose(e_jac.synthesized_eigenvalues[0], atol=1e-3)
 
     @pytest.mark.parametrize("color", [False, True])
     def test_display(self, color):
         e_pow = get_synthesis_object(im_dim=SMALL_DIM, color=color)
-        e_pow.synthesize(method='power', n_steps=50, store_progress=True)
+        e_pow.synthesize(method='power', max_steps=50, store_progress=True)
```
billbrod, line 143: I think we want some other tests here:
 - different methods
 - different number of eigenvectors
 - edge cases: `max_steps=0`, `store_progress=False` (ensure those fail in the expected ways)

billbrod (changes requested):
Looks like no meaningful changes to the notebook right? Some small changes requested to other files.

plenoptic/synthesize/eigendistortion.py
billbrod: since the top got cut off, the "neither" refers to `from typing import Tuple` and `from torch import Tensor`

lyndond: done

plenoptic/synthesize/eigendistortion.py
lyndond: done.

plenoptic/synthesize/eigendistortion.py
lyndond: done.

plenoptic/synthesize/eigendistortion.py
lyndond: done.

plenoptic/synthesize/eigendistortion.py
lyndond: done.

tests/test_eigendistortion.py
lyndond: I think diff values of ``k`` isn't necessary b/c the top should be equiv (within numerical precision) each time.

tests/test_eigendistortion.py
lyndond: done.

plenoptic/synthesize/eigendistortion.py
lyndond: These should work 99% of the time, but you're right. I've now allowed the user to change it and just set ``p=5`` ``q=2`` as defaults.

lyndond:
- Made all changes Billy requested. 
- Fleshed out documentation and included PEP typehinting
- Finished Eigendistortion demo reproducing Berardino et al 2017 eigendistortion
- Enabled GPU FrontEnd model
- Optimized eigendistortion GPU synthesis by caching dummy vector for jacobian-vector product trick
- **Reverted** to old non-Synthesis() class inheritance. Everything is now independent of any super class. 
- Added simple ``plot_distorted_image()`` method
- Added a couple parametrized test cases

examples/Demo_Eigendistortion.ipynb
```diff
@@ -0,0 +1,491 @@
+{
```
lyndond, line 1: <p>the minimal dist shouldn't look like this</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LabForComputationalVision/plenoptic/pull/73/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

billbrod (changes requested):
Mostly in `frontend.py`, because I haven't looked at that for a while.

examples/Demo_Eigendistortion.ipynb
billbrod, line 1: <p><code>po.load_images</code>  will do what you have <code>to_tensor</code>  do. The following returns the tensor (you'll still have to chop off some to make it square:</p><pre class="ql-syntax" spellcheck="false">image = po.load_images('../data/Parrot.png')
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LabForComputationalVision/plenoptic/pull/73/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

billbrod, line 1: <p>why not <code>po.imshow</code>?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LabForComputationalVision/plenoptic/pull/73/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

billbrod, line 1: <p>why does <code>k&gt;1</code>  converge faster than <code>k=1</code>?  that's unintuitive for me</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LabForComputationalVision/plenoptic/pull/73/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

billbrod, line 1: <p>maybe improve the alpha values? the distortions are pretty hard to see.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LabForComputationalVision/plenoptic/pull/73/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

plenoptic/synthesize/eigendistortion.py
```diff
@@ -98,100 +104,83 @@ class Eigendistortion(Synthesis):
...
     @classmethod
     def load(file_path, model_constructor=None, map_location='cpu', **state_dict_kwargs):
         # TODO attribute name *****
-        super().load(file_path, 'model', model_constructor, map_location, **state_dict_kwargs)
+        pass
 
     def save(self, file_path, save_model_reduced=False, attrs=['model'], model_attr_names=['model']):
         # TODO
-        super().save(file_path, save_model_reduced, attrs, model_attr_names)
+        pass
```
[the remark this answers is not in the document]
billbrod: same for the `load` function above this

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +47,68 @@ class Front_End(nn.Module):
...
-        self.luminance_scals = self.luminance_scals.view((1, 2, 1, 1))
+        if not requires_grad:  # turn off gradient
+            [p.requires_grad_(False) for p in self.parameters()]
 
-        self.contrast_scals = torch.tensor((scals[1], scals[3]), dtype=torch.float32)
-        self.contrast_scals = self.luminance_scals.view((1, 2, 1, 1))
+    def _load_pretrained(self, dtype):
+        """Load FrontEnd model weights used from Berardino et al (2017)"""
+        filts = np.load(dirname + '/front_end_filters.npy')
+        scals = np.load(dirname + '/front_end_scalars.npy')
```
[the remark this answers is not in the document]
lyndond: revamped. dirname no longer an explicit var, nor is user-facing

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +47,68 @@ class Front_End(nn.Module):
...
-            self.disk = make_disk(image_size)
+        if self._disk is not None and self._disk.shape == x.shape[-2:]:  # uses cached disk_mask if size matches
+            x = self._disk * x
+
+        elif ((self._disk is not None and self._disk.shape != x.shape[-2:])  # create new disk if disk size mismatch
+                or (self._disk is None and self.disk_mask)):  # or if disk does not yet exist
 
-            x = self.disk * x
+            self._disk = make_disk(x.shape[-1]).to(x.device)
+            x = self._disk * x
```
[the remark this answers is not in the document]
lyndond: done

plenoptic/synthesize/autodiff.py
```diff
@@ -1,27 +1,28 @@
...
     """Explicitly compute the full Jacobian matrix.
     N.B. This is only recommended for small input sizes (e.g. <100x100 image)
 
     Parameters
     ----------
-    y: torch.Tensor
+    y: Tensor
         model output with gradient attached
-    x: torch.Tensor with gradient function
+    x: Tensor with gradient function
```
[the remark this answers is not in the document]
lyndond: works in sphinx

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +47,68 @@ class Front_End(nn.Module):
...
+        # normalization scale factors
+        self.luminance_scale.data = torch.tensor((scals[0], scals[2]), dtype=dtype).view((1, 2, 1, 1))
+        self.contrast_scale.data = torch.tensor((scals[1], scals[3]), dtype=dtype).view((1, 2, 1, 1))
 
     def luminance_normalization(self, x):
-
-        s = self.luminance_scals
-        return torch.div(self.linear(x), (1 + s * self.luminance(x)))
+        s = self.luminance_scale
+        return torch.div(self.linear_pad(x), (1 + s * self.luminance_pad(x)))
```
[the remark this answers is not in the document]
lyndond: done.

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +47,68 @@ class Front_End(nn.Module):
...
 
-        self.linear = transforms.Compose([pad, linear])
-        self.luminance = transforms.Compose([pad, luminance])
-        self.contrast = transforms.Compose([pad, contrast])
-        self.softplus = nn.Softplus()
-
-        self.disk_mask = disk_mask
+        # normalization scale factors
+        self.luminance_scale.data = torch.tensor((scals[0], scals[2]), dtype=dtype).view((1, 2, 1, 1))
+        self.contrast_scale.data = torch.tensor((scals[1], scals[3]), dtype=dtype).view((1, 2, 1, 1))
```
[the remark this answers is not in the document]
lyndond: I think it's because the 1-channel input is split into two channels that each have a linear+nonlinear steps that happen in parallel. So the ordering was wrt the steps within each channel.

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +47,68 @@ class Front_End(nn.Module):
...
         x = self.contrast_normalization(x)
         x = self.softplus(x)
 
-        if self.disk_mask:
-            self.disk = make_disk(image_size)
+        if self._disk is not None and self._disk.shape == x.shape[-2:]:  # uses cached disk_mask if size matches
+            x = self._disk * x
+
+        elif ((self._disk is not None and self._disk.shape != x.shape[-2:])  # create new disk if disk size mismatch
+                or (self._disk is None and self.disk_mask)):  # or if disk does not yet exist
```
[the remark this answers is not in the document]
lyndond: it's ugly either way

plenoptic/synthesize/eigendistortion.py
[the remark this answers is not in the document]
lyndond: they now raise ``NotImplementedError``

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +47,68 @@ class Front_End(nn.Module):
...
+        self.linear = nn.Conv2d(in_channels=1, out_channels=2, kernel_size=31, bias=False)
+        self.luminance = nn.Conv2d(in_channels=1, out_channels=2, kernel_size=31, bias=False)
+        self.contrast = nn.Conv2d(in_channels=2, out_channels=2, kernel_size=31, groups=2, bias=False)
 
-        linear = nn.Conv2d(in_channels=1, out_channels=2, kernel_size=31, bias=False)
-        linear.weight = nn.Parameter(torch.tensor((filts[0], filts[3]),
-                                     dtype=torch.float32).unsqueeze(1))
+        # contrast and luminance normalization scaling
+        self.luminance_scale = nn.Parameter(torch.rand((1, 2, 1, 1), dtype=dtype))
+        self.contrast_scale = nn.Parameter(torch.rand((1, 2, 1, 1), dtype=dtype))
```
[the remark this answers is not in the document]
lyndond: changed to ``[name]_scalars``

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +47,68 @@ class Front_End(nn.Module):
         parasol on   fine time coarse space
         parasol off
     """
-    def __init__(self, disk_mask=False):
+    def __init__(self, disk_mask=False, pretrained=False, requires_grad=True, dtype=torch.float32):
```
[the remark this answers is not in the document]
lyndond: ``dtype is removed``. ``pretrained`` and ``requires_grad`` both remain to make things explicit.

plenoptic/synthesize/eigendistortion.py
```diff
@@ -98,100 +104,83 @@ class Eigendistortion(Synthesis):
...
-            Store loss after each iteration. Change in approximated eigenvalue is used as a proxy for loss and a
-            measure of convergence. Can be displayed with `plot_loss()`.
 
         Returns
         -------
-        eigendistortions: Union([torch.Tensor, List])
+        eigendistortions: Union([Tensor, List])
             Eigenvectors of the Fisher Information Matrix, ordered by eigenvalue. For Lanczos method, it is possible
             to not return any eigendistortions, in which case an empty list will be returned. This tensor points to
             the `synthesized_signal` attribute of the object. If not an empty list, then Size((num_distortions,
```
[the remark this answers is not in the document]
lyndond: That was a remnant of ``lanczos`` method. Ty for pointing that out. removed now.

examples/Demo_Eigendistortion.ipynb
lyndond: <p>When k&gt;1, each step has an additional 'orthogonalize against all other vectors' step. This carries some extra information about where eigenvectors should settle to and generally helps speed things up.</p><p> </p>

lyndond: <p>nice. done.</p><p> </p>

lyndond: <p>done.</p>

lyndond: <p>done. </p>

plenoptic/simulate/models/frontend.py
```diff
@@ -1,22 +1,18 @@
 import torch
 import torch.nn as nn
 from torchvision import transforms
-
-import numpy as np
 from ...tools.signal import make_disk
-
 import os
-dirname = os.path.dirname(__file__)
+import plenoptic as po
```
billbrod, line 6: I think there might be an issue with importing a package inside itself? (specific case I'm thinking of is if they have a version installed via pip and also a local version). I think the relative imports (like `from ...tools.signal import make_disk` above) are safer.

plenoptic/simulate/models/frontend.py
```diff
@@ -34,9 +30,7 @@ class Front_End(nn.Module):
 
     TODO
     -----
-    rerange input in [0,1]
-    make 12 parameters explicitely learnable
-    adapt to image size
+    Parameterize Conv2D as Difference of Gaussians -- the model should have 13 params total
```
billbrod, line 33: I'm not a big fan of TODOs in comments: I feel like we forget about them and they quickly get out of date, and only the author knows to check for them. Can we open an issue for it instead?

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +41,95 @@ class Front_End(nn.Module):
...
+        Parameters
+        ----------
+        disk_mask: bool
+            Apply circular Gaussian mask to center of image. The mask itself is square.
+        pretrained: bool
+            Load weights from Berardino et al. 2017. These are 31x31 convolutional filters.
+        requires_grad: bool
+            Whether or not model is trainable.
+        """
+        super().__init__()
```
billbrod, line 56: convolutional filters are always 31x31 now, right? even if `pretrained=False`. can we update docstring to make that clear?

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +41,95 @@ class Front_End(nn.Module):
...
-        contrast.weight = nn.Parameter(torch.tensor((filts[2], filts[5]),
-                                       dtype=torch.float32).unsqueeze(1))
-
-        pad = nn.ReflectionPad2d(filts.shape[-1]//2)
-
-        self.luminance_scals = torch.tensor((scals[0], scals[2]), dtype=torch.float32)
-        self.luminance_scals = self.luminance_scals.view((1, 2, 1, 1))
+    def __init__(self, disk_mask=False, pretrained=False, requires_grad=True):
+        """
+        Parameters
```
billbrod, line 47: Do parameters for initialization belong here or in the class docstring (the one on line 10 in this file)? I've been putting them in the class docstring

tests/test_models.py
```diff
@@ -277,7 +282,7 @@ def test_frontend_plot(self):
     def test_frontend_PoolingWindows(self):
         im = plt.imread(op.join(DATA_DIR, 'nuts.pgm'))
         im = torch.tensor(im, dtype=DTYPE, device=DEVICE).unsqueeze(0).unsqueeze(0)
-        frontend = po.simul.Front_End()
+        frontend = po.simul.FrontEnd(pretrained=True, requires_grad=False)
```
billbrod, line 285: Can we add some tests with `pretrained=False` and `requires_grad=True` (both together and separately)? Just to make sure all the expected functionality works (forward, synthesis, plotting) and that the filters change as they're being learned

tests/test_models.py
```diff
@@ -260,13 +260,18 @@ def test_rgc_save_load(self, tmp_path):
...
     def test_frontend(self):
         im = po.make_basic_stimuli()
-        frontend = po.simul.Front_End()
+        frontend = po.simul.FrontEnd(pretrained=True, requires_grad=False)
         frontend(im)
 
+    def test_frontend_display_filters(self):
+        frontend = po.simul.FrontEnd(pretrained=True, requires_grad=False)
+        frontend.display_filters()
+        plt.close('all')
```
billbrod, line 269: Same as below comment: can we test that this also works with `pretrained=False` and `requires_grad=True` (separately and together).

plenoptic/simulate/models/frontend.py
```diff
@@ -47,63 +41,95 @@ class Front_End(nn.Module):
...
+
+    def display_filters(self, zoom=5., **kwargs):
+        """Displays convolutional filters of FrontEnd model
+        Parameters
+        ----------
+        zoom: float
+            Magnification factor for po.imshow()
+        **kwargs:
+            Keyword args for po.imshow
+        Returns
```
billbrod, line 120: Filters can be complex-valued, right? Should we explicitly add the `plot_complex` keyword to this function?

plenoptic/synthesize/autodiff.py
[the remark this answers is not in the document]
billbrod: Oh cool, what does it look like?

plenoptic/synthesize/eigendistortion.py
```diff
@@ -203,82 +186,75 @@ def synthesize(self, method='power', e_vecs=None, tol=1e-10, n_steps=1000, seed=
             Index of each eigendistortion/eigenvalue. This points to the `synthesized_eigenindex` attribute of the
             object.
         """
-        self._set_seed(seed)
-        self.store_progress = store_progress
 
-        if e_vecs is None:
-            e_vecs = []
+        seed = np.random.randint(0, 2**32-1) if seed is None else seed
+        torch.manual_seed(seed)
```
billbrod, line 191: Three points here:
1. Nothing in this depends on numpy RNG does it? [PyTorch docs](https://pytorch.org/docs/stable/notes/randomness.html#random-number-generators-in-other-libraries) discuss setting numpy RNG as well. It says "However, some applications and libraries may use NumPy Random Generator objects, not the global RNG (https://numpy.org/doc/stable/reference/random/generator.html), and those will need to be seeded consistently as well.", but I think as long as the only randomness-related code we use is part of pytorch, we're good.

2. There is a `Synthesis._set_seed(seed)` function that sets both pytorch and numpy's RNG. Though it's `None` behavior is different.

3. What's the point of setting the seed using a random int (which itself will be based on the state of the RNG), rather than just not setting it? Does seed get stored in some way such that it's accessible to the user after the fact? (so they wouldn't have to set it when initializing, but could reproduce it afterwards)?

plenoptic/synthesize/eigendistortion.py
```diff
@@ -309,316 +285,175 @@ def _synthesize_exact(self):
...
+                             **kwargs) -> Figure:
+        r""" Displays specified eigendistortions alone, and added to the image.
 
         If image or eigendistortions have 3 channels, then it is assumed to be a color image and it is converted to
         grayscale. This is merely for display convenience and may change in the future.
 
         Parameters
         ----------
+        eigen_index: int
+
```
billbrod, line 431: `eigen_index` needs a description.

billbrod (changes requested):
Most important requested change is making sure the tests cover all possible initializations for `FrontEnd`

examples/Demo_Eigendistortion.ipynb
billbrod: <p>I feel like we should discuss that somewhere. Doesn't need to be here, but should be documented and linked to from here.</p>

billbrod: And if you don't want to insert it here or create the docs now, maybe create an issue for the eigendistortion documentation and add this point to it?

plenoptic/simulate/models/frontend.py
lyndond: Agreed. Turned into Issue #88

plenoptic/simulate/models/frontend.py
lyndond: done.

plenoptic/simulate/models/frontend.py
lyndond: Good catch. Same.

plenoptic/synthesize/eigendistortion.py
lyndond: 1. Nothing here dpeends on numpy RNG. 
2. Prev. implementation used ``Synthesis._set_seed()`` but was replaced with this in the rollback from subclassing Synthesis.
3. No real point. Changed to just not setting it now.

plenoptic/synthesize/eigendistortion.py
lyndond: done.

plenoptic/simulate/models/frontend.py
lyndond: done.

plenoptic/simulate/models/frontend.py
lyndond: Done -- changed to ``from ...tools.display import imshow``

tests/test_models.py
lyndond: done.

tests/test_models.py
lyndond: Added True/False combos for ``pretrained`` and ``requires_grad`` for all relevant tests, except for PoolingWindows FrontEnd tests where only ``requires_grad`` can be True/False; bc if the weights are untrained, the responses are nonsensical. 

This is bc right now, weights can be anything, despite the fact that the model should be a series of diff of Gaussians. This will be addressed in #88.

examples/Demo_Eigendistortion.ipynb
lyndond: This is now an issue #89.

plenoptic/synthesize/autodiff.py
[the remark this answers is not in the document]
lyndond: Typehints for `Tensor` get changed to ``torch.Tensor`` in argument string, but in docstring remain as `Tensor`.

image

plenoptic/synthesize/eigendistortion.py
```diff
@@ -182,13 +182,13 @@ def synthesize(self,
...
             `synthesized_eigenvalue` attribute of the object.
-        eigen_index: Listlike
+        eigen_index: Tensor
             Index of each eigendistortion/eigenvalue. This points to the `synthesized_eigenindex` attribute of the
             object.
         """
 
-        seed = np.random.randint(0, 2**32-1) if seed is None else seed
-        torch.manual_seed(seed)
+        if isinstance(seed, int):
```
billbrod, line 190: we should document that None is a possibility for seed (and maybe raise an exception if it's not seed or None, to catch weirdness like if they pass a float? or maybe try and cast it to an int

plenoptic/synthesize/eigendistortion.py
```diff
@@ -428,14 +429,18 @@ def plot_distorted_image(self,
...
         alpha: float, optional
             Amount by which to scale eigendistortion for image + (alpha * eigendistortion) for display.
         process_image: Callable
             A function to process the image+alpha*distortion before clamping between 0,1. E.g. multiplying by the
             stdev ImageNet then adding the mean of ImageNet to undo image preprocessing.
         ax: matplotlib.pyplot.axis, optional
             Axis handle on which to plot.
+        plot_complex: str, optional
+            Parameter for ``plenoptic.imshow`` determining how to handle complex values. Defaults to 'rectangular', i.e.
+            not complex. Can also be 'polar' or 'logpolar'; see method for details.
```
billbrod, line 443: rectangular isn't "not complex", it plots the real and imaginary as separate components. 

Also, instead of "see method for details", I think we should say "see that method's docstring" or something similar.

billbrod: That last point is really minor -- how is it done in other libraries?

plenoptic/synthesize/eigendistortion.py
```diff
@@ -454,6 +459,6 @@ def plot_distorted_image(self,
 
         img_processed = process_image(image + alpha * dist)
         to_plot = torch.clamp(img_processed, 0, 1)
-        fig = po.imshow(to_plot, ax=ax, **kwargs)
+        fig = po.imshow(to_plot, ax=ax, plot_complex=plot_complex, **kwargs)
```
billbrod, line 462: this will fail right? you're no longer importing `po`.

tests/test_models.py
```diff
@@ -258,31 +258,37 @@ def test_rgc_save_load(self, tmp_path):
...
         plt.close('all')
 
-    def test_frontend_PoolingWindows(self):
+    @pytest.mark.parametrize('requires_grad', [True, False])
+    def test_frontend_PoolingWindows(self, requires_grad):
         im = plt.imread(op.join(DATA_DIR, 'nuts.pgm'))
         im = torch.tensor(im, dtype=DTYPE, device=DEVICE).unsqueeze(0).unsqueeze(0)
-        frontend = po.simul.FrontEnd(pretrained=True, requires_grad=False)
+        frontend = po.simul.FrontEnd(pretrained=True, requires_grad=requires_grad)
         pw = po.simul.PoolingWindows(.5, (256, 256))
```
billbrod, line 292: I think we should still test both values for pretrained. they're nonsensical, but still possible. `pretrained=False` and `requires_grad=False` is also nonsensical (it doesn't make sense to me why someone would do that), but if we're allowing it, we should be testing it.

I'm not sure why it's *more* nonsensical with PoolingWindows than without

plenoptic/synthesize/eigendistortion.py
billbrod: Oh -- you are still importing `po` in this script. Should change that, like you did in frontend.py

billbrod (changes requested):

plenoptic/synthesize/eigendistortion.py
lyndond: done! ty.

tests/test_models.py
lyndond: Made a note in #88. This will be resolved in different PR where ``FrontEnd`` is finalized.

plenoptic/synthesize/eigendistortion.py
lyndond: done.

plenoptic/synthesize/eigendistortion.py
lyndond: done. i agree w/ ur second point since it's more explicit.

billbrod (approved):
Alright, looks good to me!

pehf:
Just from looking at the diff files, here are a few questions :
- how about making the demo notebook the main poster child and moving the other one to a '\getting-started' folder within the notebooks folder?
- is hard to see anything in the FrontEnd plot of the demo notebook, would it help if we did away with the mask?
- type hinting in `autodiff.py` and `eigendistortion.py` looks great, you are an inspiration
- power_method:
     - need to add documentation for argument k
     - when k is 1 we should not call a QR decomposition
     - when k is more, these are no longer a matrix_vector product, but matrix_matrix_products, this is confusing given the function names, so we should make the docstring very explicit
- randomized_svd:
    - can we add a good default p (margin)?
    - this method has fantastic performance in theory, I wonder what the comparison to power method is in our specific case, have you run tests, when would you recommend one over the other? randsvd should be much more scalable and fast from my understanding, but I also see that you have a couple of QR calls that might slow it down, although they might be quite necessary in practice for this to be sufficiently stable
- I'm not discussing changes to the front end model because this will be done in another PR

lyndond:
@pehf Thanks for your review.

Notebooks:
- I agree the quality of the frontend is currently suboptimal, I hope this will change once we finalize the model and padding isuses
- re: rearranging notebook placement and `/getting-started`, i think this is a great idea but we should hold off until we have more finalized versions before we start moving stuff around

Power method/orthogonal iteration:
- [x] documented ``k``
- [x] only calls QR when ``k>1`` 
- [x] more explicit that matmat vs matvec is being done

Randomized svd:
- [x] removed default args in ``_synthesize_randomized_svd``, since the defaults that actually matter are in the ``synthesize()`` method which then calls private method ``_synthesize_randomized_svd``.
- [x] removed defaults in private method ``_synthesize_power()`` as well for the same reason as above.
- I've found randomized_svd to be very reliable -- fast and accurate. The downside is that in its current form it only computes eigenvecs of the column space, and it's not clear how one would compute the smallest eigenval/vec pairs. This is why power method remains default.

pehf (approved):
good to go

PR #78, opened 2021-01-28
removes Sampling_and_Aliasing from treebeard

billbrod:
Since that causes treebeard to fail.


lyndond (approved):
bueno

billbrod:
Actually, this is no longer necessary -- I made the same change in PR #79 by accident

PR #87, opened 2021-02-13
Update README.md

eerosim:
text adjustments


lyndond (approved):
looks a-ok to me

README.md
```diff
@@ -46,38 +46,30 @@ Here's a table summarizing this:
...
   
-(where for all of these, "identical (resp. different) representation",
-stands for small (resp. large) l2-distance in a model's representation space)
-
 # Status
 
-This project is currently in alpha, under heavy development. Not all features
-have been implemented, and there will be breaking changes.
+This project is currently under heavy development. Not all features
+have been implemented, and there will be significant changes.
```
billbrod, line 72: I'd like to keep the phrase "breaking changes", which is specific that we're not guaranteeing code written with plenoptic now might not work in several months. "significant changes" is also true, but feels less specific.

README.md
```diff
@@ -46,38 +46,30 @@ Here's a table summarizing this:
...
+to examples that make use of them):
+
+- [Metamers](http://www.cns.nyu.edu/~lcv/texture/):
+  given a model and a reference image, stochastically generate a new image whose
+  model representation is identical to that of the reference image.
+- [Eigendistortions](https://www.cns.nyu.edu/~lcv/eigendistortions/):
+  given a model and a reference image, compute the image perturbation that produces
+  the smallest and largest changes in the model response space.  These correspond to the
+  minimal/maximal eigenvectors of the Fisher Information matrix of the representation (for deterministic models, 
+  the minimal/maximal singular vectors of the Jacobian).
```
billbrod, line 58: Any particular reason to order the synthesis methods this way? Just curious.

billbrod (changes requested):
I'd like to change "significant changes" back to "breaking changes" (or include both). Other than that, good to go.

README.md
lyndond: agree -- changed

billbrod (approved):

PR #93, opened 2021-02-18
Display fixes

billbrod:
Changes here all have to do with display issues

- `load_images`, `update_plot` and `Synthesis`'s various plotting/animating functionality now handle RGB(A) images correctly
- went through and replaced `pt.imshow` with `po.imshow` just about everywhere. Only places `pt.imshow` remains are a handful of places that were actually dealing with numpy arrays, eigendistortions.py (since that's fixed in #73), and the 01 and 04 notebooks, since they looked like they'll need more work.
- `Synthesis.__init__` raises an exception if `base_signal` is not 4d
- `update_plot`, `Synthesis.animate`, `Synthesis.plot_synthesis_status` raise exceptions if the data is not 3 or 4d, since I don't know how they should behave then
- `seaborn` dependency removed, replaced `sns.distplot` with `plt.hist`
- `update_plot`'s docstring to updated to be a little more detailed about what it expects axes and data to look like.


plenoptic/synthesize/mad_competition.py
```diff
@@ -1697,6 +1699,11 @@ def _grab_value_for_comparison(self, value, iteration=None,
         ----------
         value : {'representation', 'signal'}
             Whether to compare the representations or signals
+        batch_idx : int, optional
+            Which index to take from the batch dimension (the first one)
```
lyndond, line 1703: Remove "the first one" or rewrite to something more explicit bc i thought it meant first one is default idx (and it is, since batch_idx=0)

plenoptic/synthesize/mad_competition.py
```diff
@@ -1697,6 +1699,11 @@ def _grab_value_for_comparison(self, value, iteration=None,
...
+        channel_idx : int or None, optional
+            Which index to take from the channel dimension (the second one). If
```
lyndond, line 1705: Same here; please make more explicit or just remove

plenoptic/synthesize/mad_competition.py
```diff
@@ -1775,8 +1789,9 @@ def plot_value_comparison(self, synthesis_target=None,
             Whether to compare the representations or signals
         batch_idx : int, optional
             Which index to take from the batch dimension (the first one)
-        channel_idx : int, optional
-            Which index to take from the channel dimension (the second one)
+        channel_idx : int or None, optional
+            Which index to take from the channel dimension (the second one). If
```
lyndond, line 1793: Same here

plenoptic/synthesize/mad_competition.py
```diff
@@ -1905,8 +1920,9 @@ def plot_synthesis_status(self, synthesis_target=None, batch_idx=0,
             ``self.synthesis_target``).
         batch_idx : int, optional
             Which index to take from the batch dimension (the first one)
-        channel_idx : int, optional
-            Which index to take from the channel dimension (the second one)
+        channel_idx : int or None, optional
+            Which index to take from the channel dimension (the second one). If
```
lyndond, line 1924: and here. Actually I see that you've written it this previously since line 1922 is written this way too. I suggest changing to something more explicit where it's currently written like "the nth one"

plenoptic/synthesize/mad_competition.py
```diff
@@ -2003,8 +2019,9 @@ def animate(self, synthesis_target=None, batch_idx=0, channel_idx=0,
         ----------
         batch_idx : int, optional
             Which index to take from the batch dimension (the first one)
-        channel_idx : int, optional
-            Which index to take from the channel dimension (the second one)
+        channel_idx : int or None, optional
+            Which index to take from the channel dimension (the second one). If
```
lyndond, line 2023: here too

plenoptic/synthesize/metamer.py
```diff
@@ -472,8 +472,9 @@ def plot_value_comparison(self, value='representation', batch_idx=0,
             Whether to compare the representations or signals
         batch_idx : int, optional
             Which index to take from the batch dimension (the first one)
-        channel_idx : int, optional
-            Which index to take from the channel dimension (the second one)
+        channel_idx : int or None, optional
+            Which index to take from the channel dimension (the second one). If
```
lyndond, line 476: Here too

plenoptic/synthesize/synthesis.py
```diff
@@ -58,6 +57,9 @@ def __init__(self, base_signal, model, loss_function, model_kwargs={}, loss_func
 
         if not isinstance(base_signal, torch.Tensor):
             base_signal = torch.tensor(base_signal, dtype=torch.float32)
+        if base_signal.ndim != 4:
```
lyndond, line 60: nice

plenoptic/synthesize/synthesis.py
```diff
@@ -1439,33 +1451,56 @@ def plot_image_hist(self, batch_idx=0, channel_idx=0, iteration=None, figsize=(5
...
-            passed to sns.distplot
+            passed to plt.hist
 
         Returns
         -------
         fig : matplotlib.pyplot.Figure
             The figure containing this plot
 
         """
+        def _freedman_diaconis_bins(a):
```
lyndond, line 1462: nice

plenoptic/synthesize/synthesis.py
```diff
@@ -1717,7 +1762,7 @@ def _setup_synthesis_fig(self, fig, axes_idx, figsize,
             axes = fig.axes
         return fig, axes, axes_idx
 
-    def plot_synthesis_status(self, batch_idx=0, channel_idx=0, iteration=None,
+    def plot_synthesis_status(self, batch_idx=0, channel_idx=None, iteration=None,
```
lyndond, line 1765: General comment: I wonder if when we do our refactoring of ``synthesis`` we can pull some of these plotting tools into like a ``synthesis_diagnostics.py`` or something.

lyndond (changes requested):
tiny comments. I'm thinking tmw we should sit down and discuss ways to simplify the overall plotting architecture.

plenoptic/synthesize/mad_competition.py
billbrod: done

plenoptic/synthesize/mad_competition.py
billbrod: done

plenoptic/synthesize/mad_competition.py
billbrod: done

plenoptic/synthesize/mad_competition.py
billbrod: I'm just going to go ahead and remove all of them.

plenoptic/synthesize/mad_competition.py
billbrod: done

plenoptic/synthesize/metamer.py
billbrod: done

plenoptic/synthesize/synthesis.py
billbrod: ty

plenoptic/synthesize/synthesis.py
billbrod: ty

plenoptic/synthesize/synthesis.py
billbrod: Yeah, agreed -- see #92 for this. That's what Kate and I had discussed so far, happy to hear any more ideas.

plenoptic/simulate/canonical_computations/pooling_windows.py
```diff
@@ -15,6 +15,7 @@
 import matplotlib.pyplot as plt
 import os.path as op
 from ...tools.data import to_numpy
+from ...tools.display import imshow
```
kbonnen, line 18: There are a fair number of relative imports in our package now.  I have found their presence to be somewhat confusing on a number of occasions.  PEP8 seems to discourage relative imports to some degree (https://www.python.org/dev/peps/pep-0008/#imports).  Thoughts?

plenoptic/synthesize/synthesis.py
kbonnen: The display code needs to be refactored along with the Synthesis refactor 👍

tests/test_display.py
```diff
@@ -413,6 +423,44 @@ def test_animshow(self, as_rgb, channel_idx, batch_idx, is_complex, mini_vid):
...
+
+        class DumbModel(po.simul.PooledRGC):
+            def forward(self, *args, **kwargs):
+                output = super().forward(*args, **kwargs)
+                return output.reshape(output.numel())
+        model = DumbModel(.5, im.shape[2:]).to(DEVICE)
+        met = po.synth.Metamer(im, model)
+        met.synthesize(max_iter=3, store_progress=True)
+        # Synthesis plot_synthesis_status and animate expect 3 or 4d data --
+        # this checks that it fails (because im.squeeze() is 2d) and raises the
```
kbonnen, line 449: check that "it" fails?  Can you make it slightly more clear what these tests are testing for?

kbonnen (changes requested):
Just a few minor comments.  

The only change I want prior to commit is a slightly clearer description of the two tests that were added to test_display.py

plenoptic/simulate/canonical_computations/pooling_windows.py
billbrod: Oh woah, my impression was that absolute imports from within a package were a bad idea, because it is possible to have multiple versions of the same package on your path (for example, if you have a local version of a package and then are in a virtual environment where you've installed it via `pip`), and I thought the relative import would make sure it's grabbing the version from within your package, rather than some other version on the path. We should look into this more.

billbrod: In fact I just told Lyndon to change an absolute to relative import in #73

kbonnen: I think I might be using the wrong language here... basically I'm concerned by the fact that that we import a bunch of variables and method names and then when we call them, it's not immediately obvious that they come from outside the current folder.  So instead of calling clean_up_axes(), could we call tools.display.clean_up_axes()

tests/test_display.py
billbrod: Sure thing.

plenoptic/simulate/canonical_computations/pooling_windows.py
kbonnen: I absolutely agree with the change you recommended in #73 but I think I'd like to ask that we still use the full path to call the methods.

billbrod: Oh gotcha. That makes a lot of sense.

So the top would say somethign like `import ...tools` instead

billbrod:
Addresses #72

lyndond (approved):
looks good to me. looking forward to seeing how things will change after the major upcoming refactor.

kbonnen (approved):

PR #94, opened 2021-02-19
Avoid redundant typehinting for docs w/ sphinx

lyndond:
Avoid redundancy in typehinting function docstring & arguments. Relevant issue: #77 

Example: Note how the docstring does not have type hints, but arguments do. And the output sphinx doc (pic below) displays types in the docstring.

```
class LN(nn.Module):
    """Linear-Nonlinear model, applies a difference of Gaussians filter followed by a (optional) nonlinearity.

    Parameters
    ----------
    center:
        ['on', 'off']
    activation:
        Default nonlinearity is a softplus activation function.
    """
    def __init__(self, kernel_size,
                 center: str = 'on',
                 activation: Callable[[torch.Tensor], torch.Tensor] = F.softplus):
```                 

image


billbrod (approved):
I can't believe this is all that's necessary! We're going to eventually want to put in explicit checks about how the docstrings are formatted (see #77 ), but that doesn't belong here

pehf (approved):
yay!

PR #95, opened 2021-02-20
Model update

lyndond:
Unify trainable models 

- Got rid of FrontEnd and now have frontend module with a all models used in Berardino et al 2017, each trainable
- moved Linear and LinearNonLinear models into single linear_nonlinear module
-  applying image masks using `make_disk` now allows for rectangular images instead of assuming square images
- added proper same_padding method  

Addresses: #23 #88


plenoptic/simulate/canonical_computations/linear_nonlinear.py
```diff
@@ -1,25 +1,102 @@
...
+    """
+
+    def __init__(self, n_channels=3, kernel_size=(3, 3), default_filters=True):
         super().__init__()
 
+        self.conv = nn.Conv2d(
+            1,
+            n_channels,
+            kernel_size,
+            padding=[int((k - 1) / 2) for k in kernel_size],
```
nikparth, line 75: do we care about exactly how this is padded as default?

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,421 @@
...
+
 from ...tools.display import imshow
+from ...tools.signal import make_disk
+
+from collections import OrderedDict
+
+__all__ = ["Gaussian", "CenterSurround", "LN", "LG", "LGG", "OnOff"]
+
 
+def get_pad(kernel_size: Union[int, Tuple[int, int]]) -> Tuple[int, int, int, int]:
```
nikparth, line 16: Are you sure this will work for both even and odd-sized kernels? I guess this again depends on what padding we want as default, but depending on if it's even or odd-size I think that would require different paddings for example if you wanted to keep the same output size.

nikparth (changes requested):
Looks good to me overall had a few questions about padding- not sure if we discussed this at some point so just want to confirm what we're doing there before final approval.

plenoptic/simulate/canonical_computations/linear_nonlinear.py
lyndond: Added new same_padding method that u helped me with

plenoptic/simulate/models/frontend.py
lyndond: Added new padding method in `tools/conv.py` that u helped me with

nikparth (approved):

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
+        )
+        self.activation = activation
 
-        """
+    def forward(self, x: Tensor) -> Tensor:
+        y = self.activation(self.center_surround(x))
+        return y
+
+
+class LG(nn.Module):
```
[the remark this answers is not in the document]
billbrod: `LinearGainControl`?

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
-        self.luminance_scalars = nn.Parameter(torch.rand((1, 2, 1, 1)))
-        self.contrast_scalars = nn.Parameter(torch.rand((1, 2, 1, 1)))
 
-        # pad all transforms for convolution
-        pad = nn.ReflectionPad2d(self.linear.weight.shape[-1]//2)
-        self.linear_pad = transforms.Compose([pad, self.linear])
-        self.luminance_pad = transforms.Compose([pad, self.luminance])
-        self.contrast_pad = transforms.Compose([pad, self.contrast])
-        self.softplus = nn.Softplus()
+class LGG(nn.Module):
```
billbrod, line 269: More descriptive name?

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
-import os
+import torch.nn.functional as F
+from torch import Tensor
+
+from ...tools.conv import same_padding
 from ...tools.display import imshow
+from ...tools.signal import make_disk
+from collections import OrderedDict
 
+__all__ = ["Gaussian", "CenterSurround", "LN", "LG", "LGG", "OnOff"]
```
billbrod, line 13: Are these models implementations of the ones Alex used? Then we should cite the proper paper (or his dissertation)

plenoptic/simulate/models/frontend.py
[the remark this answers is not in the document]
billbrod: Also, should probably inherit `CenterSurround` too

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
         super().__init__()
+        self.center_surround = CenterSurround(
+            kernel_size,
+            center,
+            width_ratio_limit,
+            amplitude_ratio,
+            pad_mode=pad_mode,
+        )
+        self.luminance = Gaussian(kernel_size=kernel_size)
+        self.luminance_scalar = nn.Parameter(torch.rand(1) * 10)
```
billbrod, line 258: I'm a bit confused by this. Does this initialization lead to good behavior? I thought this number was supposed to be relatively small

plenoptic/simulate/models/frontend.py
billbrod: And inherit `CenterSurround`.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
-        self.softplus = nn.Softplus()
+class LGG(nn.Module):
+    """ Linear center-surround followed by luminance and contrast gain control,
+    and activation function.
 
-        self.disk_mask = disk_mask
-        self._disk = None  # cached disk to apply to image
+    Parameters
+    ----------
+    See `CenterSurround` class for full Parameter docstring.
```
billbrod, line 275: Seems like a good opportunity to try out something from #25: have the parameters docstring be a constant that gets reused across classes.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
+class OnOff(nn.Module):
+    """Two-channel on-off and off-on center-surround model with local contrast and
+    luminance gain control.
 
-            self._disk = make_disk(x.shape[-1]).to(x.device)
-            x = self._disk * x
+    This model is called OnOff in Berardino et al 2017.
 
-        return x
+    Parameters
```
billbrod, line 339: Same thing about inheriting `CenterSurround` and repeating the docstrings

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
+        if self.apply_mask:
+            im_shape = x.shape[-2:]
+            if self._disk is None or self._disk.shape != im_shape:  # cache new mask
+                self._disk = make_disk(im_shape).to(x.device)
+
+            y = self._disk * y  # apply the mask
+
+        return y
+
+    def display_filters(self, zoom=5.0, **kwargs):
```
billbrod, line 402: why does only `OnOff` get a display function? Wouldn't it be useful for all of them?

billbrod (changes requested):
Lot of small questions scattered throughout. I do like the use of type-hinting, that definitely cleans up the code. Same with using `__all__`.

plenoptic/simulate/models/frontend.py
lyndond: The "true" learned values are 5ish and 15ish. These can be seen in the OnOff default statedict. 

These scalars multiply the filter responses and are not additive constants (e.g. semi-saturation constants), in case that might be what you're thinking of.

plenoptic/tools/conv.py
```diff
@@ -91,3 +95,33 @@ def blur_downsample(x,filtname = 'binom5', step=(2, 2)):
...
+
+
+def same_padding(
+        x: Tensor,
+        kernel_size: Union[int, Tuple[int, int]],
+        stride: Union[int, Tuple[int, int]] = (1, 1),
+        dilation: Union[int, Tuple[int, int]] = (1, 1),
+        pad_mode: str = "circular",
+) -> Tensor:
+    """Ciruclarly pad an image tensor and output image of same dims."""
```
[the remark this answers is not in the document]
lyndond: `same_padding` pads the image with values whose entries are dictated by `pad_mode`. E.g. those entries can be filled with zeros, or some other constant, or reflected values of the image, or in this case it will fill the padded entries with values assuming the image is periodic/circular.

plenoptic/tools/conv.py
```diff
@@ -91,3 +95,33 @@ def blur_downsample(x,filtname = 'binom5', step=(2, 2)):
 def upsample_blur(x, step=(2, 2)):
     f = pt.named_filter('binom5')
     return upsample_convolve(x, filt=np.outer(f, f), step=step)
+
+
+def get_same_padding(
```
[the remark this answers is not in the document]
lyndond: You are right. Added underscore.

tests/test_metamers.py
```diff
@@ -16,6 +16,7 @@ class TestMetamers(object):
     @pytest.mark.parametrize('model', ['LNL', 'nlpd'], indirect=True)
     @pytest.mark.parametrize('loss_func', [None, 'l2', 'range_penalty_w_beta'])
     @pytest.mark.parametrize('fail', [False, 'img', 'model', 'loss'])
+
```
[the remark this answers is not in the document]
lyndond: Not sure. Removed.

tests/test_display.py
```diff
@@ -440,7 +440,8 @@ def test_synthesis_plot_shape_fail(self, einstein_img):
         # Synthesis plot_synthesis_status and animate expect 3 or 4d data --
         # this checks that plot_synthesis_status() and animate() both fail with
         # 2d data and raise the proper exception
-        class TestModel(po.simul.Linear_Nonlinear):
+
```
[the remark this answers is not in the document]
lyndond: Must have been when I was resolving merge conflicts using browser github.com ui.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
+    deviation of the surround Gaussian is constrained to be at least `width_ratio_limit`
+    times that of the center Gaussian.
+
+    Parameters
+    ----------
+    center:
+        Dictates whether center is on or off. The surround always will be the opposite
+        of the center. Must be either ['on', 'off']; default is 'on' center.
+    kernel_size: Union[int, Tuple[int, int]], optional
+    width_ratio_limit:
```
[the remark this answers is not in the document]
lyndond: Having `center_std` as well as `surround_std` gives 2 DOF in the model. `width_ratio_limit` just adds constraint to `surround_std`.

If `width_ratio_limit=2`, then `center_std=1` and `surround_std=2` is valid, but `surround_std=2.01` is also valid.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
+        amplitude_ratio: float = 1.25,
+        center_std: float = 1.0,
+        surround_std: float = 4.0,
+        pad_mode: str = "circular",
+    ):
+        super().__init__()
+
+        assert center in ["on", "off"], "center must be 'on' or 'off'"
+        assert width_ratio_limit > 1.0, "stdev of surround must be greater than center"
+        assert amplitude_ratio >= 1.0, "amplitudes must"
```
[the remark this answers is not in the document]
lyndond: Nice catch. Fixed.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = same_padding(x, self.kernel_size, pad_mode=self.pad_mode)
+        self._clamp_surround_std()  # clip the surround stdev
+        filt = self.get_filter()
+        y = F.conv2d(x, filt, bias=None)
+        return y
+
+
+class LN(nn.Module):
```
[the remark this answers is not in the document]
lyndond: We have both because this entire file is based on Alex's models and includes all models he used in his paper.

One main difference is that LinearNonlinear has no nullspace has>1 output channel, whereas frontend.LN does have a null space and only one channel output.

I do agree it is confusing.

plenoptic/simulate/canonical_computations/linear_nonlinear.py
```diff
@@ -1,25 +1,127 @@
...
+    default_filters :
+        Initialize the first three filters to: low pass, high pass on / off.
+    """
+
+    def __init__(
+            self,
+            kernel_size: Union[int, Tuple[int, int]] = (3, 3),
+            pad_mode: str = "circular",
+            default_filters: bool = True,
+    ):
```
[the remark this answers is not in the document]
lyndond: Yes that's right.

plenoptic/simulate/canonical_computations/linear_nonlinear.py
```diff
@@ -1,25 +1,127 @@
...
+            f1 = torch.outer(g1, g1)
+            f1 = f1 / f1.sum()
+
+            g2 = torch.exp(-(xs ** 2) / (variance / 3))
+            f2 = torch.outer(g2, g2)
+            f2 = f2 / f2.sum() - f1
+            f2 = f2 - f2.sum()
+
+            self.conv.weight.data[0, 0] = nn.Parameter(f1)
+            self.conv.weight.data[1, 0] = nn.Parameter(f2)
```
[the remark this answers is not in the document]
lyndond: Removed LinearNonlinear and moved Linear to naive.py

plenoptic/simulate/canonical_computations/linear_nonlinear.py
```diff
@@ -1,25 +1,127 @@
...
+        size of the receptive fields (odd size supported), [3x3] by default
+    default_filters : bool, optional
+        Initialize the first three filters to: low pass, high pass on / off.
 
-    def forward(self, x):
+    Notes
+    -----
+    The idea of "Threshold Logic Unit" goes at least as far back as:
+    McCulloch, W.S. and Pitts, W., 1943. A logical calculus of the ideas
+    immanent in nervous activity. The bulletin of mathematical biophysics
```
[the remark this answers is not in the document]
lyndond: Removed LinearNonlinear altogether.

plenoptic/simulate/canonical_computations/linear_nonlinear.py
```diff
@@ -1,25 +1,127 @@
...
+            self.conv.weight.data[1, 0] = nn.Parameter(f2)
+
+    def forward(self, x: Tensor) -> Tensor:
+        y = same_padding(x, self.kernel_size, pad_mode=self.pad_mode)
+        h = self.conv(y)
+        return h
+
+
+class LinearNonlinear(nn.Module):
+    """Canonical functional model of early visual processing.
```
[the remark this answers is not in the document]
lyndond: n/a now that it's been removed.

plenoptic/simulate/canonical_computations/linear_nonlinear.py
```diff
@@ -1,25 +1,127 @@
...
+        super().__init__()
+        if isinstance(kernel_size, int):
+            kernel_size = (kernel_size, kernel_size)
+
+        self.kernel_size = kernel_size
+        self.pad_mode = pad_mode
+
+        self.conv = nn.Conv2d(1, n_channels, kernel_size, bias=False)
+
+        if default_filters and n_channels >= 3:
```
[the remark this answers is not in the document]
lyndond: n/a now that it's removed.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
+        std: float = 3.0,
+        pad_mode: str = "circular",
+    ):
+        super().__init__()
+        assert std > 0, "Gaussian standard deviation must be positive"
+        self.std = nn.Parameter(torch.tensor(std))
+        self.kernel_size = kernel_size
+        self.pad_mode = pad_mode
+
+    def get_filter(self):
```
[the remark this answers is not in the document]
lyndond: I turned it into a `@property`. More elegant this way, and the user can use them if they want to see the filter parameterized by the std in `Gaussian` (or two stds in `CenterSurround`).

plenoptic/simulate/models/frontend.py
lyndond: Done in docstring up top.

plenoptic/simulate/models/frontend.py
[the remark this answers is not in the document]
lyndond: Changed to more descriptive names. I'm not convinced inheriting is the way to go for these models though.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
+        Standard deviation of circular Gaussian for surround. Must be at least
+        ratio_limit times center_std.
+    pad_mode:
+        Padding for convolution, defaults to "circular".
+    """
+
+    def __init__(
+        self,
+        kernel_size: Union[int, Tuple[int, int]],
+        center: str = "on",
```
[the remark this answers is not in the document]
lyndond: Changed argument `center : str` where it can be 'on' or 'off' -->`on_center: bool = True` instead.

tests/test_models.py
```diff
@@ -1,8 +1,52 @@
...
+import torch
 from conftest import DEVICE
 
 
+@pytest.fixture()
+def image_input():
+    return torch.rand(1, 1, 100, 100)
+
+
+class TestFrontEnd:
```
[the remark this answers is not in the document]
lyndond: I don't think so. It's only the `frontend.OnOff` model that has a pretrained state_dict, and although I did my best to reverse engineer those params, I'd rather not test against them.

plenoptic/simulate/models/frontend.py
lyndond: Changed to `LuminanceContrastGainControl`

plenoptic/simulate/models/frontend.py
lyndond: u right. Added to the rest of the models.

plenoptic/simulate/canonical_computations/linear_nonlinear.py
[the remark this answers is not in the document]
lyndond: Also formatted docstrings according to ur link, ty.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,407 @@
...
 from ...tools.display import imshow
+from ...tools.signal import make_disk
+from collections import OrderedDict
 
+__all__ = ["Gaussian", "CenterSurround", "LN", "LG", "LGG", "OnOff"]
 
-class FrontEnd(nn.Module):
-    """Luminance and contrast gain control, modeling retina and LGN
 
+def circular_gaussian(
```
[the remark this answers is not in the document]
lyndond: I deleted `metric.perceptual_distance.create_window`, moved and reformatted `metric.perceptual_distance.gaussian` and my `models/frontend.circular_gaussian` to a new file `simulate/canonical_computations/filters.py`, renamed them to `gaussian1d` and `circular_gaussian2d`.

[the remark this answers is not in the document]
lyndond: After a bunch of test_display tests failed, I realize my `circular_gaussian2d()` returns a tensor w/ `Size([1, n_channels, h, w])` but yours returned as `Size([n_channels, 1, h, w])`. Is there a reason why you do that? 

I added a line in `_ssim_parts` to permute my filters to match what yours would have been to fix.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,513 @@
...
+    def forward(self, x: Tensor) -> Tensor:
+        x = same_padding(x, self.kernel_size, pad_mode=self.pad_mode)
+        self._clamp_surround_std()  # clip the surround stdev
+        y = F.conv2d(x, self.filt, bias=None)
+        return y
+
+
+class LinearNonlinear(nn.Module):
+    """Linear-Nonlinear model, applies a difference of Gaussians filter followed by an
+    activation function.
```
billbrod, line 159: is it worth pointing out explicitly that the difference between this and `CenterSurround` is just the activation function? and that `CenterSurround` is the "Linear" in this functions "LienarNonlinear"?

plenoptic/simulate/models/frontend.py
billbrod: Yup, that was exactly what I was thinking of.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,513 @@
...
+        """Clamps surround standard deviation to ratio_limit times center_std"""
+        return self.surround_std.clamp(
+            min=self.width_ratio_limit * float(self.center_std), max=None
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = same_padding(x, self.kernel_size, pad_mode=self.pad_mode)
+        self._clamp_surround_std()  # clip the surround stdev
+        y = F.conv2d(x, self.filt, bias=None)
+        return y
```
billbrod, line 154: why doesn't `CenterSurround` have a `display_filters` method? Also, it feels like it should be possible to write one `display_filters` function that these all use (the only thing that really differs among them is the title).

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,513 @@
...
+Model architectures in this file are found in [1]_, [2]_. `frontend.OnOff()` has
+optional pretrained filters that were reverse-engineered from a previously-trained model
+and should be used at your own discretion.
+
+References
+----------
+.. [1] A Berardino, J Ballé, V Laparra, EP Simoncelli, Eigen-distortions of hierarchical
+    representations, NeurIPS 2017; https://arxiv.org/abs/1710.02266
+.. [2] http://www.cns.nyu.edu/~lcv/eigendistortions/ModelsIQA.html
+"""
```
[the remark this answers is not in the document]
billbrod: Also, is it worth pointing out, in each model's docstring, which model they are in the paper? Under notes or references

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,513 @@
...
+        Activation function following linear convolution.
+    activation:
+        Activation function following linear and gain control operations.
+    apply_mask:
+        Whether or not to apply circular disk mask to image.
+
+    Notes
+    -----
+    These 12 parameters (standard deviations & scalar constants) were reverse-engineered
+    from model from [1]_, [2]_. Please use at your own discretion.
```
billbrod, line 450: Should point out that the "use at your own discretion" only applies if they use the pretrained weights, right?

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,513 @@
...
+        Shape of convolutional kernel.
+    width_ratio_limit:
+        Ratio of surround stdev over center stdev. Surround stdev will be clamped to
+        ratio_limit times center_std.
+    amplitude_ratio:
+        Ratio of center/surround amplitude. Applied before filter normalization.
+    pad_mode:
+        Padding for convolution, defaults to "circular".
+    activation:
+        Activation function following linear convolution.
```
billbrod, line 441: typo: `activation` shows up twice in the docstring and `pretrained` doesn't show up at all.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,513 @@
...
+    activation:
+        Activation function following linear and gain control operations.
+    apply_mask:
+        Whether or not to apply circular disk mask to image.
```
billbrod, line 445: more details about this mask? or point to something that gives more details?

plenoptic/simulate/models/naive.py
```diff
@@ -29,3 +37,54 @@ def forward(self, img):
...
+            variance = 3
+            xs = torch.linspace(-2, 2, kernel_size[0])
+            g1 = torch.exp(-(xs ** 2) / (2 * variance))
+            f1 = torch.outer(g1, g1)
+            f1 = f1 / f1.sum()
+
+            g2 = torch.exp(-(xs ** 2) / (variance / 3))
+            f2 = torch.outer(g2, g2)
+            f2 = f2 / f2.sum() - f1
+            f2 = f2 - f2.sum()
```
billbrod, line 82: could this not use the same `gaussian` function used by the `frontend` models and `ssim`?

plenoptic/tools/conv.py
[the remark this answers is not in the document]
billbrod: right, but the docstring makes it sound like it can only do `'circular'`. can we just remove that word? 

And is the reason it's called `same_padding` because the number of pixels it adds is equal to the original size of the image? that is, it doubles the size of `x`?

also, this does assume an image, right? 4d tensor `batch, channel, height, width`. is it worth checking that?

tests/conftest.py
```diff
@@ -60,8 +58,20 @@ def get_model(name):
         return po.metric.naive.mse
     elif name == 'ColorModel':
         return ColorModel().to(DEVICE)
-    elif name == 'FrontEnd':
-        return po.simul.FrontEnd(pretrained=True, requires_grad=False).to(DEVICE)
+
+    # FrontEnd models:
```
billbrod, line 62: looks like the only tests for the `frontend` models is with the (31, 31) size. I think we should probably test other shapes and their learnability

billbrod (changes requested):
More thoughts about these. I'm liking the look of this PR, it's going to give us a solid base of models to build from. 

The lack of any inheritance in the front end models still feels a bit weird, but their relationship isn't inheritance so much as components -- center-surround is a component of the models, which is slightly different. Inheritance seems to work best for "is a type of" relationships: metamer is a type of synthesis, and therefore the `Metamer` class should inherit the `Synthesis` superclass. But `LinearNonlinear` is not a type of `CenterSurround` (or the other way round) --- `CenterSurround` is a part of `LinearNonlinear` ([Cognitive Atlas](https://www.cognitiveatlas.org/concept/id/trm_yeQ7crXFocNza/) uses a variety of relationships in this way). Anyway, this is a bit of a tangent, but I'm not sure the way to handle those types of relationships in software engineering. @kbonnen do you know?

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,513 @@
...
+        self.kernel_size = kernel_size
+        self.width_ratio_limit = width_ratio_limit
+        self.register_buffer("amplitude_ratio", torch.tensor(amplitude_ratio))
+
+        self.center_std = nn.Parameter(torch.tensor(center_std))
+        self.surround_std = nn.Parameter(torch.tensor(surround_std))
+
+        self.pad_mode = pad_mode
+
+    @property
```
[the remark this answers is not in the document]
lyndond: I believe it does get run each time. So there is no self.filt attribute, because I don't think there needs to be one. All the information required for any arbitrarily sized `h x w` filt is in the stdev  parameters (`float`s).

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,513 @@
...
+
+    def _clamp_surround_std(self) -> Tensor:
+        """Clamps surround standard deviation to ratio_limit times center_std"""
+        return self.surround_std.clamp(
+            min=self.width_ratio_limit * float(self.center_std), max=None
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = same_padding(x, self.kernel_size, pad_mode=self.pad_mode)
+        self._clamp_surround_std()  # clip the surround stdev
```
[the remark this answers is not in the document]
lyndond: It clamps the surround stdev such that it obeys `ratio_limit`. We're enforcing that `surround_std` must be greater than `ratio_limit * center_std`, or else the surround Gaussian will be too skinny compared to the center Gaussian and will not look like a proper DoG.

plenoptic/simulate/canonical_computations/filters.py
```diff
@@ -0,0 +1,86 @@
...
+    return filt
+
+
+def circular_gaussian2d(
+    kernel_size: Union[int, Tuple[int, int]],
+    std: Union[float, Tensor],
+    n_channels: int = 1,
+) -> Tensor:
+    """Creates normalized, centered circular 2D gaussian tensor with which to convolve.
+    Parameters
```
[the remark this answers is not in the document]
lyndond: Done.

plenoptic/simulate/canonical_computations/filters.py
```diff
@@ -0,0 +1,86 @@
...
+        Size of Gaussian. Recommended to be odd so that kernel is properly centered.
+    std:
+        Standard deviation of Gaussian.
+
+    Returns
+    -------
+    filt:
+        1d Gaussian with `Size([kernel_size])`.
+    """
+    assert std > 1.0, "std must be positive"
```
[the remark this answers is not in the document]
lyndond: Shoot. yes, nice catch

plenoptic/simulate/models/frontend.py
lyndond: Fixed.

plenoptic/simulate/models/frontend.py
lyndond: Done.

plenoptic/simulate/models/frontend.py
lyndond: Done

plenoptic/tools/conv.py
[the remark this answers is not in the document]
lyndond: Ah didn't see the docstring -- you're right. Fixed. And it assumes at least 2D so it is fully compatible with torch's pad function that doesn't assume 4D.  Added an assertion w/ explanation that the last two dims must be height x width.

plenoptic/simulate/models/frontend.py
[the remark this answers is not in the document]
lyndond: Actually I see now that torch places filters the way you originally defined them. So I'll do the same: `Size([out_channels, in_channels=1, h, w])`.

plenoptic/simulate/models/frontend.py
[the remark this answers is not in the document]
lyndond: `width_ratio_limit` is a lower bound rather than an upper bound. The Surround Gaussian can't be too skinny, and `width_ratio_limit` sets a bound on how skinny it is allowed to be, relative to the Center Gaussian.
I'll make this more clear int he docs.

plenoptic/simulate/models/frontend.py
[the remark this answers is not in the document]
lyndond: They're all described in the same paragraph and figure in the paper, so I'll just refer to the paper and lcv webpage for each model.

plenoptic/simulate/models/frontend.py
lyndond: This is resolved now that `CenterSurround` has moved to `canonical_computations.filters.py`

plenoptic/simulate/models/frontend.py
lyndond: Your first sentence is resolved now that it was moved to `canonical_computations.filters.py`. It's true there's some redundancy there. The weights that need to be displayed are different w/ each model as well as their corresponding titles. I don't think it's so bad.

plenoptic/simulate/models/naive.py
lyndond: Good catch. Yes it can, and now it does.

tests/conftest.py
lyndond: The beauty of these implementations is that the learnable parameters are independent of kernel size, since all we're doing is learning gaussian stdevs and some scalars.

I'm instead going to test the shape and other arguments of `gaussian1d` and `circular_gaussian2d`, the functions that actually create the conv kernels.

plenoptic/simulate/models/naive.py
```diff
@@ -28,4 +38,180 @@ def forward(self, img):
             a clone of the input image
 
         """
-        return img.clone()
+        y = 1 * img
```
billbrod, line 41: why use this instead of `img.clone()`?

plenoptic/simulate/models/naive.py
```diff
@@ -28,4 +38,180 @@ def forward(self, img):
...
+    It splits the input greyscale image into low and high frequencies.
+
+    Parameters
+    ----------
+    kernel_size:
+        Convolutional kernel size.
+    pad_mode:
+        Mode with which to pad image using `nn.functional.pad()`.
+    default_filters:
+        Initialize the first three filters to: low-pass and high-pass on/off.
```
billbrod, line 56: this description is wrong (I think it describes the old LinearNonlinear?).

plenoptic/simulate/models/naive.py
billbrod: why is it `f2 - f2.sum()`? Should it be `f2 / f2.sum()`? (so that both f1 and f2 are normalized)

plenoptic/simulate/models/naive.py
```diff
@@ -28,4 +38,180 @@ def forward(self, img):
...
+    ):
+        super().__init__()
+
+        assert width_ratio_limit > 1.0, "stdev of surround must be greater than center"
+        assert amplitude_ratio >= 1.0, "ratio of amplitudes must at least be 1."
+
+        self.on_center = on_center
+        self.kernel_size = kernel_size
+        self.width_ratio_limit = width_ratio_limit
+        self.register_buffer("amplitude_ratio", torch.tensor(amplitude_ratio))
```
billbrod, line 184: oh this `register_buffer` is neat. But why register only `amplitude_ratio` and not `width_ratio_limit`? (or others)

plenoptic/simulate/models/naive.py
```diff
@@ -28,4 +38,180 @@ def forward(self, img):
...
+            filt = on_amp * filt_surround - filt_center
+
+        filt = filt / filt.sum()
+
+        return filt
+
+    def _clamp_surround_std(self) -> Tensor:
+        """Clamps surround standard deviation to ratio_limit times center_std"""
+        return self.surround_std.clamp(
+            min=self.width_ratio_limit * float(self.center_std), max=None
```
billbrod, line 210: does clamping in this way change `surround_std` in place? Or will this method return a value different than the actual `surround_std` Parameter and, if so, can that mess up learning?

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,414 @@
...
+
+        weights = torch.cat(
+            [
+                self.center_surround.filt,
+                self.luminance.filt,
+            ],
+            dim=0,
+        ).detach()
+
+        title = ["linear", "luminance norm",]
```
billbrod, line 198: `luminance norm` is a weird title to me (norm makes me think of a scalar, like "l2 norm"). This is the luminance filter used to normalize the output, right? is there a more descriptive name we could give?

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,414 @@
...
+        kernel_size: Union[int, Tuple[int, int]],
+        width_ratio_limit: float = 4.0,
+        amplitude_ratio: float = 1.25,
+        pad_mode: str = "circular",
+        pretrained=False,
+        activation: Callable[[Tensor], Tensor] = F.softplus,
+        apply_mask: bool = False,
+    ):
+        super().__init__()
+        kernel_size = (31, 31) if pretrained else kernel_size
```
billbrod, line 371: I think this should be documented and should throw an exception if `kernel_size != (31, 31)`, because silently changing an initialization arg is a really tricky thing to catch.

plenoptic/simulate/models/frontend.py
```diff
@@ -1,96 +1,414 @@
...
+        self.apply_mask = apply_mask
+        self._disk = None  # cached disk to apply to image
+
+    def forward(self, x: Tensor) -> Tensor:
+        y = torch.cat((self.on(x), self.off(x)), dim=1)
+
+        if self.apply_mask:
+            im_shape = x.shape[-2:]
+            if self._disk is None or self._disk.shape != im_shape:  # cache new mask
+                self._disk = make_disk(im_shape).to(x.device)
```
billbrod, line 403: I think there's an edge case where this will fail if they run the model once on the cpu, then move the image over to the gpu and run it again (or similarly run it on an image on the gpu, same size as the cpu image).

So should probably check device as well.

plenoptic/simulate/models/frontend.py
```diff
@@ -102,14 +420,52 @@ def display_filters(self, zoom=5., **kwargs):
...
-        title = ['linear on', 'linear off',
-                'luminance norm on', 'luminance norm off',
-                'contrast norm on', 'contrast norm off']
+        title = [
+            "linear on",
+            "linear off",
+            "luminance norm on",
+            "luminance norm off",
+            "contrast norm on",
+            "contrast norm off",
```
billbrod, line 441: same comment about luminance / contrast norm as above

tests/test_models.py
```diff
@@ -1,30 +1,75 @@
...
+    def test_gradient_flow(self, model):
+        img = torch.ones(1, 1, 100, 100).to(DEVICE)
+        y = model(img)
+        assert y.requires_grad
+
+    def test_onoff(self):
+        mdl = po.simul.OnOff(7, pretrained=False).to(DEVICE)
 
+    def test_pretrained_onoff(self):
+        mdl = po.simul.OnOff(7, pretrained=True).to(DEVICE)
```
billbrod, line 42: this is a weird test -- `kernel_size=7` is ignored if `pretrained=True` (see comment above)

tests/test_models.py
```diff
@@ -33,3 +78,30 @@ def test_grad(self, basic_stim):
         L = po.simul.Laplacian_Pyramid().to(DEVICE)
         y = L.analysis(basic_stim)
         assert y[0].requires_grad
+
+
+class TestFilters:
```
billbrod, line 83: yeah, I like this way of testing the kernels. 

but this doesn't test anything about the frontend models' learnability. Can we do something to test that it doesn't end up with a broken graph or the like?

billbrod: oh wait, that's what the `test_gradient_flow` does, right? nevermind, ignore this

billbrod (changes requested):
Just a couple of small changes left here.

Has this changed enough that @nikparth should take another look?

plenoptic/simulate/models/naive.py
lyndond: This way if we ever need to chain this into something the gradient can flow undisturbed.

plenoptic/simulate/models/naive.py
lyndond: `amplitude_ratio` determines how the center surround filters are normalized and need to be specified (e.g. for `frontend.OnOff`'s state_dict. The `width_ratio_limit` is only necessary for training.

plenoptic/simulate/models/naive.py
lyndond: It does in-place yes, but is ok for learning. It's similar in spirit to how people do weight clipping. The value is clamped first, then the forward pass will happen, and backprop will make adjustments to the clamped value just fine.

plenoptic/simulate/models/frontend.py
lyndond: Fair enough. Changed them to say "linear filt", "luminance filt", "contrast filt"

plenoptic/simulate/models/frontend.py
lyndond: I agree. Changed to asserting (31, 31) when pretrained=True.

plenoptic/simulate/models/frontend.py
lyndond: Changed to saying filt instead of norm

tests/test_models.py
lyndond: Fixed and added new test that tests for assertionwarning when kernel_size does not match 31x31

plenoptic/simulate/models/frontend.py
lyndond: Good catch. Fixed.

plenoptic/simulate/models/naive.py
lyndond: Fixed.

plenoptic/simulate/models/naive.py
billbrod: beautiful

billbrod (approved):
This looks good to me!

PR #107, opened 2021-03-06
Binder -- Enable online viewing/running of Jupyter examples

lyndond:
Changes:

-  Fixed bug w/ eigendistortions when k=1
- ``from tqdm.auto import tqdm`` for nicer notebook formatting when running synthesis with eigendistortions
- Re-ran 02_Eigendistortions.ipynb and Demo_Eigendistortion.ipynb w/ fixed code
- removed ``pt.imshow`` from notebooks
- Added Binder badges to README.md that will always point to master branch examples folder
- Added Binder badges to individual ipynb for 02_Eigendistortions.ipynb and Demo_Eigendistortion.ipynb so people will be able to open them online from readthedocs.io or from the github repo
- Added packages and subpackages to setup.py in order for Binder to pip install properly. Otherwise, it just throws package not found errors.
- Limit `setup.py` to torch<1.8 because steerpyr is breaking tons of steerpyr related tests bc of the torch.fft module, but Nikhil's future PR should fix it.


plenoptic/synthesize/eigendistortion.py
```diff
@@ -146,7 +146,7 @@ def synthesize(self,
                    max_steps: int = 1000,
                    p: int = 5,
                    q: int = 2,
-                   tol: float = 1e-8,
+                   tol: float = 1e-7,
```
lyndond, line 149: Added more reasonable tolerance, given that we're working with float32 by default

plenoptic/synthesize/eigendistortion.py
```diff
@@ -351,11 +351,11 @@ def _synthesize_power(self,
             Fv = fisher_info_matrix_vector_product(y, x, v, _dummy_vec)
             Fv = Fv - shift * v  # minor component
 
-            v_new = torch.qr(Fv)[0] if k > 1 else Fv
+            v_new = torch.qr(Fv)[0]  # (ortho)normalize vector(s)
```
lyndond, line 354: When k=1 before, vector wasn't being normalized after each iter. 
Using QR, it's orthonormalized the same was as when k>1. But since there's only 1 vec, it's just being normalized bc there's no other vecs to orthogonalize against. There's no noticeable performance hit in using QR when k=1; and it's cleaner to just use the same method.

setup.py
```diff
@@ -5,39 +5,60 @@
...
+    ],
+    keywords=[
+        "Visual Information Processing",
+        "PyTorch",
+              ],
+    packages=[
+        "plenoptic",
+        "plenoptic.metric",
+        "plenoptic.simulate",
+        "plenoptic.simulate.models",
```
lyndond, line 35: Needed to add all packages and subpackages (e.g. metric, simulate.models, simulate.canonical_computations) in order for Binder to pip install the package properly.

lyndond (commented):
added simple in-line comments explaining changes

plenoptic/synthesize/eigendistortion.py
```diff
@@ -3,7 +3,7 @@
 from .autodiff import jacobian, vector_jacobian_product, jacobian_vector_product
 import numpy as np
 import warnings
-from tqdm import tqdm
+from tqdm.auto import tqdm
```
billbrod, line 6: Just want to flag this -- if we go with `tqdm.auto`, we'll want to be consistent across all uses of it, and probably add a bit of information about additional requirements involved (at least [ipywidgets](https://ipywidgets.readthedocs.io/en/stable/user_install.html); I've had issues getting it the notebook tqdm working before).

setup.py
```diff
@@ -5,39 +5,60 @@
...
-    url='https://github.com/pehf/plenoptic',
-    author='LabForComputationalVision',
-    author_email='pef246@nyu.edu',
+    name="plenoptic",
+    version="0.1",
+    description="Visual Information Processing",
+    license="MIT",
+    url="https://github.com/LabForComputationalVision/plenoptic",
+    author="LabForComputationalVision",
+    author_email="pef246@nyu.edu",
```
billbrod, line 22: We should probably change this from Pierre's email, right? If we say the author is the lab, then we should have some lab-affiliated email here that we're okay getting contacted at.

We could also list all of us and all our emails: https://stackoverflow.com/questions/9999829/how-to-specify-multiple-authors-emails-in-setup-py

Not sure which is better

setup.py
```diff
@@ -5,39 +5,60 @@
...
+        "plenoptic.synthesize",
+        "plenoptic.tools",
+    ],
+    package_dir={
+        "plenoptic": "plenoptic/"
+    },
+    package_data={
+        "plenoptic": [
+            "simulate/models/weights/*.pt",
+        ]
```
billbrod, line 46: what's the user of `package_dir` and `package_data`? I haven't used them before

billbrod (changes requested):
The only change I think we need is the author/author_email fields of `setup.py`, but there are a couple of places I'd like to start a conversation (and probably make issues for)

billbrod:
Looking at the notebooks, it looks like the synthesis didn't actually run? (all the progress bars say `0/X`) But the images are there, so I'm confused.

setup.py
lyndond: Yeah was thinking the same. I'd rather not have my own email on it and would prefer perhaps a general lcv email or something.

plenoptic/synthesize/eigendistortion.py
lyndond: I agree w/ you. In running the NBs I noticed the ascii progress bars looked really wonky ( and decided if users were going to run this in Binder then they should look polished.

billbrod:
Ah yeah, that could definitely be it. Very strange.

setup.py
lyndond: Removed author email altogether.

setup.py
lyndond: I actually think we might not need it so I deleted.

setup.py
```diff
@@ -1,10 +1,10 @@
 #! /usr/bin/env python
 
-from setuptools import setup, Extension
+from setuptools import setup
 import importlib
-import os
+from importlib import util
```
billbrod, line 5: Do we use `util` somewhere?

billbrod (changes requested):
I'm not sure that `util` import is necessary, but otherwise I think this is good to go.

billbrod:
Oh one other requested change: can you document somewhere the steps to get `tqdm` working in notebooks? with `tqdm.auto`, it won't fallback to the ascii version if it can't use the fancy one -- it will raise confusing exceptions (speaking from experience). So our requirements should point out that, in addition to installing jupyter, they'll also need `ipywidgets` (I think that's the only extra requirement)

setup.py
lyndond: For some reason I got local ``Cannot find reference 'util' in '__init__.pyi' `` when just using ``import importlib``.  Adding that line made it go away. I think I'll  just remove the first ``import importlib`` and use calls directly to ``util``

billbrod (approved):
All looks good to me!

README.md
```diff
@@ -181,27 +182,47 @@ will install all the requirements necessary for plenoptic to run.
...
+   
+2. Use
+   [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels).
+   Again, if you followed the instructions above:
+
+``` sh
+# activate your 'base' environment, the default one created by miniconda
+conda activate 
+# install jupyter lab and nb_conda_kernels in your base environment
+conda install -c conda-forge jupyterlab ipywidgets
```
ThomasYerxa, line 211: funnily enough I just had to do this to download the fashionMNIST dataset inside the Demo_Steerpyr notebook

billbrod: Yeah, I think pytorch / torchvision uses the notebook `tqdm` (and thus requires `ipywidgets`) for all their downloads

ThomasYerxa:
looks good to me, left one comment on the tutorial notebook that isn't related to this PR.

ThomasYerxa (approved):

plenoptic/synthesize/eigendistortion.py
```diff
@@ -351,11 +351,11 @@ def _synthesize_power(self,
             Fv = fisher_info_matrix_vector_product(y, x, v, _dummy_vec)
             Fv = Fv - shift * v  # minor component
 
-            v_new = torch.qr(Fv)[0] if k > 1 else Fv
+            v_new = torch.qr(Fv)[0]  # (ortho)normalize vector(s)
 
             lmbda_new = fisher_info_matrix_eigenvalue(y, x, v_new, _dummy_vec)
 
-            d_lambda = (lmbda.sum() - lmbda_new.sum()).norm()  # stability of eigenspace
+            d_lambda = (lmbda - lmbda_new).norm()  # stability of eigenspace
```
Bichidian, line 358: Why make this change?

lyndond: I experimented with both and I think the new line is more precise. Each indiv eigenvec must converge to some value rather than total sum of eigenvalues converging.

PR #111, opened 2021-03-15
Fix metric.nlpd and metric.nspd for inputs on GPU

Bichidian:
This solves the issue #110.

### DIFF
--- a/examples/04_Perceptual_distance.ipynb
+++ b/examples/04_Perceptual_distance.ipynb
@@ -76,8 +76,7 @@
     }
    ],
    "source": [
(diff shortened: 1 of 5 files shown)


lyndond (approved):
Thanks -- looks fine to me. The jupyter notebook will change immensely in the future but the GPU-enabling of the .py files is greatly appreciated. 

Heads up your tests might fail because of PyTorch 1.8 messing up our steerpyr fft tests.

lyndond (approved):

billbrod (approved):
This all looks good to me. We really should be more consistent on how we name common arguments (e.g., should images be called`im` or `IM` or `img` or `image`), but that's a later problem.

lyndond (approved):

PR #114, opened 2021-03-17
Remove pooling

billbrod:
This pull request removes large swathes of code that I was using for my foveated metamer project and that we don't want to support generally:

 - removes `pooling.py` and `pooling_windows.py`, the reimplementation and extension of the windows used in Jeremy's 2011 paper, have been moved to a [new public LCV repo](https://github.com/LabForComputationalVision/pooling-windows), with tests and a readme, though no tutorial (and not on pip)
 - removes `ventral_stream.py`, the `PooledVentralStream` models I built, has been moved to my (currently-private) research repo for the project
 - removes the code I was using to normalize the representation of those models. Kate is using a more principled way of doing this for the Portilla-Simoncelli texture stats
 - removes `fraction_removed`, `loss_change_fraction` and support for SWA (stochastic weight averaging, and thus `torchcontrib`). All of these made the synthesis optimization code more complicated, for little benefit.
 - overhauls save/load: changed load from a class method to an instance one. so now you have to initialize the Synthesis method and then call load, which modifies it in-place. This allows us to avoid figuring out how to deal with models (which can get quite large). We check that the initialization was called with the same base signal, model, and loss function.
 - changes to tests for all of this. also removed some extraneous tests (e.g., tests of animate and plot in `test_metamers.py` and `test_mad.py`)
 - changes to notebooks for all of this. There are some places that will need to be revisited after the Portilla-Simoncelli model is pulled in, where I talk about the use of custom `plot_representation` methods and coarse-to-fine optimization (in particular, I edited out blocks of code, which still refer to `PooledV1`)


plenoptic/synthesize/synthesis.py
```diff
@@ -1075,117 +972,89 @@ def save(self, file_path, save_model_reduced=False, attrs=['model'],
             save_dict[k] = attr
         torch.save(save_dict, file_path, pickle_module=dill)
 
-    @classmethod
     @abc.abstractmethod
-    def load(cls, file_path, model_attr_name='model', model_constructor=None, map_location='cpu',
-             **state_dict_kwargs):
-        r"""load all relevant stuff from a .pt file
+    def load(self, file_path, map_location='cpu',
+             check_attributes=['base_signal', 'base_representation',
```
lyndond, line 977: Should this even be optional?

plenoptic/synthesize/synthesis.py
```diff
@@ -1075,117 +972,89 @@ def save(self, file_path, save_model_reduced=False, attrs=['model'],
...
-            any additional kwargs will be added to the model's
-            state_dict before construction (this only applies if the
-            model is a dict, see above for more description of that)
-
-        Returns
-        -------
-        synthesis : plenoptic.synth.Synthesis
-            The loaded synthesis object
-
+        check_attributes : list, optional
```
lyndond, line 1003: Regarding this and ``save()``: what if we just force users to save/load very specific attrs. Since we're not tethered to their model anymore and are just dealing w our own Synth objects, we should know/force exactly what is saved and loaded. It could cut down on complexity in the load function.

plenoptic/synthesize/synthesis.py
```diff
@@ -1075,117 +972,89 @@ def save(self, file_path, save_model_reduced=False, attrs=['model'],
...
+                                f" initialization, but got attr {k}!")
+            if isinstance(getattr(self, k), torch.Tensor):
+                try:
+                    if not torch.allclose(getattr(self, k), tmp_dict[k]):
+                        raise Exception(f"Saved and initialized {k} are different! Initialized: {getattr(self, k)}"
+                                        f", Saved: {tmp_dict[k]}, difference: {getattr(self, k) - tmp_dict[k]}")
+                except RuntimeError:
+                    raise Exception(f"Attribute {k} is a different shape in saved and initialized versions!"
+                                    f" Initialized: {getattr(self, k).shape}, Saved: {tmp_dict[k].shape}")
+            elif k == 'loss_function':
```
lyndond, line 1041: Confused as to what's going on here.

lyndond (changes requested):
Notebook comments: mainly that they're long and we should consider splitting advanced usage into their own nb. Don't have to do that in this PR though.

General comments: Mainly curious about users really needing this any flexibility in what attrs to save/load, now that we've stripped models out?

plenoptic/synthesize/synthesis.py
billbrod: I added some comments trying to explain this

plenoptic/synthesize/synthesis.py
billbrod: Should `check_attributes` be optional? See my response in your next comment about this.

plenoptic/synthesize/synthesis.py
billbrod: I agree: the user does not set the `check_attribute`: `Synthesis` is not user-facing, it's inherited by other synthesis methods. If you look at `Metamer` and `MADCompetition`, the user doesn't specify the attributes to save/load. But `Metamer` and `MADCompetition` have different attributes to check (in particular, MAD has `base_representation_1` and `base_representation_2`, and handles loss functions slightly differently, so it checks that itself and doesn't have the super class do it); this allows the developer of new synthesis methods to specify which attributes should be checked.

Does that make sense?

lyndond: Ah yes makes sense now. Forgot this was an abc.

lyndond (approved):
g2g

kbonnen (approved):
I asked for 1-2 localized changes and made a couple of comments (some are in notebooks).  But I'm going to go ahead and submit this review as approved so that you're not blocked because of me.

Bichidian (commented):
I found two places where the arguments seem not to be used by the functions.

plenoptic/synthesize/synthesis.py
```diff
@@ -1075,117 +972,101 @@ def save(self, file_path, save_model_reduced=False, attrs=['model'],
...
+            ``Synthesis`` object and the loaded one. Checking the model is
+            generally not recommended, since it can be hard to do (checking
+            callable objects is hard in Python) -- instead, checking the
+            ``base_representation`` should ensure the model hasn't functinoally
+            changed.
+        objective_function_kwargs : dict, optional
+            Additional arguments to pass to the objective function call.
+        pickle_load_args :
+            any additional kwargs will be added to ``pickle_module.load`` via
+            ``torch.load``, see that function's docstring for details.
```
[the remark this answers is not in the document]
billbrod: Good catch! `objective_function_kwargs` I eventually decided not to use, and `pickle_load_args` should be passed to `torch.load`

tests/test_display.py
```diff
@@ -463,8 +463,11 @@ def forward(self, *args, **kwargs):
 
 
 def template_test_synthesis_all_plot(synthesis_object, iteration,
-                                     plot_synthesized_image, plot_loss, plot_representation_error, plot_image_hist,
-                                     plot_rep_comparison, plot_signal_comparison, fig_creation):
+                                     plot_synthesized_image, plot_loss,
+                                     plot_representation_error,
+                                     plot_image_hist, plot_rep_comparison,
+                                     plot_signal_comparison, fig_creation,
+                                     width_ratios={}):
```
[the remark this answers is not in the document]
billbrod: Yup -- it shouldn't be an argument at all, but created in the function and passed to `plot_synthesis_status`. Whoops

Bichidian (approved):

PR #115, opened 2021-03-18
Test fixes

billbrod:
Quality-of-life changes to tests:
- reduce time tests make, especially `test_mad.py`, by removing unnecessary tests, making use of fixtures to reduce unnecessary loading or initializations, and caching the python environment
- making use of fixtures requires `pytest>=5.1.2`
- `build` Github action (which runs the contents of `tests/` folder) now only runs in open PRs (to master) and on a weekly schedule (in master). we were unnecessarily running too many tests. We should also be able to manually trigger tests, but I couldn't check that because I think it needs to be in master first
- ffmpeg no longer a requirement for `build`, but still required for `tutorials` (the notebooks); we use the [setup ffmpeg](https://github.com/marketplace/actions/setup-ffmpeg) action for this, which seems (so far at least) to be more reliable
- `tutorials` now runs on same set of actions as `build`
- adds a bit of info to `CONTRIBUTING` about making use of `pytest.mark.parametrize` and the fixtures. 
- adds info to `README` about ffmpeg and animation backends (as well as in notes on the different animate functions)
- adds file `conftest.py`, which contains the fixtures and other objects that are shared across tests (`DEVICE, `DATA_DIR`, `DTYPE`)

@nikparth and @lyndond I changed your test scripts most substantially, so you both should take a look. I don't think I removed anything that was being tested (for the steerable pyramid, I may have actually added some), but I had to restructure things in order to make use of fixtures.


tests/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+                         as_gray=False).to(DEVICE)
+    return img[..., :256, :256]
+
+
+@pytest.fixture(scope='package')
+def basic_stim():
+    return po.make_basic_stimuli().to(DEVICE)
+
+
+def get_model(name):
```
lyndond, line 45: nice

plenoptic/synthesize/mad_competition.py
```diff
@@ -2097,6 +2097,18 @@ def animate(self, synthesis_target=None, batch_idx=0, channel_idx=None,
             The animation object. In order to view, must convert to HTML
             or save.
 
+        Notes
```
lyndond, line 2100: not sure how I feel about all these being duplicated across diff files

lyndond (approved):
Looks fine to me. I'm not a fan of the repetition of the ffmpeg bit i docstrings.

plenoptic/synthesize/mad_competition.py
billbrod: Yeah, that's fair. This is a symptom of a broader issue we've talked about -- with the inheritance of methods across `Synthesis` means the docstrings get duplicated or changed slightly. We have an issue for it, #25, and the solution seaborn uses is to have string constants containing the info that can then be inserted into the docstring. We could check other libraries to see how they handle it too. 

I don't have a problem with the info showing up in multiple places (in fact, I think it should -- the user should see the info where it's relevant to them, though I'd also be fine with a brief discussion in most docstrings, pointing to a "main" one with longer discussion), but it makes it hard to keep everything up-to-date

tests/test_steerable_pyr.py
```diff
@@ -84,212 +86,170 @@ def check_parseval(im ,coeff, rtol=1e-4, atol=0):
...
+        # height has k[0]==0)
+        height = max([k[0] for k in spyr.pyr_coeffs.keys() if isinstance(k[0], int)]) + 1
+        # couldn't come up with a way to get this with fixtures, so we
+        # instantiate it each time.
+        spyr_not_downsample = po.simul.Steerable_Pyramid_Freq(img.shape[-2:], height, spyr.order,
+                                                              is_complex=spyr.is_complex,
+                                                              downsample=False, tight_frame=True)
+        spyr_not_downsample.to(DEVICE)
+        spyr_not_downsample.forward(img)
+        check_band_energies(spyr.pyr_coeffs, spyr_not_downsample.pyr_coeffs)
```
nikparth, line 160: this all looks good to me.  A more general point about these tests (not necessarily for your PR) I was thinking about is do we really need the tests that test the same shape but just transposed? For example, an image that is (128, 256) and (256, 128) will have the same scales because we set scales based on the min size. And I'm not sure that there is really anything else that would change in the pyramid. I'm not sure what the current runtime is for the pyr tests but if it is long, then these tests feel a bit redundant to me but it's not a big deal either way.

nikparth (approved):
left a brief comment in the pyr tests about something we might want to discuss, but not necessary to change anything for this PR. Everything else looks good to me. I would tend to agree with lyndon that if we can find a way to reduce the duplicate documentation but still have the documentation point to the original in all of the places it is relevant then that would be nice. (The string constant option from seaborn doesn't seem too bad to me.)

tests/test_steerable_pyr.py
billbrod: The pyramid tests take 5 minutes, so I think the amount of time is fine. And I'd lean towards keeping them in because they should be the same, so it's worth checking to make sure -- it would be really bad if they weren't. I based this off the existing tests, so I don't know if there was a reason why we were explicitly testing both variants beforehand.

nikparth: Yeah this wasn't about your PR, was mostly about why we were doing that from the beginning. but if it's 5 mins then doesn't matter, but if we add more tests down the line and it ends up being cumbersome, maybe worth revisiting.

PR #117, opened 2021-03-20
Run tests on GPU and fix thus found errors

Bichidian:
In tests files, I move models to GPU when it is available, and fix errors that are discovered in GPU tests. These are all done after pulling the branch `remove_pooling`, which is currently a PR in review.

`test_plenoptic.py` and `test_model.py` did not use GPU when available, so I made them do so. `test_display.py` partially does, and I did not increase coverage of GPU because I think it's not necessary for all display functions. `test_eigendistortion.py`, `test_metamers.py`, `test_mad.py` and `test_steerable_pyr.py` were already set to used GPU when available. Most of these files produce some errors when tested on GPU, and I believe I have fixed all of them.


lyndond (approved):
Looks fine to me. Not sure if this one will be pulled to master before #115, but if it's after then you'll likely have to make a few adjustments.

Bichidian:
Yes, I think I need to wait for #114 and #115.

plenoptic/synthesize/mad_competition.py
```diff
@@ -365,7 +365,7 @@ def update_target(self, synthesis_target, step):
                     setattr(self, attr, torch.nn.Parameter(getattr(self, attr+'_all')[synthesis_target]))
                 else:
                     try:
-                        setattr(self, attr, getattr(self, attr+'_all')[synthesis_target].clone().to('cpu'))
+                        setattr(self, attr, getattr(self, attr+'_all')[synthesis_target].clone())
```
billbrod, line 368: Does `clone` automatically move to the cpu?

plenoptic/synthesize/synthesis.py
```diff
@@ -1037,7 +1037,10 @@ def load(self, file_path, map_location='cpu',
                         raise Exception(f"Saved and initialized {k} are different! Initialized: {getattr(self, k)}"
                                         f", Saved: {tmp_dict[k]}, difference: {getattr(self, k) - tmp_dict[k]}")
                 except RuntimeError:
-                    raise Exception(f"Attribute {k} is a different shape in saved and initialized versions!"
+                    if getattr(self, k).device != tmp_dict[k].device:
+                        raise Exception(f"Attribute {k} are on different devices in saved and initialized versions!"
+                                        f" Initialized: {getattr(self, k).device}, Saved: {tmp_dict[k].device}")
```
billbrod, line 1042: I don't think this should  actually be an error, should it? If someone saves on the GPU, then sets `map_location='cpu'` to load on the CPU, this will throw an Exception, which we don't want.

But in order to compare them, they'll need to be on the same device, so maybe copy the saved version to the initialized device?

tests/test_display.py
```diff
@@ -610,10 +610,10 @@ def synthesized_met(self, request):
         img, model = request.param.split('-')
         # make the images really small so nothing takes as long
         if img == 'rgb':
-            img = po.load_images(op.join(DATA_DIR, 'color_wheel.jpg'), False)
+            img = po.load_images(op.join(DATA_DIR, 'color_wheel.jpg'), False).to(DEVICE)
             img = img[..., :16, :16]
         else:
-            img = po.load_images(op.join(DATA_DIR, 'nuts.pgm'))
+            img = po.load_images(op.join(DATA_DIR, 'nuts.pgm')).to(DEVICE)
```
billbrod, line 616: This is good! I'm pretty sure this means that, for those synthesis tests, synthesis gets run on the GPU and then we test that display works as expected? That's a good test and I think sufficient for testing device in test_display.py -- the various display code should all copy things onto cpu for plotting.

tests/test_mad.py
```diff
@@ -159,11 +159,10 @@ def test_save_load(self, curie_img, test_all, fail, tmp_path):
                 loss = lambda *args, **kwargs: 1
             mad_copy = po.synth.MADCompetition(curie_img, model, model2, loss_function=loss)
             with pytest.raises(Exception):
-                mad_copy.load(op.join(tmp_path, "test_mad_save_load.pt"),
-                              map_location=DEVICE)
+                mad_copy.load(op.join(tmp_path, "test_mad_save_load.pt"))
         else:
             mad_copy = po.synth.MADCompetition(curie_img, model, model2)
-            mad_copy.load(op.join(tmp_path, "test_mad_save_load.pt"), map_location=DEVICE)
+            mad_copy.load(op.join(tmp_path, "test_mad_save_load.pt"))
```
billbrod, line 165: why did you remove the map_location arg here?

tests/test_metamers.py
```diff
@@ -41,12 +41,10 @@ def test_metamer_save_load(self, einstein_img, model, loss_func, fail, tmp_path)
...
             met_copy = po.synth.Metamer(einstein_img, model, loss_function=loss, loss_function_kwargs=loss_kwargs)
             with pytest.raises(Exception):
-                met_copy.load(op.join(tmp_path, "test_metamer_save_load.pt"),
-                              map_location=DEVICE)
+                met_copy.load(op.join(tmp_path, "test_metamer_save_load.pt"))
         else:
             met_copy = po.synth.Metamer(einstein_img, model, loss_function=loss, loss_function_kwargs=loss_kwargs)
-            met_copy.load(op.join(tmp_path, "test_metamer_save_load.pt"),
-                          map_location=DEVICE)
+            met_copy.load(op.join(tmp_path, "test_metamer_save_load.pt"))
```
billbrod, line 47: why did you remove the map_location arg here?

billbrod (changes requested):
This looks good to me, I just have a couple of questions I'd like clarity on before approving.

Bigger picture question: how do we know how many of our tests are running on GPU? Is there an easy way to check that? I assume the errors that you caught were where half the tensors involved ended up on GPU and half on the CPU, but what about those (like `test_models.py`), where nothing was happening on the GPU? Is there an easy way to check that we've caught them all (or, like much of `test_display.py`, don't care)?

tests/test_mad.py
Bichidian: This is because `self.saved_signal`, `self.saved_representation`, `self.saved_signal_gradient` and `self.saved_representation_gradient` are always moved to CPU regardless of whether other variables are on GPU. When `map_location="gpu"`, these variables will be loaded to GPU, and will encounter problem when new values (which are on CPU) are appended to them. So currently one cannot use `map_location="gpu"`. In order to support it, I think we need to force these four variables to be loaded to CPU regardless of `map_location`.

plenoptic/synthesize/synthesis.py
Bichidian: You're right. So I can change the code to be asserting `torch.allclose(getattr(self, k).to(tmp_dict[k].device), tmp_dict[k])`.
The resulting behavior would be tolerating initiated and loaded data to be on different devices, and eventually using the device of `tmp_dict`, which is determined by `map_location` when it is not `None`.

plenoptic/synthesize/mad_competition.py
Bichidian: No, it doesn't. Do you need to move to CPU here?

tests/test_display.py
Bichidian: Yes, this means that tests in `TestMADDisplay` and `TestMetamerDisplay` are run on GPU when available.

Bichidian:
> This looks good to me, I just have a couple of questions I'd like clarity on before approving.
> 
> Bigger picture question: how do we know how many of our tests are running on GPU? Is there an easy way to check that? I assume the errors that you caught were where half the tensors involved ended up on GPU and half on the CPU, but what about those (like `test_models.py`), where nothing was happening on the GPU? Is there an easy way to check that we've caught them all (or, like much of `test_display.py`, don't care)?

I basically went through all tests and moved models and data to `DEVICE` if they were not already there. It is not impossible that I have missed something. Another option is to write at the beginning of the file:
`if torch.cuda.is_availble(): torch.set_default_tensor_type(torch.cuda.FloatTensor)`
This will create all tensors on GPU when GPU is available, removing the need of writing `.to(DEVICE)` or `device=DEVICE` all the time, and also making sure every test is being run on GPU when applicable. But this can also break things when you create tensors that should be on CPU. Also, writing `.to(DEVICE)` or `device=DEVICE` was the norm in this project.

plenoptic/synthesize/mad_competition.py
billbrod: Looking through the history, it looks like I added that `.to('cpu')` to get `MADCompetition` working with the plotting and animating code, so assuming those tests in `test_display.py` pass, then this is fine.  (My guess is there's some copying over to cpu that happens in the display code, which is the proper way to do it)

tests/test_mad.py
billbrod: Ah, that's a good point. We should probably fix that in this PR (or, have load raise an error if `map_location='gpu'` and open an issue describing this). I think that's a good solution, though. As I was thinking through this, I double-checked what I do for `to`, and I'm pretty sure I allow those variables to be passed to the gpu, which is also no good; they should just never change device. I think we don't currently test `to()`, which we should do as well.

plenoptic/synthesize/synthesis.py
billbrod: I think your solution is correct, though I guess the question is whether we want everything to end up on the device of `tmp_dict[k]` or of `self`. Right now, I think the final device is determined by `map_location`, which is what we want, I think

billbrod:
Hmm, I do like that idea of setting the default tensor type at the top of the tests, which will force every test onto the gpu. then, if there are tests we don't want to run on the gpu, we'd have to make that explicit, which is nice (so we know what's supported on both and which is cpu-only). We probably want to support both types of usage (if it's not too hard), since people will do both.

Bichidian:
Actually, what I initially considered was applying this to all files, not just test files. I'm not yet sure about the behavior of `torch.set_default_tensor_type` across multiple files. If I could restrict its effect for the test files, then it would be more manageable for this PR. However, such a change would not totally remove the need of using `DEVICE`, since we call many external functions (e.g. `po.load_images`) that would not be covered by `torch.set_default_tensor_type`. So this change still would not force every test to GPU without possibility of human error. For this PR, I think I will keep the current solution.

lyndond:
Does setting default tensor type have global effect? If so then I don't support it since I don't want our package messing with other people's default workflows.

billbrod:
Yeah, I agree with both of you. Let's avoid using `set_default_tensor_type` then. Still do wish there were an issue way to see what we're testing on GPUs, but I guess there's no easy fix.

Before approving, I would still like the two threads I haven't resolved addressed.

billbrod (approved):
This looks good to me! I like the solution for `to`, that's a nice way of doing it

PR #119, opened 2021-03-29
Steer pyr demo

ThomasYerxa:
Demonstrates the use of steer pyr by (1) using it as a front end for a conv net classifier on FashionMNIST and (2) comparing the eigendistortions w/ a baseline classifier that doesn't use a steer pyr frontend.


ThomasYerxa:
converted to PR prematurely. closing for now

ThomasYerxa:
When the dust settles, the only significant change coming from this branch should be the Demo_Steerable_Pyramid.ipynb notebook. That should be relatively stable to the one appearing in this PR, so feel free to glance at the notebook if curious but ignore everything else.

billbrod:
I just looked at the steerable pyramid notebook. Let me know if you have trouble merging `main` into this and whether I should look at any other files

ThomasYerxa:
Questions:
1. Was there a particular reason we opted for a frequency space implementation (boundary conditions are nicer etc.)? In "1.1: Visualizing wavelets" it would be nice to mention if so.
2. For the steerability demo: is this sufficient/interesting? would it be more compelling i.e. to steer to an intermediate orientation and then show that if you initialize a pyramid that includes the steered-to-orientation you can recover the same coefficients.

billbrod:
I've gone through and edited the text here, take another look at it and tell me what you think. @nikparth can you also take a look at this?

- I think the final section, on probing the effect of using the steerable pyramid as a frontend, is a great idea, but I think we need some more text at the end to explain. I tried adding some, but it still feels like we ran out of steam while writing it and are just saying "whelp, you make sense of it" . I also think need a bit more detail about what the networks are being trained to do, some framing, etc.
- I'm tempted to actually split that whole section off into a separate notebook, so we have two notebooks: steerable pyramid basics and steerable pyramid-as-model component. I could add something about classic visual models there too. Thoughts?
- Re: your question 1. I think it's because that was the only one we really want people to use: it handles any order and reconstruction is exact. @nikparth do you remember if there was another reason.
- Re: question 2. I think this is fine for now, but agree that steering an intermediate representation would be more compelling. Can you add info to #204 ? Add all the notes you have, any code, etc, so that you or someone else can pick it up later.

nikparth (approved):
I think overall this seems fine as a first release. I added some comments on wording in the notebook, but my main comments would be to remove the standard CNN demonstration and comparison to the pyramid. I think it's fine to just show that the pyramid can be added as a module to build on with standard pytorch modules. And then I'd say remove the eigendistortion section as well and we could have another notebook that is about using plenoptic methods with canonical visual models.

billbrod (approved):

PR #120, opened 2021-03-30
Complex pyramid

nikparth:
Change to the core steerable pyramid code and utilities to allow for use of the ComplexTensor features in pytorch 1.8
- changes all pyramid code and utils to use the new ComplexTensor dtype
- makes changes to use the new torch.fft library
- minor changes in pyramid tools (removal of batchfftshift functions)


plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -93,16 +94,14 @@ def __init__(self, image_shape, height='auto', order=3, twidth=1, is_complex=Fal
         self.downsample = downsample
         self.tight_frame = tight_frame
```
lyndond, line 95: self.tight_frame = True if norm="ortho" else False
Alternatively we could just do away with tight_frame altogether and just deal w/ conditioning on self.fft_norm.

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -346,29 +343,18 @@ def forward(self, x, scales=[]):
...
+                    band = fft.ifftshift(banddft)
+                    band = fft.ifft2(band, dim=(-2,-1), norm=self.fft_norm)
+                    
                     if not self.is_complex:
-                        band = torch.unbind(band, -1)[0]
-                        self.pyr_coeffs[(i, b)] = band
-                        self.pyr_size[(i, b)] = tuple(band.shape[-2:])
+                        pyr_coeffs[(i, b)] = band.real
                     else:
                         if self.tight_frame:
```
lyndond, line 354: Potentially condition on self.fft_norm instead

lyndond (changes requested):
Minor suggestions and changes. 

Can u also add tests for `tensor_to_pyr()`, and `steer_coeffs()`

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -79,7 +80,7 @@ class Steerable_Pyramid_Freq(nn.Module):
     """
 
     def __init__(self, image_shape, height='auto', order=3, twidth=1, is_complex=False,
-                  downsample=True, tight_frame=True):
+                  downsample=True,  tight_frame=False):
```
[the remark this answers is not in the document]
nikparth: I went back and forth on this.. the norm argument is about the actual normalization on the fft whereas the tight_frame has the sqrt(2) stuff inside the code. 
Like I guess since we don't provide an option to set the fft norm arguments directly I could just call this norm, but I didn't want to confuse it with the norm in numpy or pytorch which is a string ("ortho" , "backward", "forward")

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
nikparth: yeah only thing is I dont want people to assume that setting "ortho" is the only thing that makes it tight frame but idk maybe that's not a big deal since we hide it from the user anyway.

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -306,7 +305,8 @@ def forward(self, x, scales=[]):
             If downsample is true, representation is an OrderedDict of the coefficients.
 
         """
-        self.pyr_coeffs = OrderedDict()
+        pyr_coeffs = OrderedDict()
+        self.pyr_size = OrderedDict()
```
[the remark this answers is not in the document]
nikparth: well the self.pyr_coeffs --> pyr_coeffs is because we dont want to store the coeffs as an attribute anymore for memory reasons. the self.pyr_size was more for readability since it always gets filled in the forward, might as well put it there, so it's close to where it's actually used, but I guess we can put it back in the initialization to keep all attributes there.

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -306,7 +305,8 @@ def forward(self, x, scales=[]):
             If downsample is true, representation is an OrderedDict of the coefficients.
 
         """
-        self.pyr_coeffs = OrderedDict()
```
kbonnen, line 309: So we're no longer storing pyr_coeffs in the class.  I'm basically on board, but can you walk me through the reasons why you decided that?

nikparth: it's primarily for when using the pyramid with large batches to limit the memory usage when we have for example the coefficients stored as a dictionary and then we return as a tensor in the non-downsampled version, the coeffs were stored in two places which doubles the memory. We thought the easiest thing would just be to let the module return the coeffs and the user can do whatever with it and delete it when necessary and that way the pyramid object is low-memory itself.

kbonnen (approved):
General review: I added a couple of questions.  There was one place that I noted that a comment block needed to be updated with a new variable name.

Note: I've been using the complex_pyramid branch successfully with replicate-portilla-simoncelli.  The only real barrier to using it was figuring out how complexTensors (and their correspondign types) are created by fft.  Initially I thought we might want some comments for what's going on with this but now I think that's not our responsibility to explain.

I'm signed off since my one change request is so small.

plenoptic/tools/conv.py
```diff
@@ -92,8 +92,8 @@ def blur_downsample(x,filtname = 'binom5', step=(2, 2)):
     return correlate_downsample(x, filt=np.outer(f, f), step=step)
 
 
-def upsample_blur(x, step=(2, 2)):
-    f = pt.named_filter('binom5')
+def upsample_blur(x, filtname ='binom5', step=(2, 2)):
+    f = pt.named_filter(filtname)
```
billbrod, line 96: We should add a docstring here and include the possible filters (just copying over the relevant part from `pt.named_filter` docstring

plenoptic/tools/data.py
```diff
@@ -12,6 +12,21 @@
...
+        np.int32      : torch.int32,
+        np.int64      : torch.int64,
+        np.float16    : torch.float16,
+        np.float32    : torch.float32,
+        np.float64    : torch.float64,
+        np.complex64  : torch.complex64,
+        np.complex128 : torch.complex128
+    }
+
+TORCH_TO_NUMPY_TYPES = {value : key for (key, value) in NUMPY_TO_TORCH_TYPES.items()}
```
billbrod, line 29: I like this solution!

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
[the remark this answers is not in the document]
billbrod: agreed with Kate's point -- would be nice to have some details on when someone would want to set this true or false. (might not belong here, could be in the notebook)

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -185,21 +179,21 @@ def __init__(self, image_shape, height='auto', order=3, twidth=1, is_complex=Fal
                 Ycosn_recon = Ycosn_forward
 
             himask = pointOp(log_rad, self.Yrcos, Xrcos)
-            self._himasks.append(torch.tensor(himask).unsqueeze(0).unsqueeze(-1))
+            self._himasks.append(torch.tensor(himask).unsqueeze(0))
```
billbrod, line 182: why have the masks changed shape?

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -321,89 +316,98 @@ def forward(self, x, scales=[]):
...
+                #compute filter output at each orientation
                 for b in range(self.num_orientations):
+                    '''
+                    band pass filtering is done in the fourier space as multiplying by the fft of a gaussian derivative.
+                    The oriented dft is computed as a product of the fft of the low-passed component,
+                    the precomputed anglemask (specifies orientation), and the precomputed hipass mask (creating a bandpass filter)
+                    the complex_const variable comes from the Fourier transform of a gaussian derivative.
+                    Based on the order of the gaussian, this constant changes.
+                    
+                    '''
```
billbrod, line 350: minor point: we can use the `'''` as multi-line comment in the middle of a function, but I think I would prefer the `#` for consistency. Most syntax highlighters (at least, my emacs one and the github one here) seem to make the `'''` comment look like regular strings, which is confusing

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -321,89 +316,98 @@ def forward(self, x, scales=[]):
...
-                        band = torch.unbind(band, -1)[0]
-                        self.pyr_coeffs[(i, b)] = band
-                        self.pyr_size[(i, b)] = tuple(band.shape[-2:])
+                        pyr_coeffs[(i, b)] = band.real
                     else:
+                        '''
+                        Because the input signal is real, to maintain a tight frame 
+                        if the complex pyramid is used, magnitudes need to be divided by sqrt(2) 
+                        because energy is doubled.
+                        '''
```
billbrod, line 367: can the pyramid only be tight frame if it's complex? (all the `if self.tight_frame` statements are nested within or paired with `if self.is_complex`). If so, we should clarify that in the docstring and raise an exception if they try to set `is_complex=False, tight_frame=True`

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -412,11 +416,15 @@ def convert_pyr_to_tensor(self, pyr_coeffs):
         ----------
         pyr_coeffs: `OrderedDict`
             the pyramid coefficients
```
billbrod, line 418: we talked about this a bit in person -- should have more detail in the docstring about how batches, channels, real/imag are all interwoven.

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -321,89 +316,98 @@ def forward(self, x, scales=[]):
...
+            lo0 = fft.ifftshift(lodft)
+            lo0 = fft.ifft2(lo0, dim=(-2,-1), norm=self.fft_norm)
+            pyr_coeffs['residual_lowpass'] = lo0.real
+            self.pyr_size['residual_lowpass'] = tuple(lo0.real.shape[-2:])
+
+        return pyr_coeffs
 
 
-    def convert_pyr_to_tensor(self, pyr_coeffs):
+    def convert_pyr_to_tensor(self, pyr_coeffs, split_complex=False):
```
billbrod, line 410: minor point: can this be a class method? You check `self.downsample` but the same thing can be checked with a try/except when you try to concatenate across bands.

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -777,14 +764,16 @@ def _recon_levels(self, pyr_coeffs, recon_keys, scale):
 
 
 
-    def steer_coeffs(self, angles, even_phase=True):
+    def steer_coeffs(self, pyr_coeffs, angles, even_phase=True):
```
billbrod, line 767: minor point: can this be a class method? only use of `self` is in checking `self.num_orientation` and `self.num_scales`, but both can be gotten from the keys of `pyr_coeffs`

billbrod (changes requested):
This looks good! Much cleaner and, with the new comments, much clearer about what's going on. I have a couple questions scattered throughout.

Also, I'm going to leave the notebook until #119

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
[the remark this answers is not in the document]
nikparth: yeah I think when we do a second pass on notebooks we can include that in the notebook because it's really dependent on the problem and usage and so it's not so easy to do just in the docstring for the class.

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -346,29 +343,18 @@ def forward(self, x, scales=[]):
                     anglemask = self._anglemasks[i][b]
 
                     # bandpass filtering
-                    banddft = lodft * anglemask * himask
-                    banddft = torch.unbind(banddft, -1)
-                    # (x+yi)(u+vi) = (xu-yv) + (xv+yu)i
                     complex_const = np.power(np.complex(0, -1), self.order)
```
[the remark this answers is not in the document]
nikparth: done

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
nikparth: done in code

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
nikparth: because the masks don't need to be unsqueezed for the last dimension since the complex tensor is now just 4 dimensional instead of 5

tests/test_display.py
```diff
@@ -364,46 +364,10 @@ def test_imshow(self, as_rgb, channel_idx, batch_idx, is_complex, mini_im,
...
-                    raise Exception(f"Created {len(fig.axes)} axes, but expected {n_axes}!")
-                plt.close('all')
-        else:
-            with pytest.raises(TypeError):
-                po.pyrshow(coeffs, batch_idx=batch_idx, channel_idx=channel_idx)
-
+    def test_display_test_signals(self):
+        po.imshow(po.tools.make_synthetic_stimuli(128));
+        po.imshow(po.load_images(DATA_DIR + '/512'));
+        po.imshow(po.load_images(DATA_DIR + '/256'));
```
billbrod: what happened to `test_pyrshow`?

plenoptic/tools/signal.py
```diff
@@ -328,8 +251,8 @@ def rectangular_to_polar(real, imaginary):
     TODO: torch.abs(cmplx) torch.angle(cmplx)
```
billbrod, line 251: this docstring needs to be updated, since now it does operate on a complex tensor

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
```diff
@@ -481,22 +501,38 @@ def convert_tensor_to_pyr(self, pyr_tensor):
...
 
         Returns
         ----------
         pyr_coeffs: `OrderedDict`
             pyramid coefficients in dictionary format
         """
-
+        num_channels = pyr_info[0]
+        split_complex = pyr_info[1]
+        pyr_keys = pyr_info[2]
```
billbrod, line 517: I guess this info can't be inferred somehow? I don't love the idea of returning what's basically a dummy variable from one method to pass to this one, but if it can't be avoided...

I have a preference for splitting out these three values into three separate arguments, to make it clearer. But if the use case is always just grabbing the value returned from `convert_pyr_to_tensor` and passing it here, I guess it doesn't make too much of a difference. what do others think?

billbrod (changes requested):
A couple small questions here. I'd also like the points I brought up earlier addressed (just requests for more info on tight frame).

plenoptic/tools/conv.py
billbrod: I still think this is true, but all the functions in `conv.py` need their docstrings revisited, so I think this can wait.

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
nikparth: The pyramid can be tight_frame even if it's not complex- see the new pyramid file for updated code- not all of the if statements are nested with a self.is_complex now.

tests/test_display.py
nikparth: added back- was removed in merge with main accidentally

plenoptic/simulate/canonical_computations/steerable_pyramid_freq.py
nikparth: well so the `pyr_keys` piece of the pyr_info list is a useful thing to return because it allows the user to matchup the keys with the tensor channel order. the other two things that are returned yes are more just for the paired `convert_tensor_to_pyr` function. They can't be inferred unless I turn these functions back into to class methods and change class attributes that stores these things when the convert function is called? Idk what you think is better..  And lastly yes I could split them- it just makes it easier to handle when you don't need them.

plenoptic/tools/signal.py
nikparth: updated

lyndond (approved):
g2g

billbrod (approved):
This is good with me!

PR #122, opened 2021-04-01
use pytest-xdist to parallelize test_display (on a single runner)

billbrod:
This PR adds the use of the `pytest-xdist` plugin to parallelize the `test_display.py` script (on a single runner; our tests are also parallelized across runners). The others are run as normal. This reduces the time required for `test_display` from ~30 minutes to ~20, while the others are unchanged.


lyndond (approved):
nice! g2g

Bichidian (approved):

PR #124, opened 2021-04-13
MS-SSIM

Bichidian:
The result of this implementation is confirmed to be the same as the pip package `pytorch-msssim`. Probably need to add more tests.


Bichidian:
It turns out that previously I have only tested on images whose shapes are multiples of 16, which means that padding never mattered. When I test on images of arbitrary shapes, my result is no longer the same as `pytorch-msssim`. So I did padding according to the [original code written in MATLAB](https://ece.uwaterloo.ca/~z70wang/research/iwssim/msssim.zip), and then my result is the same as the original code (up to a numeric error). My result is still not the same as `pytorch-msssim` since it uses a different padding mode from the original code.

Another thing is that the `weight` in `_ssim_parts` seems to have mistakenly computed matrix multiplication instead of elementwise multiplication, which I found out because it broke on non-square images. After correcting this, one of the tests fail, because when `weighted=True` the SSIM values are no longer close to the previously computed values. I think it means that the previous values are also incorrect.

plenoptic/metric/perceptual_distance.py
```diff
@@ -250,6 +245,83 @@ def ssim_map(img1, img2, dynamic_range=1):
...
 
+def ms_ssim(img1, img2, dynamic_range=1, power_factors=None):
+    r"""Multiscale structural similarity index (MS-SSIM)
+
+    As described in [1]_, multiscale structural similarity index (MS-SSIM) is
+    an improvement upon structural similarity index (SSIM) that takes into
+    account the perceptual distance between two images on different scales.
+
+    SSIM is based on three comparison measurements between the two images:
+    luminance, contrast, and structure. All of these are computed in windows
```
pehf, line 256: computed convolutionally

plenoptic/metric/perceptual_distance.py
```diff
@@ -250,6 +245,83 @@ def ssim_map(img1, img2, dynamic_range=1):
...
+
+    References
+    ----------
+    .. [1] Wang, Zhou, Eero P. Simoncelli, and Alan C. Bovik. "Multiscale
+       structural similarity for image quality assessment." The Thrity-Seventh
+       Asilomar Conference on Signals, Systems & Computers, 2003. Vol. 2. IEEE, 2003.
+
+    """
+    if power_factors is None:
+        power_factors = [0.0448, 0.2856, 0.3001, 0.2363, 0.1333]
```
pehf, line 307: these particular values have been calibrated for images where 5 scales can be computed, so I think it would be nice to assert the size of the input images passed to this function and make sure they are of the expected overall size. Images of shape 512×512 (even 256x256) or bigger should be fine, but computing SSIM on the smallest scale for images smaller than that will not be possible

pehf (commented):
minor suggestions, otherwise looks good to me

plenoptic/metric/perceptual_distance.py
Bichidian: The SSIM is supposed to use a 11x11 Gaussian convolutional window, but currently supports height and width smaller than 11 by setting the window size to `min(11, height, width)`. Through this, MS-SSIM also currently supports smaller images. I can add a warning for SSIM when height or width is smaller than 11, and a warning for MS-SSIM when the coarsest scale has height or width smaller than 11. Or I can consider these situations not worth supporting and raise an exception for these situations. What do you think?

lyndond: I support the adding of warning for small images for when e.g. log2(imsize)<5

lyndond (approved):
Looks fine to me apart from the potential warning thing w/ im shape.

plenoptic/metric/perceptual_distance.py
```diff
@@ -186,6 +181,10 @@ def ssim(img1, img2, weighted=False, dynamic_range=1):
     else:
         mssim = (map_ssim*weight).sum((-1, -2)) / weight.sum((-1, -2))
 
+    if min(img1.shape[2], img1.shape[3]) < 11:
```
billbrod, line 184: shouldn't this warning just be raised in `_ssim_parts`? It looks like it gets raised by all of `ssim`, `ssim_map`, and `ms_ssim`, so may as well reduce the repeating (also, raising it closer to where the relevant bit happens in the code makes it easier to read)

plenoptic/metric/perceptual_distance.py
```diff
@@ -247,9 +246,96 @@ def ssim_map(img1, img2, dynamic_range=1):
...
+    contrast map and structure map (the "contrast-structure map") is computed
+    for all but the coarsest scales, and the overall SSIM map is only computed
+    for the coarsest scale. Their mean values are raised to exponents and
+    multiplied to produce MS-SSIM:
+    .. math::
+        MSSSIM = {SSIM}_M^{a_M} \prod_{i=1}^{M-1} ({CS}_i)^{a_i}
+    Here :math: `M` is the number of scales, :math: `{CS}_i` is the mean value
+    of the contrast-structure map for the i'th finest scale, and :math: `{SSIM}_M`
+    is the mean value of the SSIM map for the coarsest scale. If at least one
+    of these terms are negative, the value of MS-SSIM is zero.
```
billbrod, line 280: if I understand correctly, the exponents `a_M` and `a_i` are the values of the `power_factors` arg? would be nice to make that clear.

plenoptic/metric/perceptual_distance.py
billbrod: When images are really large, do these numbers still make sense? For example, if the image is 2048 x 2048, we only compute the components on the first 5 scales, right? And so effectively ignore the coarsest info?

billbrod (changes requested):
This looks good to me, just a couple of small questions.

I like the changes you made to the docstrings and comments, I think that helps clarify them.

plenoptic/metric/perceptual_distance.py
Bichidian: The warnings for SSIM and MS-SSIM are different. I initially tried to raise a warning in `_ssim_parts`, and catch the warning and raise another warning in `ms_ssim`, but then I think it's cleaner to write them separately.

billbrod: ah I missed the difference between them. That makes sense to me then.

plenoptic/metric/perceptual_distance.py
Bichidian: I think it still makes sense. The way I think about this is that if a 512x512 image and a 2048x2048 image are displayed with the same visual angle to a viewer, then the scales that should be computed are the same.

plenoptic/metric/perceptual_distance.py
Bichidian: Yes, I will add this explanation.

plenoptic/metric/perceptual_distance.py
```diff
@@ -247,9 +246,97 @@ def ssim_map(img1, img2, dynamic_range=1):
...
+    for all but the coarsest scales, and the overall SSIM map is only computed
+    for the coarsest scale. Their mean values are raised to exponents and
+    multiplied to produce MS-SSIM:
+    .. math::
+        MSSSIM = {SSIM}_M^{a_M} \prod_{i=1}^{M-1} ({CS}_i)^{a_i}
+    Here :math: `M` is the number of scales, :math: `{CS}_i` is the mean value
+    of the contrast-structure map for the i'th finest scale, and :math: `{SSIM}_M`
+    is the mean value of the SSIM map for the coarsest scale. If at least one
+    of these terms are negative, the value of MS-SSIM is zero. The values of
+    :math: `a_i` are taken from the argument `power_factors`.
```
billbrod, line 281: does `a_M` also come from `power_factors`? And `M` and `i` are just two different indices?

plenoptic/metric/perceptual_distance.py
billbrod: sounds goodt o me. I'm going to mark this as resolved (since I think the warning you added covers @lyndond and @pehf 's comments, but they should re-open this if not)

billbrod (changes requested):

plenoptic/metric/perceptual_distance.py
Bichidian: Oh `M` is the number of scales, which is written just under the formula. It's just an index like `i`. You can see that `i` goes from `1` to `M-1` in the product. Do you think this should be made clearer?

billbrod: oh okay, so `a_M` is the final value in `power_factors`, and `a_i` are the values up to that. I think I'd just say `The values of :math: a_i and a_M are taken from power_factors` or the like, but this isn't important.

Bichidian: Ok I changed it to `a_i, i=1,...,M`

billbrod (approved):
All looks good to me!

plenoptic/metric/perceptual_distance.py
billbrod: works for me!

PR #125, opened 2021-05-03
hotfix for weird caching issue

billbrod:
The scheduled runs over the weekend failed while creating the environment with a strange pip error, the one seen [here](https://stackoverflow.com/questions/67273590/pip-21-1-cant-import-invalidschemecombination). As linked in that post, it's actually a [caching issue](https://github.com/pypa/pip/issues/9880), and so making sure not to cache pip solves the issue.


Bichidian (approved):

lyndond (approved):
looks good to me

PR #126, opened 2021-05-06
Perceptual metrics

Bichidian:
I worked on Laplacian pyramid and NLPD (normalized Laplacian pyramid distance) and think they are in good shape now. I still need to work on NSPD (normalized steerable pyramid distance), and create a tutorial for perceptual metrics in general.


plenoptic/metric/perceptual_distance.py
```diff
@@ -387,18 +390,19 @@ def nlpd(IM_1, IM_2):
...
 
     Returns
     -------
     distance: float
+        The normalized Laplacian Pyramid distance.
 
     Note
     ----
-    only accepts single channel images
+    The input images have to be single-channel.
```
pehf, line 405: We may be able to support batch and multichannel by cleaning our current LaplacianPyramid implementation, which would simplify the regression on perceptual data - do you think it is worth it?

tests/test_plenoptic.py
```diff
@@ -85,6 +86,20 @@ def ssim_analysis():
     return sio.loadmat(ssim_analysis, squeeze_me=True)
 
 
+@pytest.mark.parametrize('odd', [0, 1])
```
pehf, line 89: it would be nice to cover the multichannel case in our test

pehf (commented):
I suggested a few changes in my commits (recursive blur_downsample and renaming), feel free to keep what makes sense to you. Also I wonder if we really need to continue supporting all possible blurring filter from `pt.named_filters`, what do you think?

Bichidian:
> Also I wonder if we really need to continue supporting all possible blurring filter from `pt.named_filters`, what do you think?

I think it's not a burden. Are you asking this because many of the filters are not really blurring operations? If so, maybe we can limit to only the binomial ones.

plenoptic/metric/perceptual_distance.py
Bichidian: I think it is, will implement batch and multichannel.

plenoptic/metric/perceptual_distance.py
```diff
@@ -70,11 +71,13 @@ def _ssim_parts(img1, img2, dynamic_range, pad=False):
...
-    window = circular_gaussian2d(real_size, std, n_channels)
+    window = circular_gaussian2d(real_size, std=std)
 
     # these two checks are guaranteed with our above bits, but if we add
     # ability for users to set own window, they'll be necessary
-    if (window.sum((-1, -2)) > 1).any():
+    window_sum = window.sum((-1, -2))
+    if not torch.allclose(window_sum, torch.ones_like(window_sum)):
+        print(window.sum((-1, -2)))
         warnings.warn("window should have sum of 1! normalizing...")
```
billbrod, line 81: This `print` should be included in the warning (`warnings` and `print` get separately caught if the user wants to filter by them)

plenoptic/metric/perceptual_distance.py
```diff
@@ -56,9 +57,9 @@ def _ssim_parts(img1, img2, dynamic_range, pad=False):
                           f" img1: {img_ranges[0]}, img2: {img_ranges[1]}. "
                           "Continuing anyway...")
     (n_batches, n_channels, height, width) = img1.shape
```
billbrod, line 59: Should probably wrap this in a try/except so that it gives a helpful error message if `img1` is not 4d (currently, this returns `ValueError: not enough values to unpack (expected 4, got X)`)

In general, we should probably write a helper function somewhere that checks dimensionality of images, because we'll want to use it in a lot of places.

And also explicitly check that `img2` is 4d as well.

plenoptic/metric/perceptual_distance.py
```diff
@@ -56,9 +57,9 @@ def _ssim_parts(img1, img2, dynamic_range, pad=False):
...
-    if n_channels > 1:
-        warnings.warn("SSIM was developed on grayscale images, no guarantee "
-                      "it will make sense for more than one channel!")
+    if n_channels > 1 or img2.shape[1] > 1:
+        warnings.warn("SSIM was developed on grayscale images, so the channel dimension "
+                      "is treated as another batch dimension.")
     if img2.shape[-2:] != (height, width):
```
billbrod, line 63: do `img1` and `img2` have to have the same number of channels (if they don't have a single channel)? If I'm reading this correctly, they can either have the same number of batches (in which case they're compared batch-wise) or one of them can have a single batch -- do we want something similar for channels?

plenoptic/metric/perceptual_distance.py
```diff
@@ -350,14 +363,17 @@ def downsample(img):
 
 
 def normalized_laplacian_pyramid(im):
```
billbrod, line 365: Do we want to allow for user-specified number of scales?

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
 
     Parameters
     ----------
-    IM_1: torch.Tensor
-        image, (1 x 1 x H x W)
-    IM_2: torch.Tensor
-        image, (1 x 1 x H x W)
-
+    IM_1: torch.Tensor of shape (batch, channel, height, width)
```
billbrod, line 418: doesn't have to happen in this PR, but we need to go through all these argument names and make them consistent: we've got `img1`, `im_1` and `IM_1` all in this file.

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
...
-        dist.append(torch.sqrt(torch.mean((y[i][0] - y[i][1]) ** 2) + epsilon))
+        dist.append(torch.sqrt(torch.mean((y1[i] - y2[i]) ** 2, dim=(2, 3)) + epsilon))
 
-    return torch.stack(dist).mean()
+    return torch.stack(dist).mean(dim=0)
 
 
-def nspd(IM_1, IM_2, O=1, S=5, complex=True):
+def normalized_steerable_pyramid(im, sigmas=None):
+    """Compute the normalized Steerable Pyramid using pre-optimized parameters. (under construction)
```
billbrod, line 448: Should specify what we mean by "Under construction"

I vote we add something in `Notes` that says something along the lines of "this works but is currently not a good human perceptual metric. more work optimizing the weights (or whatever they're called) is necessary."

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
...
+
+    Arguments
+    --------
+    im: torch.Tensor of shape (batch, channel, height, width)
+        Image, or batch of images. Channels are also treated as batches.
```
billbrod, line 453: Wording of "Channels are also treated as batches" strikes me as a bit unclear. I'd say something like "computed independently for each batch and channel" (because that's what it does, right?)

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
 
     Parameters
```
billbrod, line 416: for all these distances, we should specify how they handle multi-batch / channel images: if the two have the same number, they're compared one-to-one, and otherwise the one with a single batch/channel is compared to all the batch/channels in the image with multiple batch/channels. Any other situation will throw an error.

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
...
+    --------
+    im: torch.Tensor of shape (batch, channel, height, width)
+        Image, or batch of images. Channels are also treated as batches.
+
+    Returns
+    -------
+    normalized_spyr_coeffs: dict of torch.Tensor
+        The normalized Steerable Pyramid with 5 scales and 4 orientations. The keys
+        of this dictionary are the same as the output of `Steerable_Pyramid_Freq`.
+    """
```
billbrod, line 460: Since you changed `Laplacian_Pyramid` -> `LaplacianPyramid` (which is good!), we should do the same for `Steerable_Pyramid_Freq`

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
...
 
-def nspd(IM_1, IM_2, O=1, S=5, complex=True):
+def normalized_steerable_pyramid(im, sigmas=None):
+    """Compute the normalized Steerable Pyramid using pre-optimized parameters. (under construction)
+
+    Arguments
+    --------
+    im: torch.Tensor of shape (batch, channel, height, width)
+        Image, or batch of images. Channels are also treated as batches.
+
```
billbrod, line 454: Missing documentation for `sigmas`

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
...
+    Returns
+    -------
+    normalized_spyr_coeffs: dict of torch.Tensor
+        The normalized Steerable Pyramid with 5 scales and 4 orientations. The keys
+        of this dictionary are the same as the output of `Steerable_Pyramid_Freq`.
+    """
+
+    if sigmas is None:
+        sigmas = pickle.load(open(dirname + "/nspd_sigmas.pickle", mode="rb"))
```
billbrod, line 463: best way to construct the pathname is with `os.path.join`, which works across all OSs

plenoptic/metric/perceptual_distance.py
```diff
@@ -3,12 +3,13 @@
...
-from ..simulate.canonical_computations import local_gain_control_dict, rectangular_to_polar_dict
+from ..simulate.canonical_computations import LaplacianPyramid, Steerable_Pyramid_Freq
 from ..simulate.canonical_computations.filters import circular_gaussian2d
+from ..simulate.canonical_computations.non_linearities import local_gain_control
 from ..tools.conv import same_padding
 
 import os
+import pickle
 
 dirname = os.path.dirname(__file__)
```
billbrod, line 14: `dirname` should be `DIRNAME`, since it's a constant.

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
...
+        sigmas = pickle.load(open(dirname + "/nspd_sigmas.pickle", mode="rb"))
+    spyr = Steerable_Pyramid_Freq(im.shape[-2:], height=5, order=3, is_complex=False, downsample=True).to(im.device)
+    spyr_coeffs = spyr.forward(im)
+    normalized_spyr_coeffs = {}
+    for key in spyr_coeffs.keys():
+        if key == "residual_lowpass":
+            normalized_spyr_coeffs[key] = spyr_coeffs[key]  # No normalization for residual_lowpass
+        elif key == "residual_highpass":
+            continue
+        else:
```
billbrod, line 472: can we add some comments explaining why we handle the residual highpass and lowpass differently?

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
...
+            _, normalized_spyr_coeffs[key] = local_gain_control(spyr_coeffs[key], epsilon=sigmas[key])
+    return normalized_spyr_coeffs
```
billbrod, line 474: why are they called `sigmas` in this function and `epsilon` in `local_gain_control`?

plenoptic/metric/perceptual_distance.py
```diff
@@ -399,62 +415,86 @@ def nlpd(IM_1, IM_2):
...
 
-    return torch.stack(dist).mean()
+    return torch.stack(dist).mean(dim=0)
 
 
-def nspd(IM_1, IM_2, O=1, S=5, complex=True):
+def normalized_steerable_pyramid(im, sigmas=None):
+    """Compute the normalized Steerable Pyramid using pre-optimized parameters. (under construction)
+
+    Arguments
```
billbrod, line 450: Similar to normalized Laplacian, do we want to allow user-specified height and order?

plenoptic/simulate/canonical_computations/laplacian_pyramid.py
```diff
@@ -47,89 +49,37 @@ def analysis(self, x):
...
-#     def synthesis(self, y):
-#
-#         x = y[self.n_scales-1]
-#         for scale in range(self.n_scales-1, 0, -1):
-#             y_up = self.upsample_convolve(x) #, output_size=y[scale-1].shape)
-#             x = y[scale-1] + y_up
-#
-#         return x
+    def forward(self, x):
+        return self.analysis(x)
```
billbrod, line 85: I feel like we should have the method names here be consistent with the steerable pyramid -- so no analysis and synthesis, but just `forward` and `recon_pyr`

plenoptic/tools/conv.py
```diff
@@ -1,143 +1,132 @@
...
-    Returns:
-        [type]: [description]
+    Parameters
+    ----------
+    image: torch.Tensor of shape (batch, channel, height, width)
+        Image, or batch of images. Channels are also treated as batches.
+    filt: 2-D torch.Tensor
+        The filter to correlate with the input image
+    padding_mode: string, optional
+        One of "constant", "reflect", "replicate", "circular" or "zero" (same as "constant")
```
billbrod, line 19: why have `"zero"` as an option if it's redundant with `"constant"`?

tests/test_metric.py
```diff
@@ -125,32 +128,33 @@ def test_msssim(self, einstein_img, curie_img):
         curie_img.requires_grad_()
         assert po.metric.ms_ssim(einstein_img, curie_img).requires_grad
 
-    @pytest.mark.parametrize('func_name', ['noise', 'mse', 'ssim', 'ms-ssim'])
-    @pytest.mark.parametrize('size_A', [1, 3])
-    @pytest.mark.parametrize('size_B', [1, 2, 3])
+    @pytest.mark.parametrize('func_name', ['mse', 'ssim', 'ms-ssim', 'nlpd', 'nspd'])
+    @pytest.mark.parametrize('size_A', [(1, 1), (5, 3), (5, 3), (5, 1), (5, 1), (5, 3)])
+    @pytest.mark.parametrize('size_B', [(1, 1), (5, 3), (1, 1), (1, 3), (3, 1), (5, 2)])
```
billbrod, line 133: we have no even batch values in these, should probably include one

tests/test_metric.py
```diff
@@ -125,32 +128,33 @@ def test_msssim(self, einstein_img, curie_img):
         curie_img.requires_grad_()
         assert po.metric.ms_ssim(einstein_img, curie_img).requires_grad
 
-    @pytest.mark.parametrize('func_name', ['noise', 'mse', 'ssim', 'ms-ssim'])
-    @pytest.mark.parametrize('size_A', [1, 3])
-    @pytest.mark.parametrize('size_B', [1, 2, 3])
+    @pytest.mark.parametrize('func_name', ['mse', 'ssim', 'ms-ssim', 'nlpd', 'nspd'])
```
billbrod, line 131: the test for `po.tools.add_nosie` was removed -- why?

tests/test_metric.py
```diff
@@ -216,16 +220,6 @@ def test_nspd(self, einstein_img, curie_img):
...
-    def test_nspd2(self, einstein_img, curie_img):
-        curie_img.requires_grad_()
-        assert po.metric.nspd(einstein_img, curie_img, O=3, S=5, complex=True).requires_grad
-        curie_img.requires_grad_(False)
-
-    def test_nspd3(self, einstein_img, curie_img):
-        curie_img.requires_grad_()
-        assert po.metric.nspd(einstein_img, curie_img, O=1, S=5, complex=False).requires_grad
-        curie_img.requires_grad_(False)
-
```
billbrod, line 228: why remove these two and not `test_nspd`? I thought these were removed because `test_batch_handling` handles the testing for `nspd`, but then why leave `test_nspd` in?

tests/test_models.py
```diff
@@ -1,20 +1,20 @@
-#/usr/bin/env python3
+# /usr/bin/env python3
```
billbrod, line 1: we should make this line consistent across all tests -- either remove it or correct it (to `#!/usr/bin/env python3`) (if the script is executable and you just run `./test_models.py`, the `#!` line tells your shell which python to use -- not sure it's necessary)

tests/test_models.py
```diff
@@ -1,20 +1,20 @@
...
 import torch
-import os.path as op
-from test_metric import osf_download
-from plenoptic.simulate.canonical_computations import (gaussian1d, circular_gaussian2d)
-from conftest import DEVICE, DATA_DIR
+import numpy as np
+import pytest
+import matplotlib.pyplot as plt
+import os
+os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'
```
billbrod, line 17: what is this for? I forget, could we add a comment?

billbrod, line 17: what is this for? I forget, could we add a comment?

tests/test_models.py
```diff
@@ -73,6 +73,35 @@ def test_local_gain_control_dict(self, basic_stim):
...
+        img = curie_img[0:253, 0:234]
+        lpyr_po = po.simul.LaplacianPyramid(n_scales=n_scales).to(DEVICE)
+        y_po = lpyr_po(img)
+        lpyr_pt = pt.pyramids.LaplacianPyramid(img.squeeze().cpu(), height=n_scales)
+        y_pt = [lpyr_pt.pyr_coeffs[(i, 0)] for i in range(n_scales)]
+        assert len(y_po) == len(y_pt)
+        for x_po, x_pt in zip(y_po, y_pt):
+            x_po = x_po.squeeze().detach().cpu().numpy()
+            assert np.abs(x_po - x_pt)[:-2, :-2].max() < 1e-5
+            # TODO: There is some problem with padding on right and bottom edge, should figure out why
```
billbrod, line 102: if this is still a problem, make an issue for it

billbrod (changes requested):
I've dropped a variety of notes throughout this, but I'm really pleased to see so much of the code cleaned up in this.

Couple bigger points:

- `conv.py` went through a pretty substantial overhaul. In particular, `step` was removed from all the functions, and `odd` has been added. Can you discuss why those changes were made some more?
- Additionally, several of the functions have had `filtname` removed and now only work with binomial filters. Why?
- For the `nspd_sigmas.pickle`: how many values is this? the way we handle the constants for `frontend.py` is not to have the values in the script itself (see `_pretrained_state_dict()`), and I think we should do similarly for this, assuming it's also a small number of values (because it makes the diff useful)

plenoptic/metric/perceptual_distance.py
Bichidian: I think maybe I should create a new `nspd` branch from the current head of `perceptual-metrics` (this branch), and remove nspd completely from `perceptual-metrics` before merging it into `main`. I think people should not want to use the unfinished nspd.

tests/test_models.py
Bichidian: I don't know what this is. These changes are just my editor automatically rearranging the imports.

tests/test_models.py
Bichidian: Shouldn't it be `pytest test_models.py` instead of `python test_models.py` anyway? I don't see the point of specifying the executable.

plenoptic/metric/perceptual_distance.py
Bichidian: The normalization weights are learned for the Laplacian pyramid with 6 scales, so it has to be 6 scales.

plenoptic/metric/perceptual_distance.py
Bichidian: I think the `print` statement was used for debugging and left here by mistake. I'll just delete it.

plenoptic/metric/perceptual_distance.py
Bichidian: Yes, both batch and channel dimensions follow the broadcast rule.

plenoptic/metric/perceptual_distance.py
Bichidian: Yes that's better.

plenoptic/tools/conv.py
Bichidian: When the function `F.pad` takes the argument `mode="constant"` without specifying the value of the constant, the padded constant is default to 0. I added the option of `"zero"` because I thought it was clearer than `"constant"`, but still kept the `"constant"` option.

tests/test_metric.py
Bichidian: I assumed the purpose of these tests was to make sure the gradient works. Maybe I can change the names of these tests to things like `test_nlpd_grad`.

plenoptic/tools/conv.py
```diff
@@ -1,143 +1,132 @@
...
+    x: torch.Tensor of shape (batch, channel, height, width)
+        Image, or batch of images. Channels are also treated as batches.
+    n_scales: int, optional
+        Apply the blur and downsample procedure recursively `n_scales` times.
+    order_plus_one: int, optional
+        One plus the order of the binomial coefficient filter. Must be at least 2.
+        The 2D blurring filter is obtained by computing this 1D filter's outer
+        product with itself, and has shape (order_plus_one, order_plus_one).
+    scale_filter: bool, optional
+        If true (default), the filter sums to 1 (ie. it does not affect the DC
```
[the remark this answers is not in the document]
Bichidian: Yes, it is the same as `"binomN"` in pyrtools `named_filter`. The size of the filter is actually `order + 1`.

tests/test_metric.py
Bichidian: The reason I removed it was probably because it is a small function and the batch handling is simple (like as simple as addition), so the testing it together with `ssim`, `ms-ssim`, `nlpd` and `nspd` was not convenient. In fact I'm also going to remove `mse` from this test for convenience. So I think I can write another function to test for `noise` and `mse`.

tests/test_models.py
Bichidian: Yeah I haven't figured it out yet, will make an issue.

plenoptic/metric/perceptual_distance.py
Bichidian: True, most functions need to check the dimensionality of images. I think for this PR I will not do a systematic fix.

tests/test_models.py
Bichidian: I have found the difference in boundary handling. The pyrtools implementation `pt.upConv` performs padding after upsampling. My implementation here `po.tools.upsample_convolve` performs padding before upsampling, and, depending on the parity of the image, sometimes performs additional zero padding after upsampling up to one row/column. The implementation of pyrtools sometimes produces inconsistency: for example, it sometimes effectively performs "reflect2" padding when asked to perform "reflect1" padding due to zeros introduced in upsampling. I have found a way to let my code match pyrtools, but I think my current implementation is better, so I will not change it. I will replace this TODO message in the test file with an explanatory comment.

plenoptic/metric/perceptual_distance.py
```diff
@@ -430,11 +449,22 @@ def nlpd(IM_1, IM_2):
...
+    for i in range(2):
+        if img1.shape[i] != img2.shape[i] and img1.shape[i] != 1 and img2.shape[i] != 1:
+            raise Exception("Either img1 and img2 should have the same number of "
+                            "elements in each dimension, or one of "
+                            "them should be 1! But got shapes "
+                            f"{img1.shape}, {img2.shape} instead")
+    if img1.shape[1] > 1 or img2.shape[1] > 1:
+        warnings.warn("NLPD was designed for grayscale images and here it will be computed separately for each "
+                      "channel (so channels are treated in the same way as batches).")
+
```
billbrod, line 465: given that this is a check we do regularly, might be worth writing a helper function for it. doesn't necessarily need to be in this PR though

plenoptic/metric/perceptual_distance.py
billbrod: That's fine with me, would be curious to hear what @pehf thinks (not sure if anyone else has thought about using this)

tests/test_metric.py
```diff
@@ -131,33 +131,37 @@ def test_msssim(self, einstein_img, curie_img):
...
+                else:
+                    tgt_size = None
+                    break
+            if tgt_size is None:
+                with pytest.raises(Exception, match="Either img1 and img2 should have the same number of "
+                                                    "elements in each dimension, or one of them should be 1"):
+                    func(A, B)
+            elif tgt_size[1] > 1:
+                with pytest.warns(Warning, match="computed separately for each channel"):
+                    assert func(A, B).shape == tuple(tgt_size)
```
billbrod, line 162: oh I like this `pytest.warns` / `pytest.raises` with `match=` syntax

plenoptic/metric/perceptual_distance.py
billbrod: Sounds good

plenoptic/tools/conv.py
billbrod: So `"zero"` must be the same `"constant"` here because we can't pass any additional arguments to `F.pad` via `same_padding`.

I'm not a big fan of this -- since we're basically passing the arguments directly to `F.pad`, I think we should keep the arguments the same. So I'd prefer we remove `"zero"` and note that `"constant"` will zero-pad. I think we should be consistent with pytorch, basically.

plenoptic/tools/conv.py
```diff
@@ -1,143 +1,132 @@
...
-        filtname (str, optional): [description]. Defaults to 'binom5'.
-        step (tuple, optional): [description]. Defaults to (2, 2).
-
-    Returns:
-        [type]: [description]
+    if padding_mode == "zero":
+        padding_mode = "constant"
+    assert isinstance(image, torch.Tensor) and isinstance(filt, torch.Tensor)
+
+    n_channels = image.shape[1]
```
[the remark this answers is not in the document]
billbrod: still feel this way

tests/test_metric.py
billbrod: Yes, I think that would greatly clarify things

tests/test_models.py
billbrod: I agree, that's correct. I'll remove these.

tests/test_models.py
billbrod: gotcha. we should probably have a PR (probably as part of running python black?) that standardizes the imports.

tests/test_models.py
billbrod: Gotcha, yes please make those changes. It sounds like this is a pyrtools bug then? can you create an issue there?

billbrod (changes requested):
This looks good! 

I'm still confused as to why you replaced some `named_filter` with just the ability to use binomial filters -- why's that?

Beyond that, there are three changes you mentioned making that I don't see yet:
1. moving the nspd over to its own branch and removing it from this
2. renaming some of the tests in `test_metric.py` to `test_nlpd_grad`, etc
3. replacing the TODO in `test_models.py` with a descriptive comment

Once you've done those and answered my question, I'm good to approve this

billbrod:
Also I just pushed a change removing `#!` from the top of the test files. I'm pretty sure that's the fault of my editor.

plenoptic/metric/perceptual_distance.py
billbrod: if you decide not to put it in this PR, can you create an issue for this? and paste your code in there.

billbrod:
Oh also, `_ssim_parts` docstring should include what it returns.

billbrod:
I just pushed a small set of changes to fix some of the docstrings: correcting underline length and section names, similar to #178 but I put them here because I thought the merge might become a headache otherwise

plenoptic/tools/conv.py
Bichidian: Ok, will do.

plenoptic/tools/conv.py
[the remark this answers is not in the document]
Bichidian: Ok

tests/test_models.py
Bichidian: I think it's not necessarily a bug in prytools, it's just a slightly different boundary handling logic and doesn't make a meaningful difference.

Bichidian:
<p>I found it very difficult to write a cross-platform code to unzip TID2013, so I decided to let the user to do the whole download-unzip process manually. The code I wrote was for testing and only works on Linux system.</p>

---

 <a href='https://app.reviewnb.com/LabForComputationalVision/plenoptic/pull/126/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Bichidian:
<p>The NLPD paper uses "DMOS" which should be a decreasing function of MOS, but somehow I cannot find its precise definition. I will explain MOS better.</p>

---

 <a href='https://app.reviewnb.com/LabForComputationalVision/plenoptic/pull/126/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

plenoptic/metric/perceptual_distance.py
billbrod: created issue in #185

billbrod (approved):
These look good to me!

pehf:
This looks good to me. Here are some minor suggestions:
- in `blurDn`, we use a recursion for blurring images multiple times and this code could be simplified to be more readable and intuitive.
- in the perceptual distance notebook, it might be worth adding a sentence at the top describing why one might be interested in perceptual distances (at the moment this is done in the paragraph on 'understanding SSIM' which comes after the technical description). Also, it would be nice to make sure that the reported MSE values match the range of images, which is 0-1 throughout.

pehf (approved):

PR #131, opened 2021-05-27
Fix docs issue

billbrod:
We shouldn't pin the pip version. see discussion here: https://github.com/spatialaudio/nbsphinx/issues/563

It was an issue with Jinja2 that got fixed upstream, but by pinning pip we were having issues getting the recent version with the fix.


lyndond (approved):

pehf (approved):
lgtm

PR #136, opened 2021-08-02
Synthesis refactor

billbrod:
This PR is the refactor of `Synthesis`, `Metamer`, and `MADCompetition`, to make them simpler and easier to understand.

This PR involves:
- stripping out a lot of the unnecessary code from `Metamer` and `MADCompetition` to make them easier to understand, and reducing their attributes.
- `Synthesis` has been reduced to a very small class, with only the things we *think* it shares across methods. That's just `save`, `load`, and `to`, plus requiring an implementation of `synthesis`.
- adds `SimpleMetamer`, a very straight-forward metamer method to demonstrate the basics of synthesis (without the bells and whistles that `Metamer` has for coarse-to-fine and storing progress to allow for animating) with an associated notebook `testing-new-synthesis`. That notebook should either be reworked (if we think it's useful to have around) or deleted by merge.
- The synthesis plotting / animating methods have all been moved into `plenoptic/synthesize/metamer.py` as individual functions (not methods) that take a `Metamer` object as their first instance, which cleans up its namespace greatly (and analogously for `MADCompetition`)


plenoptic/synthesize/metamer.py
```diff
@@ -103,126 +93,503 @@ class Metamer(Synthesis):
...
+        return loss, grad_norm, self.optimizer.param_groups[0]['lr'], pixel_change
+
+    def synthesize(self, max_iter: int = 100,
+                   optimizer: Union[None, torch.optim.Optimizer] = None,
+                   scheduler: Union[None, torch.optim.lr_scheduler._LRScheduler] = None,
+                   store_progress: Union[bool, int] = False,
+                   stop_criterion: float = 1e-4, stop_iters_to_check: int = 50,
+                   coarse_to_fine: Literal['together', 'separate', False] = False,
+                   coarse_to_fine_kwargs: Dict[str, float] = {'change_scale_criterion': 1e-2,
+                                                              'ctf_iters_to_check': 50}):
```
lyndond, line 557: returns `--> torch.Tensor`

lyndond (approved):
This structure is a great improvement from before! I think it's fine to merge as-is.

nikparth (approved):
Looks good to me. As far as notebooks. I made some changes to the testing-new-synthesis.ipynb that I think show a nice simple demonstration with the simple_metamer class. I feel like this may be a good thing to include somewhere in the official tutorials especially for people coming from the deep learning community and just wanting something stripped down and simple to work with. But that can be done in another PR if we decide to.

PR #139, opened 2021-11-04
Flatiron Jenkins continuous integration

lyndond:
With Dylan's help, we are setting up GPU testing w/ Jenkins at Flatiron Inst. 

- All pytest files run green on both gpu and cpu


jenkins/Jenkinsfile
```diff
@@ -0,0 +1,53 @@
...
+        HOME = pwd(tmp:true)
+	OMP_NUM_THREADS = 4
+      }
+      steps {
+	sh 'python3 -c "import torch; torch.cuda.current_device()"'
+	sh 'python3 -m venv --system-site-packages $HOME'
+	sh '''#!/bin/bash -ex
+	  source $HOME/bin/activate
+	  pip3 install .
+      for f in tests/test_*py; do python3 -m pytest $f; done
```
billbrod, line 26: probably worth adding a comment here to say why we do it this way instead of the obvious `python3 -m pytest tests`

plenoptic/metric/perceptual_distance.py
```diff
@@ -66,7 +66,7 @@ def _ssim_parts(img1, img2, dynamic_range):
 
     real_size = min(11, height, width)
     std = torch.tensor(1.5).to(img1.device)
-    window = circular_gaussian2d(real_size, std=std, out_channels=n_channels)
+    window = circular_gaussian2d(real_size, std, n_channels, img1.device)
```
billbrod, line 69: is adding a device arg better than calling `.to(device)` on the resulting tensor?

billbrod: especially because it doesn't look like we're consistent: `models.naive.Gaussian` doesn't accept a device arg. Is the idea that models don't accept a device arg, but components can?

plenoptic/simulate/models/naive.py
```diff
@@ -136,7 +136,9 @@ def filt(self):
         if self._filt is not None:  # use old filter
             return self._filt
         else:  # create new filter, optionally cache it
-            filt = circular_gaussian2d(self.kernel_size, self.std, self.out_channels)
+            device = self.std.device
+            filt = circular_gaussian2d(self.kernel_size, self.std, self.out_channels, device)
+            filt = filt.to(device)
```
billbrod, line 141: is the `filt.to(device)` necessary with the device arg passed to `circular_gaussian2d`?

plenoptic/simulate/models/naive.py
```diff
@@ -235,27 +237,28 @@ def __init__(
 
         self.cache_filt = cache_filt
         self._filt = None
+        self._input_device = torch.device("cpu")
```
billbrod, line 240: I think this is unused, yeah? in which case, should delete it

plenoptic/simulate/models/naive.py
```diff
@@ -235,27 +237,28 @@ def __init__(
...
+            filt_surround = circular_gaussian2d(self.kernel_size, self.surround_std, self.out_channels, device)
 
             # sign is + or - depending on center is on or off
-            sign = torch.as_tensor([1. if x else -1. for x in self.on_center])
+            sign = torch.as_tensor([1. if x else -1. for x in self.on_center]).to(device)
             sign = sign.view(self.out_channels, 1, 1, 1)
-            filt = on_amp * (sign * (filt_center - filt_surround))
 
-            filt = filt / filt.sum()
+            filt = on_amp * (sign * (filt_center - filt_surround))
```
billbrod, line 258: why did we get rid of  the normalization of the filter?

plenoptic/synthesize/eigendistortion.py
```diff
@@ -282,7 +282,7 @@ def _synthesize_exact(self) -> Tuple[Tensor, Tensor]:
 
         J = self.compute_jacobian()
         F = J.T @ J
-        eig_vals, eig_vecs = torch.symeig(F, eigenvectors=True)
+        eig_vals, eig_vecs = torch.linalg.eigh(F, UPLO="U")
```
billbrod, line 285: isn't `torch.linalg` only available starting in `torch` 1.10?

billbrod: Also, this seems unrelated to the Jenkins / GPU tests issue, and so probably shouldn't be part of this PR?

tests/test_eigendistortion.py
```diff
@@ -139,17 +139,16 @@ def test_display(self, model, einstein_img, color_img, method, k):
...
 class TestAutodiffFunctions:
 
     @pytest.fixture(scope='class')
     def state(self, einstein_img):
         """variables to be reused across tests in this class"""
-        torch.manual_seed(0)
 
         k = 2  # num vectors with which to compute vjp, jvp, Fv
-        einstein_img = einstein_img[..., :16, :16]  # reduce image size
+        einstein_img = einstein_img[..., 100:100+16, 100:100+16]  # reduce image size
```
billbrod, line 151: why grab a different part of the image?

tests/test_eigendistortion.py
```diff
@@ -193,16 +192,20 @@ def test_jac_vec_prod(self, state):
     def test_fisher_vec_prod(self, state):
         x, y, x_dim, y_dim, k = state
 
-        V = torch.randn((x_dim, k), device=DEVICE)
+        V, _ = torch.linalg.qr(torch.ones((x_dim, k), device=DEVICE), "reduced")
+        U = V.clone()
```
billbrod, line 196: - same comment about `torch.linalg` and torch 1.10
- why use the cloned version?

billbrod (changes requested):
I've got a couple of small questions to address here:
- when should device be an argument vs should we call `.to(device)`
- you used `torch.linalg` in a couple of places, but I thought that wasn't supported until 1.10
- there are some changes in eigendistortions.py that look unrelated to this PR

plenoptic/synthesize/eigendistortion.py
nikparth: I believe torch.linalg.eigh is still there in torch 1.8 which we require so I don't think this is an issue

tests/test_eigendistortion.py
```diff
@@ -193,16 +192,20 @@ def test_jac_vec_prod(self, state):
...
 
-        Fv2 = jac.T @ jac @ V  # manually compute product to compare accuracy
+        Fv2 = jac.T @ jac @ U  # manually compute product to compare accuracy
 
         assert Fv.shape == (x_dim, k)
-        assert Fv2.allclose(Fv, rtol=1E-2)
+        if DEVICE != torch.device("cuda"):
+            assert Fv2.allclose(Fv, rtol=1E-2) 
+        else:
+            assert Fv2.allclose(Fv, atol=1E-6)
```
nikparth, line 208: why check rtol for not cuda and atol for cuda? is there not a way to keep these the same just for clarity?

tests/test_metric.py
```diff
@@ -197,20 +198,25 @@ def test_msssim_analysis(self, msssim_images):
     def test_nlpd(self, einstein_img, curie_img):
         curie_img.requires_grad_()
         assert po.metric.nlpd(einstein_img, curie_img).requires_grad
+        curie_img.requires_grad_(False)
 
     def test_nspd(self, einstein_img, curie_img):
         curie_img.requires_grad_()
         assert po.metric.nspd(einstein_img, curie_img).requires_grad
+        curie_img.requires_grad_(False)
```
nikparth, line 206: why does the requires_grad need to be set to False after setting it to True first?

nikparth (approved):
had a couple minor questions but nothing really important. Seems good to go otherwise.

jenkins/Jenkinsfile
billbrod: back to `pytest tests`, so this isn't relevant anymore

plenoptic/synthesize/eigendistortion.py
lyndond: yeah as nikhil said, it works as of 1.8 so there's no harm

tests/test_eigendistortion.py
lyndond: There wasn't much going on in the top corner of the image. Wanted to use a patch with more variation so that the eigenvectors would be better behaved.

plenoptic/simulate/models/naive.py
lyndond: done

plenoptic/simulate/models/naive.py
lyndond: Good catch. No it is not. Removed.

plenoptic/simulate/models/naive.py
lyndond: It didn't really make sense bc the filter contains both pos and neg values, so it isn't clear what dividing by its sum would achieve.

tests/test_metric.py
lyndond: It's for juggling w/ pytest fixtures. `curie_img` is used in a bunch of tests instead of instantiating new data each test.

tests/test_eigendistortion.py
lyndond: - torch.linalg intro'd in 1.8.
- cloned bc i'm checking against values and didn't want to accidentally refer to same underlying object

plenoptic/metric/perceptual_distance.py
lyndond: removed device arg in favour of better device inference within the `circular_gaussian2d` function

tests/test_eigendistortion.py
lyndond: made them both atol for clarity

billbrod (approved):

PR #141, opened 2021-11-10
replace treebeard with jupyter

billbrod:
regular jupyter now includes [jupyter execute](https://twitter.com/palewire/status/1458083565191655424), so we can replace treebeard in our tests with this, which will be better supported. We also now automatically test all `examples/*ipynb` notebooks and all `tests/test*py` scripts (rather than having to manually indicate which we wanted to test). Updates `CONTRIBUTING.md` to reflect those changes.


billbrod:
Failing check at this point is `Demo_eigendistortion`, which is the result of an incomplete merge (it has `<<<<< HEAD`, etc scattered throughout it). Once that's fixed, I'll change this to a regular PR.

lyndond (approved):
looks dandy

billbrod:
The most recent pushes I made should fix the issue we've had periodically where the name of the tests change and so the PR requires old tests that no longer exist and then I have to go manually change the name of the test. Now, there's a test called `check` which will only complete if everything else (test scripts and notebooks) runs successfully (using [this](https://github.com/re-actors/alls-green)). This can then be the only thing we check. So, one more change to the branch protection rules are necessary to set this up, but that should be the last one.

nikparth (approved):
lgtm. I didn't do a deep dive so if there are any specific concerns you wanted me to go through in detail then lmk otherwise all seems fine.

PR #152, opened 2021-12-16
Fix failing python 3.7 tests

billbrod:
following Dylan's advice to remove the cache step (and replace it with caching from the `setup-python` action) in order to fix the failing python 3.7 tests. based on the error we were getting, he thought it likely that it was caused by an old version of `setuptools` and that the path list in our cache step looked questionable.


nikparth (approved):
lgtm

lyndond (approved):
looks a-ok to me

PR #153, opened 2021-12-16
MADCompetition fixes

### LINKED ISSUE #147: MADCompetition.to() doesn't move initial_image
Change that

billbrod:
- `Synthesis.load` now handles `map_location` correctly, so you can save on a GPU and load on a CPU
 - `Synthesis.load` doesn't assume the loss function(s) it checks act on single channel images, instead checking them on images shaped like the saved ones.
 - `MADCompetition.to()` moves `initial_signal` as well, because it's weird not to (this wasn't actually breaking synthesis, but it was causing problems if you want to e.g., visualize afterwards)
 - Adds tests for above changes
 - `SimpleMAD` notebook has been updated. This is the most substantial change: I set the seed and synthesis args so the two-pixel version runs to completion, added more text explaining why the two-pixel results look the way they do, and added a section running L1 vs L2 on a checkerboard image, explaining why its results look the way they do as well.

Closes #147, #148, #149, #150

This requires #139 to be merged first, since several new tests require GPUs.


plenoptic/synthesize/synthesis.py
```diff
@@ -128,8 +129,15 @@ def load(self, file_path: str,
                                     f" Self: {getattr(self, k)}, "
                                     f"Saved: {tmp_dict[k]}")
         for k in check_loss_functions:
-            # this way, each is a 1x1x64x64 tensor
-            tensor_a, tensor_b = torch.rand(2, 1, 1, 64, 64)
+            # want to find an attribute that looks like our input, but we don't
+            # know what they're named, so we find one that's a tensor and has 4
+            # dimensions.
+            for attr in check_attributes:
+                if isinstance(getattr(self, attr), torch.Tensor) and getattr(self, attr).ndim == 4:
```
nikparth, line 136: im not totally sure what's going on here? why do we not know what the attributes are named? Shouldn't that be standardized?  This seems complicated and also a bit strange because is it not possible that there isn't another 4-d tensor ever in the attributes?

nikparth (commented):
everything looks good to me but had a question about that one change in the synthesis class regarding checking the input tensor... Just want to understand that before approving.

plenoptic/synthesize/synthesis.py
billbrod: We don't know what they're named because we decided to have the attributes be less consistently but more transparently named (#135). That is, the synthesized images that we're outputting should be named `metamer`, `eigendistortion`, `geodesic`, `mad_image` instead of a consistent `synthesized_signal`, since it's more obvious what they are then.

But I think you're right that it's complicated and possible that there's not a 4d tensor attribute (in particular, I'm not sure about geodesic). Do you have another solution? I'm just looking for the properly shaped tensor that the loss function can accept. Originally I was just assuming they always had 1 channel, but that failed for MAD Competition between something and a VGG16-based loss, since VGG16 requires a 3-channel input.

lyndond (approved):
Agree w/ nikhil that the loading thing is a bit strange but it's nbd

plenoptic/synthesize/synthesis.py
nikparth: It looks like in the load, you're only using that img_attr to get it's shape correct? Why not save the image shape as an attribute that has the same name across classes like "synthesized_shape" or something and then use that?  Cuz that'll just be a tuple so it's no issue to carry around. Or am I missing something and are you actually using the "img_attr" somewhere else?

billbrod: oh yeah, that makes a lot of sense. I'll do that.

billbrod: Fixed this, but I ended up doing something similar to get the device, so... should I add a `device` attribute as well? I'd rather avoid holding onto too many attributes. In particular, I'm worried about this attribute getting out of sync -- it's possible that the individual tensor attributes can change devices independently. I don't think that's good, but right now we do nothing to stop it

nikparth: Resolved in slack conversation. Current solution seems fine.

nikparth (approved):
Everything looks good to me.

PR #163, opened 2022-07-14
Jenkins

billbrod:
For the weekly GPU job on `main`, run `pip install -U .` to upgrade any dependencies. This matches how our cpu tests work, grabbing the most recent versions of our dependencies (rather than relying on cached dependencies)


billbrod:
As part of this PR, we're dropping support for torch version `1.12.0`, owing to issue #162. that bug should be fixed in subsequent releases.

Bichidian (approved):

lyndond (approved):
LGTM

PR #174, opened 2022-08-26
Type hinting & Python Black

lyndond:
These changes are entirely cosmetic and have no functionality changes wrt the codebase. I mainly ran Black formatting on `tools` and `eigendistortion`, and manually added typehinting + edited docstrings in tools.


kbonnen (approved):
Wow.  I'm curious what the percentage of lines that changed with these edits (that changed no function).

Looks good to me.  Have we decided as a group that this is the formatting we want everything to have?  Python black + type hinting?

lyndond:
@kbonnen Thanks for the review. You're right in that no functionality changes were made -- just cosmetic. 

Python Black is certainly a highly opinionated formatter, but v straightfwd way to go. Just hit a "Format Document" button and let jesus take the wheel.

billbrod:
Yeah, I think we decided on python black in order to avoid having a long drawn-out discussion about style.

plenoptic/tools/data.py
```diff
@@ -132,12 +142,13 @@ def load_images(paths, as_gray=True):
...
-        raise Exception("Somehow ended up with other than 4 dimensions! "
-                        "Not sure how we got here")
+        raise Exception(
+            "Somehow ended up with other than 4 dimensions! " "Not sure how we got here"
+        )
     return images
 
 
-def convert_float_to_int(im, dtype=np.uint8):
+def convert_float_to_int(im: np.ndarray, dtype=np.uint8) -> np.ndarray:
```
billbrod, line 151: dtype here should use `Literal`

plenoptic/tools/optim.py
```diff
@@ -1,30 +1,28 @@
...
 """
 import torch
-from typing import Union
+from torch import Tensor
+from typing import Optional, Tuple
 import numpy as np
 
 
-def set_seed(seed: Union[int, None] = None):
+def set_seed(seed: Optional[int] = None) -> None:
```
billbrod, line 9: why `Optional[int]` over `Union[int, None]`?

billbrod (changes requested):
Small changes requested. I'm about to also push a bunch of small changes in a similar vein, so double-check that those make sense to you

billbrod:
I removed a commented-out function (`power_spectrum` in `tools/signal.py`). We clearly weren't using it, so I got rid of it.

plenoptic/tools/optim.py
lyndond: They're equivalent https://stackoverflow.com/a/51710151 . Code bases I've worked in mainly go with Optional[whatever] because it's cleaner.

plenoptic/synthesize/eigendistortion.py
```diff
@@ -140,83 +170,102 @@ def to(self, *args, attrs=[], **kwargs):
...
+        seed: Optional[int] = None,
+    ) -> Tuple[Union[Tensor, List], Tensor, Tensor]:
         r"""Compute eigendistortions of Fisher Information Matrix with given input image.
 
         Parameters
         ----------
-        method: {'exact', 'power', 'randomized_svd'}, optional
-            Eigensolver method. Jacobian tries to do eigendecomposition directly (
+        method:
+            {'exact', 'power', 'randomized_svd'} Eigensolver method. Jacobian tries to do eigendecomposition directly (
```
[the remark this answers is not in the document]
lyndond: Done.

plenoptic/tools/data.py
lyndond: Using `Union[np.uint8, np.uint16]` since Literal is for for builtin types.

plenoptic/tools/optim.py
billbrod: sounds good, that's also what the sciware typing session recommended

billbrod (approved):

PR #177, opened 2022-09-02
bugfix: expand in PS model

theowoo:
This bug affects non-square inputs.


billbrod (approved):
Thanks for catching this and adding a test! It looks good to me.

kbonnen:
Sorry for the slow reply and thanks @theowoo!  This fix looks good to me.  As you might guess, we are using exclusively square stimuli right now so this was missed in the matlab --> python conversion.

kbonnen (approved):
Thanks again for contributing the fix @theowoo.

PR #181, opened 2022-10-18
Make the installs for the GPU and CPU tests the same

billbrod:
Small changes were necessary to make the install for the GPU-tests on jenkins have the same flags as for the CPU tests on Github actions.

This was a problem because the cached dependencies used in the GPU tests were getting out of sync (the matplotlib version required a newer numpy version, in #177 ), which the `--upgrade` flag should fix. (Note that we install the requirements using `requirements.txt` in the Dockerfile, which means they'll all get cached. The actions of the Jenkinsfile don't get cached, but that has the same `pip install .` that the CPU tests use. Effectively, we've just split it into two steps)

Additionally, this PR removes the `-e` flag from the CPU tests, since we should be testing the production install.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Install dependencies
         # nbclient 0.5.5 is the first version that includes jupyter execute
         run: |
-          pip install --upgrade --upgrade-strategy eager -e .
+          pip install --upgrade --upgrade-strategy eager .
           pip install jupyter ipywidgets
           pip install "nbclient>=0.5.5"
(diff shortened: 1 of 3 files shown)


nikparth (approved):
LGTM

dylex:
Looks good to me. I might be inclined to use explicit file names rather than the `*.npy` glob just in case people have other npy files around during install, but fine either way.

billbrod:
> Looks good to me. I might be inclined to use explicit file names rather than the `*.npy` glob just in case people have other npy files around during install, but fine either way.

Ah, that's a good point. I'll make that change. Thanks for looking this over!

PR #182, opened 2022-10-31
Remove dill

billbrod:
Our most recent test were failing, due to a problem with `dill`, which is a library that allows for pickling of more python callables. The most recent several versions of `dill` (at least `0.3.4` through `0.3.6`) give an error when loading a `Metamer` or `MADCompetition` object with the most recent version of `torch` (`1.13`). The error message was completely opaque to me (`AttributeError: Can't get attribute 'NoneType' on <module 'builtins' (built-in)>`) and I couldn't find any relevant info.

So, I'm removing `dill` as a requirement and switching back to using the built-in pickling. The reason I had added `dill` was to allow the pickling of arbitrary callables, but we've since limited what we expect (in particular, I used to allow `Metamer` to accept any callable as a model, but we've since changed that to require it to be an object, not a function). The only limitation I could find now when swapping dill back for the built-in was the saving of lambdas, but, importantly, custom functions still work as long as they're explicitly defined with a `def` block  (and [at the top level of a module](https://stackoverflow.com/a/36995008), though that's probably less relevant for users). So I've updated the tests and docstrings appropriately.

### DIFF
--- a/jenkins/requirements.txt
+++ b/jenkins/requirements.txt
@@ -9,5 +9,4 @@ requests>=2.21
 imageio>=2.5
 pytest>=5.1.2
 scikit-image>=0.15.0
-dill
 einops>=0.3.0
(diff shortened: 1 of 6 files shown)


lyndond (approved):
LGTM

nikparth (approved):
LGTM

PR #184, opened 2022-12-12
change python versions

### LINKED ISSUE #74: Segfault with python 3.9
We're running into segfaults on python 3.9, which looks like it's a [pytorch issue](https://github.com/pytorch/pytorch/issues/50014). In our code, It looks like it happens whenever `.synthesize()` is called, and, looking at the pytorch issue, that's because of `.backward()`.

Minimum example:
```
import torch; import plenoptic as po
im = torch.randn((1, 1, 256, 256), requires_grad=True)
v1 = po.simul.PooledV1(.5, im.shape[-2:])
v1(im).sum().backward()
```

It looks like everything works fine, but then when you quit, you should get a message that the process was terminal by signal `SIGSEGV` or something that similarly tells you there was a segfault.

billbrod:
Drop support for 3.6, add for 3.9 and 3.10

closes #74 , #156

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     needs: [get_notebooks]
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, 3.10]
+        python-version: [3.7, 3.8, 3.9, '3.10']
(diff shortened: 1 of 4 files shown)


lyndond (approved):
LGTM

Bichidian (approved):

PR #188, opened 2022-12-22
GitHub actions fixes

### LINKED ISSUE #186: Fix codecov issue
This is the problem: https://github.com/pytest-dev/pytest-cov/issues/388

I came up with a solution in pyrtools, but I'm not sure it's the best one.

### LINKED ISSUE #187: github actions fixes
Run [action-updater](https://vsoch.github.io/action-updater/getting_started/user-guide.html) on our CI yml file. It will fix the issue with API rate limit (which is due to `setup-ffmpeg`, as well as the warnings about `set-output` and nodejs versions

billbrod:
This updates our github actions file so that we no longer hit the rate limit issue (which was due to an issue with `setup-ffmpeg`) and removes the various warnings about nodejs versions (by updating other versions). Also fixes the issue where codecov thought our coverage was 0% for everything, which is because we no longer install an editable version of the package (`pip install -e .`) during the tests, but the production version (`pip install .`). The fix for that is not super elegant, but works.

closes #187 , closes #186


lyndond (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -14,19 +14,25 @@ jobs:
...
   get_notebooks:
     runs-on: ubuntu-latest
     outputs:
       notebook: ${{ steps.get-notebooks.outputs.nb }}
     steps:
-      - uses: actions/checkout@v2
-      - id: get-notebooks
+    - uses: actions/checkout@v3
+    - id: get-notebooks
         # it's weird to me, but the quotes around \n should *not* be escaped or it breaks
```
Bichidian, line 33: I think this comment is no longer valid.

Bichidian (changes requested):

.github/workflows/ci.yml
billbrod: good catch, I removed it

Bichidian (approved):

PR #201, opened 2023-04-03
Tutorial zero

### LINKED ISSUE #99: Write tutorial 1
Conceptual overview of package, giving high-level intro to synthesis. Eero will be involved in this.

billbrod:
Closes #99 

This PR adds a new page to the documentation, providing a conceptual introduction to stimulus synthesis with no code. It steps through cone metamers and then foveated metamers briefly. I also overhauled the "simple example" notebook to turn it into a quickstart and made the Documentation index page into an actual introduction.


billbrod:
This is largely about the docs, so view it on [read the docs](https://plenoptic--201.org.readthedocs.build/en/201/) (which now builds docs for our PRs). [here's](https://readthedocs.org/projects/plenoptic/builds/) the general list of all builds

EricThomson:
In general, this is a huge improvement from page zero! The previous iteration started me off telling me about problems with reproducibility, and now we start with a real introduction to the library. This is a really nice improvement. The conceptual introduction is a really nice touch. 

I have a few comments on some of the pages, please take them with a grain of salt this is not my area of expertise, or even close to it. 

### HOME page
I really like the general intro. In the first paragraph, I'd consider stealing from the readme for the last sentence "These synthetic images can then be used in psychophysics experiments for further investigation."

I found the stuff on 'why perform stim synthesis' to be a bit less polished. People who want that are getting it in the Conceptual Introduction, people who don't need it aren't going to need it and aren't going to get much new, and will be a bit underwhelmed as it doesn't seem as well articulated or thought out as the stuff in the conceptual introduction. At worst, I think it may just be better to cut that part, at best tighten that part up to justify its existence. I really think it might be better to just go right into synthesis methods, which is a great section!

I would consider breaking this page with headers, so after that part have a header on `Synthesis methods`, and then `Models, Metrics, and Model Components` (or something more felicitous). 

### Conceptual intro
Great stuff on metamers. For Fig 8 I'd mention the red ellipse in the legend (you mention it in the text, but people will be looking for what it means in the legend probably). 

It might be nice to mention the other kinds of synthesis problems that you approach in plenoptic: based on the conceptual introduction people might think it's all about constructing metamers, but the other three synthesis methods mentioned in the home page are cool and I wonder if it would be possible to give a little bit more detail in the second to last paragraph. E.g., maybe expand that out to discuss just how much you have in this library, maybe make a new section and call it `Beyond metamers` and cover eigendistortions etc. 

### Quickstart
It might be helpful to explain the model a bit before initializing the code (you do explain it later, but by then some people will already be frustrated "what is going on what are these things I'm doing?". I think you don't need to go overboard with explanation, but something simple like "here we will use plenoptic's api to implement an simple model -- a 2d gaussian low pass filter" (or whatever). 

The explanation of the metamers of Einstein is really good!

As I said, I don't have much this is really good stuff it makes me want to use the library!

billbrod:
Thanks for your feedback Eric!

PR #205, opened 2023-05-02
Imageio fix

billbrod:
This change fixes a recent issue with imageio that led to issues with:
1. Automatic conversion from floats in `[0,1]` to int8 images.
2. Changing what happens when `as_gray=False` and loading in a grayscale image.

Additionally, this adds support for RGBA images.


lyndond (approved):
lgtm

PR #226, opened 2023-11-16
Small documentation-related fixes

billbrod:
- Rename "modules" to "API Documentation" in docs, which is more stanard
- Adds urlchecker action, to make sure all our urls are valid


billbrod:
The `check_urls` action is no longer required for merge, because it also checks outputs of the jupyter notebooks, which we can't guarantee to be available. In particular, it includes a message from pytorch saying it's downloading from a specific url which you can't access yourself directly. So need to remember to check it manually.

PR #230, opened 2023-12-14
Updates installation instructions and CONTRIBUTING

billbrod:
Adds some info to help explain optional dependencies and the branching workflow (@ericthomson I based a lot of that on what you have in Caiman, which is great!)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,22 +2,19 @@
 
 `plenoptic` is a python library of tools to help researchers better understand
 their models. While the majority of authors are members or alumni of NYU's [Lab
-for Computational Vision](https://www.cns.nyu.edu/~lcv/), we welcome other
-contributors!
+for Computational Vision](https://www.cns.nyu.edu/~lcv/), we are open to
+contributions from anyone!
 
 First, please check out the [Code of Conduct](CODE_OF_CONDUCT.md) and read it
-before going any further. You may also want to check out the [README](README.md)
-for a longer overview of the project and how to get everything installed, the
-tutorials (Jupyter notebooks found in the `examples/` folder) for some examples
-of how to interact with the library, and the
-[documentation](https://plenoptic.readthedocs.io/).
+before going any further. You may also want to check out the [main page of the
+documentation](https://plenoptic.readthedocs.io/en/latest/) for a longer
+overview of the project and how to get everything installed, as well as pointers
+for further reading, depending on your interests.
 
 If you encounter any issues with `plenoptic`, please open an
-[issue](https://github.com/LabForComputationalVision/plenoptic/issues)!
-We have a template for bug reports, following it (so you provide the
-necessary details) will make solving it easier. Right now, we have
-enough on our plate that we're not considering any enhancements or new
-features --- we're focusing on implementing what we plan to do.
+[issue](https://github.com/LabForComputationalVision/plenoptic/issues)! We have
+a template for bug reports, and following it (so you provide the necessary
+details) will make solving your problem much easier.
 
 If you'd like to help improve `plenoptic`, we have a bunch of issues
 we're working through. For people who are not already familiar with
@@ -28,8 +25,11 @@ make progress on one of the existing issues, please give it a try.
 
 In order to submit changes, create a branch or fork of the project, make your
 changes, add documentation and tests, and submit a [Pull
-Request](https://github.com/LabForComputationalVision/plenoptic/pulls). The
-amount and form of documentation to add depends on the size of the submitted
+Request](https://github.com/LabForComputationalVision/plenoptic/pulls). See
+[contributing to the code below](#contributing-to-the-code) for more details on
+this process.
+
+The amount and form of documentation to add depends on the size of the submitted
 changes. For a significant change (a new model or synthesis method), please
 include a new tutorial notebook that walks through how to use them. For
 enhancements of existing methods, you can probably just modify the existing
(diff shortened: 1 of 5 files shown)


docs/install.rst
```diff
@@ -81,9 +91,30 @@ package:
...
 
-In either case, to open the notebooks, navigate to the ``examples/`` directory
-under this one on your terminal and activate the environment you install jupyter
-into (``plenoptic`` for 1, ``base`` for 2), then run ``jupyter`` and open up the
-notebooks. If you followed the second method, you should be prompted to select
-your kernel the first time you open a notebook: select the one named
-"plenoptic".
+   .. attention:: This method only works with conda environments. If you are using another method to manage your python virtual environments, you'll have to use one of the other methods.
+
+3. Manually install the kernel in your virtual environment. This requires only a single jupyter install and is the most general solution (it will work with conda or any other way of managing virtual environments), but requires you to be a bit more comfortable with handling environments. Again, if you followed the instructions to create a ``conda`` environment named ``plenoptic``::
```
EricThomson, line 96: I'm having trouble seeing how this is different from option 2 and what it is buying you. I'd consider explaining more.

EricThomson (commented):
Overall looks great, not many comments.

CONTRIBUTING.md
```diff
@@ -2,22 +2,19 @@
 
 `plenoptic` is a python library of tools to help researchers better understand
 their models. While the majority of authors are members or alumni of NYU's [Lab
-for Computational Vision](https://www.cns.nyu.edu/~lcv/), we welcome other
-contributors!
+for Computational Vision](https://www.cns.nyu.edu/~lcv/), we are open to
+contributions from anyone!
```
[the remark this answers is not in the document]
billbrod: I think you're probably right.

CONTRIBUTING.md
```diff
@@ -55,28 +55,67 @@ We try to keep all our communication on Github, and we use several channels:
 
 ## Contributing to the code
 
+### Contribution workflow
```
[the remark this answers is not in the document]
billbrod: thank you!

CONTRIBUTING.md
```diff
@@ -55,28 +55,67 @@ We try to keep all our communication on Github, and we use several channels:
...
+
+1. Go to the [plenoptic repo](https://github.com/LabForComputationalVision/plenoptic/) and click on the `Fork` button at the top right of the page. This creates a copy of plenoptic in your Github account.
+2. You should then clone *your fork* to your local machine and create an editable installation. To do so, follow the instructions for an editable install found in our [docs](https://plenoptic.readthedocs.io/en/latest/install.html#installation), replacing `git clone https://github.com/LabForComputationalVision/plenoptic.git` with `git clone https://github.com/<YourUserName>/plenoptic.git`.
+
+You should probably also install all the optional dependencies, so that you can run tests, build the documentation, and run the jupyter notebooks locally. To do so, run `pip install -e .[docs,dev,nb]` from within the copy of `plenoptic` on your machine (see [this section](https://plenoptic.readthedocs.io/en/latest/install.html#jupyter) of our documentation for information on how to set up jupyter if you don't want an extra copy of it in this environment).
+
```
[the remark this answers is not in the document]
billbrod: oh that's not a bad idea. I've generally worked on my own repos, so haven' tused upstream as much

CONTRIBUTING.md
```diff
@@ -55,28 +55,67 @@ We try to keep all our communication on Github, and we use several channels:
...
+#### Creating a new branch
 
+As discussed above, each feature in `plenoptic` is worked on in a separate branch. This allows us to have multiple people developing multiple features simultaneously, without interfering with each other's work. To create your own branch, run the following from within your `plenoptic` directory:
+
+```bash
+# switch to main and make sure it's up-to-date
+git checkout main
```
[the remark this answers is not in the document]
billbrod: hmm, I agree what do you think about what I added?

docs/install.rst
```diff
@@ -62,13 +72,13 @@ package:
    $ conda activate plenoptic
    $ conda install -c conda-forge jupyterlab ipywidgets torchvision
 
-   This is easy but, if you have multiple conda environments and want to use
-   Jupyter notebooks in each of them, it will take up a lot of space.
+   This is the easiest but, if you have multiple virtual environments and want
+   to use Jupyter notebooks in each of them, it will take up a lot of space.
 
 2. Use `nb_conda_kernels
```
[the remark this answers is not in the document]
billbrod: running `conda activate` is just to make sure they're in the right environment (if they just opened the terminal, it's not a problem, but if they've been doing something else, it might be)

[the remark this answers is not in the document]
billbrod: I've tried to add a bit more explanation of the options and what their advantages are, is this any clearer?

docs/install.rst
```diff
@@ -58,42 +58,49 @@ with a random string, ``matplotlib`` will tell you the available choices.
...
 
    .. attention:: This method only works with conda environments. If you are using another method to manage your python virtual environments, you'll have to use one of the other methods.
 
-3. Manually install the kernel in your virtual environment. This requires only a single jupyter install and is the most general solution (it will work with conda or any other way of managing virtual environments), but requires you to be a bit more comfortable with handling environments. Again, if you followed the instructions to create a ``conda`` environment named ``plenoptic``::
+3. Install jupyte in your ``base`` environment and manually install the kernel in your virtual environment. This requires only a single jupyter install and is the most general solution (it will work with conda or any other way of managing virtual environments), but requires you to be a bit more comfortable with handling environments. Again, if you followed the instructions to create a ``conda`` environment named ``plenoptic``::
 
    $ # activate your 'base' environment, the default one created by conda/miniconda
    $ conda activate
```
EricThomson, line 106: should this also be `conda activate base`?

docs/install.rst
```diff
@@ -106,15 +113,42 @@ package, depending on how you wish to handle virtual environments:
 
    ``/path/to/jupyter/env`` is the path to your base conda environment, and depends on the options set during your initial installation. It's probably something like ``~/conda`` or ``~/miniconda``. See the `ipython docs <https://ipython.readthedocs.io/en/stable/install/kernel_install.html>`_ for more details.
 
+   With this setup, similar to option 2, you have a single jupyter install that can run kernels from any virtual environment. The main difference is that it can run kernels from *any* virtual environment (not just conda!) but that you have to run an additional line after installing ``ipykernel``  into the environment (``python -m ipykernel install ...``).
```
EricThomson, line 116: Maybe mention another main difference: there is less stuff installed in base: no need to install `nb_conda_kernels` because you are instead pointing toward the base jupyter path (or whatever) from your virtual environment, so there is an even bigger savings than in option 2. 

This is really useful stuff, going above and beyond standard installation help.  :smile:

docs/install.rst
```diff
@@ -106,15 +113,42 @@ package, depending on how you wish to handle virtual environments:
...
+      -
+   *  -
+      - |:white_check_mark:| Automatically finds new environments with ``ipykernel`` installed
+      -
+   *  - 3. Manual kernel installation
+      - |:white_check_mark:| Flexible: works with any virtual environment setup
+      - |:x:| More complicated
+   *  -
+      - |:white_check_mark:| Requires only one jupyter installation
+      - |:x:| Extra step for each new environment
```
EricThomson, line 145: At this point this now could be a standalone Medium article. 😄 Super helpful. For people that know they are going to use conda, I wonder if there is much of a payoff for option 3 (unless `nb_conda_kernels` is large, in which case there is more than just flexibility perhaps).

docs/install.rst
```diff
@@ -106,15 +113,42 @@ package, depending on how you wish to handle virtual environments:
...
-Once you have jupyter installed and the kernel set up, navigate to the
-``examples/`` directory under this one on your terminal and activate the
-environment you installed jupyter into (``plenoptic`` for 1, ``base`` for 2 or 3),
-then run ``jupyter`` and open up the notebooks. If you followed the second or
-third method, you should be prompted to select your kernel the first time you
-open a notebook: select the one named "plenoptic".
+Running the notebooks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Once you have jupyter installed and the kernel set up, navigate to plenoptic's ``examples/`` directory on your terminal and activate the environment you installed jupyter into (``plenoptic`` for 1, ``base`` for 2 or 3), then run ``jupyter`` and open up the notebooks. If you followed the second or third method, you should be prompted to select your kernel the first time you open a notebook: select the one named "plenoptic".
```
EricThomson, line 152: Why not just go to the dir, `conda activate plenoptic`, and get started?

EricThomson (commented):
Much improved!

docs/install.rst
billbrod: `nb_conda_kernels` is pretty small.

If you're definitely using conda to manage your environments, the main reason to avoid using `nb_conda_kernels` is because you might not want jupyter to see all your environments. for example, maybe you generally don't use jupyter, so you installed it to look at notebooks in a handful of environments. and using `nb_conda_kernels` could clutter up your interface. but it would be a pretty small difference (and if the discussion makes sense to a user, I assume they're already fairly familiar with what's going on)

docs/install.rst
billbrod: you have to activate the environment that jupyter is in, so you'll run `conda activate plenoptic` only if you followed method 1. does that make sense?

PR #254, opened 2024-02-21
Updates python versions and trove classifiers

### LINKED ISSUE #247: Packaging updates
Copying from [pyopensci review](https://github.com/pyOpenSci/software-submission/issues/150):

> I would change trove classifiers
> - the package is not in beta annymore if you are using in the lab
> - you can also specify specific pythons: 3.7, etc
>
> You probably want to raise the minimum python requires and adhere to [SPEC0](https://scientific-python.org/specs/spec-0000/) since you are depending on core scientific python libraries
>
> right now core packages are at 3.10-3.12, but if you need numba then you might need to raise it to 3.9 and test on 3.9-3.11 since https://github.com/numba/numba/issues/9197

billbrod:
With this, we update our supported python versions from 3.7-3.11 to 3.10-3.12, in order to be compliant with [SPEC0](https://scientific-python.org/specs/spec-0000/).

From now on, following that, we will support three python minor versions, rolling them over at the end of the year.

Closes #247

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -32,7 +32,7 @@ If applicable, add screenshots to help explain your problem.
 
 **System (please complete the following information):**
  - OS: [e.g. Mac (with version), Ubuntu 18.04]
- - Python version [e.g. 3.7]
+ - Python version [e.g. 3.12]
  - Pytorch version [e.g., 1.4]
  - Plenoptic version [e.g. 0.1]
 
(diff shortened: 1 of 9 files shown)


billbrod:
@NickleDave: I was thinking I should mention SPEC-0 specifically, to make it clear why I'm supporting these three versions, and when I'll change which versions I support. Does that belong in the readme? the installation page of the docs? `CONTRIBUTING`? all three?

I couldn't find a discussion of this on the documentation for numpy or scipy, so I'm wondering if it's too "in the weeds"

PR #257, opened 2024-02-22
Updates action versions, change how notebooks are tested

billbrod:
This does two things:
- Updates action versions to get rid of the warning during github actions
- Changes how notebooks are tested: we were previously parallelizing across all of them, which I think was actually costing us time (and making it output to navigate), so now we run them all in one action.


billbrod:
Turns out that this now looks like it will 22 to 28 minutes for the notebooks, whereas before, if things ran successfully, they would take ~20 minutes. I think the small increase in time is worth it, given that our previous setup would occasionally just hang during one of the notebooks, and how difficult it was to navigate the output.

PR #259, opened 2024-05-31
Memory improvement: torch tensor replacements when applicable

### LINKED ISSUE #31: replace instances with torch.tensor when applicable
Instantiating a tensor via torch.tensor should be avoided when torch.from_numpy or torch.as_tensor can be used instead. This is bc torch.tensor always copies data, whereas the other to do not, or at the very least avoid when possible.

hmd101:
Replaced ``torch.tensor`` with ``torch.as_tensor`` or ``torch.from_numpy``where applicable in order to save memory.
I didn't replace ``torch.tensor``when argument ``requires_grad``was present. If that should be desired, I can amend the PR.

Example:
```Python
data = [1, 2, 3]

# will not share memory with original data. Resulting tensor is independent of the input data
tensor_0 = torch.tensor(data) 

# shares memory with original data. Changes of the original data will affect tensor and vice versa
tensor_1 = torch.as_tensor(data) 

 # resulting tensor and numpy array will share memory and changes will be reflected in the other
array = np.array([1, 2, 3])
tensor_2 = torch.from_numpy(array)
```

Resolves #31


hmd101:
Opening new PR that doesn't have the formatting changes #260

PR #260, opened 2024-05-31
Torch tensor replacements

hmd101:
Replaced ``torch.tensor`` with ``torch.as_tensor`` in order to save memory.
Note that [torch.as_tensor](https://pytorch.org/docs/stable/generated/torch.as_tensor.html) only shares the parameters ``data``, ``dtype``, and ``device``with [torch.tensor](https://pytorch.org/docs/stable/generated/torch.tensor.html#torch.tensor), which has additional parameters, such as ``required_grad``. Adjustments were made accordingly.

Example:
```Python
data = [1, 2, 3] # can be list, tuple, numpy array, scalar

# will not share memory with original data. Resulting tensor is independent of the input data
tensor_0 = torch.tensor(data) 

# shares memory with original data. Changes of the original data will affect tensor and vice versa
tensor_1 = torch.as_tensor(data) 
```

Resolves issue #31 .


billbrod (approved):
This looks great, thanks for putting it together Hanna!

PR #276, opened 2024-07-26
Dynamically Adjust Number of Columns in plot_representation Based on Number of Loss Plots

hmd101:
**Changes:**

- Replaced the hardcoded number of columns in the `plot_representation` function with a variable that dynamically adjusts based on the number of loss statistics in the `data` tensor.
- This change allows the function to handle varying numbers of loss sub-plots more flexibly.

**Impact:**

- Improves the adaptability of the `plot_representation `function.
- Remedies shape and axis errors thrown when calling `plot_synthesis_status`.

### DIFF
--- a/src/plenoptic/simulate/models/portilla_simoncelli.py
+++ b/src/plenoptic/simulate/models/portilla_simoncelli.py
@@ -929,16 +929,29 @@ class PortillaSimoncelli(nn.Module):
           between each orientation at each scale (summarized using Euclidean
           norm)
 
-        If self.n_scales > 1, we also have:
+        If self.n_scales > 1, we also have combination of the following, where all
+        cross-correlations are summarized using Euclidean norm over the 
+        channel dimension:
         
         - cross_scale_correlation_magnitude: the cross-correlations between the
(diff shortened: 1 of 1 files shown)


src/plenoptic/simulate/models/portilla_simoncelli.py
```diff
@@ -929,16 +929,29 @@ def plot_representation(
...
+          of the image.
+
+        - cross_channel_correlation_magnitude: the cross-correlations between 
+          the channels of the magnitude of the pyramid coefficients.
+
+        - cross_channel_correlation_real: the cross-correlations between the
+            channels of the real component of the pyramid coefficients, referred
+            to as 'phase' in Brown et al., 2023.
+
+
```
billbrod, line 954:
```suggestion        
```

These are only present for the color version, right?

billbrod (changes requested):
Thanks for this Hanna! If you can remove the bit of the docstring that only applies to the color model, this is good to go.

billbrod (approved):

PR #278, opened 2024-08-05
Adds conda install instructions

### LINKED ISSUE #104: Put on conda
Talk to Flatiron SCC about this, since they've done it for some packages.

billbrod:
We're now up on conda-forge, so this updates the installation instructions to reflect that.

closes #104


BalzaniEdoardo (commented):
The explanations are very clear and the content is all there. Structurally, I would probably change it a little bit
- intro section: saying plenoptic can be install from pip & conda, explain briefly pros and cons (pip can be used for installing from source vs why using conda instead)
- conda install section: all your conda content
- pip install section: all pip content including editable mode and optional deps.

Mixing the two is ok but a bit harder to follow

billbrod:
That makes a lot of sense, I was just being lazy. I pushed a reorganization of the docs that hopefully makes things clearer for the user with a heavy use of tabs. This makes the raw rst file a bit harder to read, but I think it's worth it! Let me know what you think

docs/install.rst
```diff
@@ -5,58 +5,112 @@ Installation
...
-$ # install in editable mode with `-e` or, equivalently, `--editable`
-$ pip install -e .
+$ conda install plenoptic -c conda-forge
+
+.. _pip:
+
+Installing with pip
+-------------------
+
+Wile ``conda`` handles both virtual environment management **and** package installation, ``pip`` only installs packages; you'll need to use some other system to manage your virtual environment. You can use ``conda`` to manage your virtual environment and ``pip`` to install packages, but note that when using pip with a conda environment, all pip commands should come after all the conda commands (e.g., you shouldn't run ``conda install matplotlib`` after ``pip install plenoptic``), because conda is unaware of any changes that pip makes. See `here <https://www.anaconda.com/blog/using-pip-in-a-conda-environment>`_ for more details.
```
BalzaniEdoardo, line 32:
```suggestion
While ``conda`` handles both virtual environment management **and** package installation, ``pip`` only installs packages; you'll need to use some other system to manage your virtual environment. You can use ``conda`` to manage your virtual environment and ``pip`` to install packages, but note that when using pip with a conda environment, all pip commands should come after all the conda commands (e.g., you shouldn't run ``conda install matplotlib`` after ``pip install plenoptic``), because conda is unaware of any changes that pip makes. See `here <https://www.anaconda.com/blog/using-pip-in-a-conda-environment>`_ for more details.
```

docs/install.rst
```diff
@@ -5,58 +5,112 @@ Installation
...
+      $ python -m venv path\to\environments\plenoptic-venv
+      # activate the environment
+      $ path\to\environment\plenoptic-venv\bin\activate
+      $ pip install plenoptic
+
+Note that when using ``venv``, you have to decide where you'd like to place the folder containing all files related to the virtual environment. If the virtual environment is related to the development of a package, as in :ref:`source`, it is typical to place them within the repository for that package. Otherwise, it is typical to place them all in a single directory somewhere so they're easy to keep track of.
+
+.. _source:
+
+Installing from source
```
BalzaniEdoardo, line 60: I would add a (for delevopers) in the tile. something like this
```suggestion
Installing from source (for developers)
```

docs/install.rst
```diff
@@ -5,58 +5,112 @@ Installation
...
 
+Note that, with the above setup, all files related to your virtual environment are stored in a hidden directory named ``.venv`` within the ``plenoptic/`` directory you cloned. Therefore, if you delete the ``plenoptic/`` directory, you'll need to rerun the setup above to create a new virtual environment.
+
 .. attention:: To install ``plenoptic`` in editable mode, you need ``pip >= 21.3`` (see pip's `changelog <https://pip.pypa.io/en/stable/news/#id286>`_). If you run into `an error <https://github.com/LabForComputationalVision/plenoptic/issues/227>`_ after running the ``pip install -e .`` command, try updating your pip version with ``pip install --upgrade pip``.
 
 .. _optional-deps:
+
 Optional dependencies
----------------------
+^^^^^^^^^^^^^^^^^^^^^
```
BalzaniEdoardo, line 100: This should probably be a sub-section of the installing from source;

BalzaniEdoardo (approved):
Looks good. The notebook setup section is complex but the tabs makes it easier to focus on only what you need. I may add an extra label for each solution 

1. **[Easy]** Install ....
2. **[Intermediate]** ...
3. **[Intermediate]** ...

docs/install.rst
billbrod: I think it's not, because you could also use it to setup the environment for running jupyter notebooks (see the jupyter section farther down).

PR #280, opened 2024-08-13
adds Hanna to citation.cff

billbrod:
Adds Hanna Dettki to the citation file, following her internship.

@hmd101 can you confirm that this is how you'd like to be credited in a citation? Do you have an [orcID](https://orcid.org/)? If so, we'll add it. If not, I'd encourage you to get one.


hmd101 (approved):
Just added my orcid ID, thanks!

PR #285, opened 2024-08-22
Updates github org name

billbrod:
Updates links from `LabForComputationalVision/plenoptic` to `plenoptic-org/plenoptic`


billbrod:
As part of this, also corrected lots of `http` links to `https`

PR #292, opened 2024-09-12
Fix broken to method, test non-standard models/metrics, don't assume dtype

### LINKED ISSUE #283: `metamer.to` not working?
Running the following produces unexpected behavior:

```python
img = po.data.einstein()
ps = po.simul.PortillaSimoncelli(img.shape[-2:])
met = po.synth.MetamerCTF(img, ps, loss_function=po.tools.optim.l2_norm)
met.to('cuda')
```

This raises the following warning: 

```
/mnt/home/wbroderick/plenoptic/src/plenoptic/synthesize/synthesis.py:211: UserWarning: model has no `to` method, so we leave it as is...
  warnings.warn("model has no `to` method, so we leave it as is...")
```

which should only happen if caling `met.model.to` raises an AttributeError`, which it doesn't.

Confusingly, `met.model(met.metamer)` works after this.

However, running `met.synthesize()` after the aboveraises `ValueError: can't optimize a non-leaf Tensor`, so something gets messed up with the optimizer.

I think the proper way to do this is to make use of `register_buffer` or similar, and then remove our definition of `def to()`, like we've done for the `PortillaSimoncelli` and `SteerablePyramid` modules.

billbrod:
As documented in #283, calling `.synthesize()` after `.to()` raises a `ValueError` because `met.metamer` was no longer a leaf tensor. This is because, for some reason, `Tensor.to` is considered a differentiable operation. I'm not sure why that's desirable behavior, but this PR adds a solution: when moving tensors, we call `tensor.data.to` (instead of `tensor.to`) and then check whether the original version required a gradient and, if so, add it back. This is similar to how we behave in `load`. 

You can find some discussions of this issue: [on the pytorch forum](https://discuss.pytorch.org/t/moving-tensors-between-devices/83261/3), and on stackoverflow ([one](https://stackoverflow.com/questions/63662624/how-do-i-retain-grads-and-also-change-device-type-in-pytorch), [two](https://stackoverflow.com/questions/65301875/how-to-understand-creating-leaf-tensors-in-pytorch)).

Additionally,:
- the warning being raised was because I was trying to set `self.model`, instead of `self._model`, which was raising an `AttributeError` because `model` doesn't have a setter method. `self.model.to` was still being called (and modified in place), which was why `met.model(met.metamer)` worked without a problem. Replacing `self.model` with `self._model` solves this problem.
- moves this `self._model.to` out of `Synthesis` and into the child classes (eigendistortion, metamer), since not all synthesis objects will have model attributes.
- This is wrapped in try/except because we do not require that models have a to method (see [model requirements](https://plenoptic.readthedocs.io/en/latest/models.html)). it is possible that a model will not have a to method but still behaves correctly, basing all their tensor dtype and devices on their input. so we simply raise a warning.
- adds tests for this behavior (just call `synthesize()` after `to()` to our existing tests)

closes #283

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -46,9 +46,10 @@ metric meets the following requirements and see :class:`plenoptic.metric` for
 some examples):
 
 * a metric must be callable, accept two 4d ``torch.Tensor`` objects as inputs,
-  and return a scalar as output. This can be a ``torch.nn.Module`` object, like
-  models, but the examples metrics are all functions.
+  and return a scalar as output. It can be a ``torch.nn.Module`` object or other
+  callable object, like models, as well as a function.
 * when called on two identical inputs, the metric must return a value of 0.
+* it must always return a non-negative number.
 
 Finally, :class:`plenoptic.synthesize.metamer.Metamer` supports coarse-to-fine
 synthesis, as described in [PS]_. To make use of coarse-to-fine synthesis, your
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -12,8 +12,12 @@ way to check whether your model meets the following requirements, and see
 * should inherit ``torch.nn.Module`` (this is not strictly necessary, but will
   make meeting the other requirements easier).
 * must be callable, be able to accept a 4d ``torch.Tensor`` as input, and return
-  a 3d or 4d ``torch.Tensor`` as output. If you inherit ``torch.nn.Module``,
-  implementing the ``forward()`` method will make your model callable.
+  a 3d or 4d ``torch.Tensor`` as output.
+
+  * If you inherit ``torch.nn.Module``, implementing the ``forward()`` method
+    will make your model callable.
+  * Otherwise, implement the ``__call__()`` method.
+
 * the above transformation must be differentiable by ``torch``. In practice,
   this generally means you perform all computations using ``torch`` functions
   (unless you want to write a custom ``.backward()`` method).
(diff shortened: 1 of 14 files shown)


billbrod:
Note that in `Synthesis.to()`, for tensors I'm essentially doing:
```python
attr = getattr(self, k)
attr = attr.data.to(*args, **kwargs)
if getattr(self, k).requires_grad:
    attr = attr.requires_grad_()
setattr(self, attr, k)
```

which is weird because I'm basically resetting the tensor to its data. however, I can't seem to get around that. Replacing the .to line with either of the following  fails:

```diff
- attr = attr.data.to(*args, **kwargs)
+ attr = attr.to(*args, **kwargs)
```

```diff
- attr = attr.data.to(*args, **kwargs)
+ attr.data = attr.data.to(*args, **kwargs)
```

The first fails because `attr.grad` is None (because `attr` is no longer a leaf tensor) and the second fails because of a type issue with the (Adam) optimizer when `synthesize()` is called before and after the `to()` call (and they both fail in the same way if requires_grad is explicitly checked or not).

So I guess we're stuck with this solution.

billbrod:
While testing these, realized there were several other related changes to make:
- I document that models don't have to be `torch.nn.Module`, but didn't test that. Now I do (and made a change to `validate_model` so it passes).
- I document that metrics can be objects, but didn't test that. Now I do.
- Realized that metrics must always return non-negative numbers, so added that to `validate_metric`.
- In several places, there were functions that accepted a tensor as input and initialized another tensor always with `dtype=torch.float32`. These functions would thus fail if the input had a different dtype. They now initialize the new tensor with the same dtype as the input.

PR #294, opened 2024-09-20
updates install instructions

billbrod:
Following @NickleDave's point in #238 , tells people to run `pip install -e .[dev]` as the default for local install, since they'll probably want the tests.

Also wraps the install target in quotes when including optional dependencies, which is necessary for `zsh`.


docs/install.rst
BalzaniEdoardo, line 1: I was going through the install instruction once again, and I have some minor suggestions:

1. I would start a paragraph at line 8 after "separate instructions for the two methods." Add a line break and start a new paragraph with the next sentence.
2. The discussion about pip install in a conda environment seems a bit a side note "You can use conda to manage your virtual environment ...for more details.", maybe it is better to suggest the use of venv, and have an admonition of type info/suggestion with this information.  
3. In the installation from source instructions you create the venv in the package root directory, while in the pip installation you have a placeholder path. You could add a comment in the install code snippet saying something like "# create the environment (usually in the package root folder)".
4. This is also content of an admonition (extra info): "With an editable copy, any changes locally will be automatically reflected in your installation (under the hood, this command uses symlinks)."
5. I am not sure that the last two section belongs in the install instruction ("Running the notebooks" and "ffmpeg...". Maybe worth considering adding an extra doc page for it. Also not really urgent, and maybe not in this PR?
6. Related to point 5, i would rename the "Running notebooks locally" something like "Setting Up Jupyter for Example Notebooks", which is a title that fits better with the install page scope.

For the rest it is really good

BalzaniEdoardo (approved):
I left mostly very minor comments I had after going through the install notes once again.

In general this are very good instruction, my points were almost all directed to streamlining the instructions by either "moving extra information in admonitions" or "moving content in a dedicated page".

The extra content you are adding, like explaining the editable mode, is valuable, but could make the core of the instruction a bit verbose; so I would keep the content, but as an optional side note.

billbrod:
I agree with your points, went ahead and made those changes. Let me know what you think!

For now, I moved the jupyter content to a separate file; I'm going to do a bit of restructure of the docs soon, so exactly how things are arranged will change, but I agree it's separate.

BalzaniEdoardo (approved):
all good!

PR #305, opened 2024-11-22
Updates to steerable pyramid

billbrod:
This PR has two small updates to the steerable pyramid:
- We had previously raised an error when `order=0`, which was due to a [reconstruction error](https://github.com/plenoptic-org/plenoptic/wiki/Meeting-notes#evening-1). ~Looking into it, this error only happens for the complex pyramid and we now just raise a warning (as for odd-sized images, which have the same problem).~
- ~We do, however, raise an exception if the user tries to create a complex pyramid with order=0 that is tightframe, since that is impossible.~
- We raise an exception for any pyramid with `order=0, is_complex=True`. The rationale is explained in #326 
- Furthermore, we add a small section to the steerable pyramid notebook showing how to use torch / torchvision to pad or pad and crop, in order to change the behavior at the boundary.
- Expands the docstrings to explain the order of the scales/orientations that the pyramid and PS statistics are returned in (fine to coarse, all orientations at each scale before moving on)
- Removes unused `max_orientations` arg from `_recon_keys` method, which was a holdover from pyrtools.

There's a question about where the warning about reconstruction belongs: during pyramid object construction or during the call to the method that actually does the reconstruction (`recon_pyr`)? -- Decided to keep them during construction


hmd101 (approved):

hmd101:
Looks good to me!

BalzaniEdoardo (approved):
Looks good to me, and I agree that the error should be raised at initialization.

billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/305

PR #310, opened 2024-12-09
adds git fetch --tags

billbrod:
By default, our Jenkins set up doesn't grab git tags, but we need them in order to properly version plenoptic when it's installed before building the docs (since sphinx includes the version in the output). This makes sure to fetch them before installing plenoptic in the docs build.


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/310

billbrod:
Additionally, we make sure to copy the outputs of the documentation build to a different directory -- originally, the clone of the plenoptic-documentation repo was happening to the docs directory, and so it was clobbering the output.

PR #314, opened 2025-02-13
Update how dynamic version is set

billbrod:
While putting together PRs for nemos and pynapple to use `setuptools_scm` to set the package version dynamically, realized that setuptools_scm doesn't recommend writing version to file anymore but using `importlib.metadata` ([docs](https://setuptools-scm.readthedocs.io/en/latest/usage/), in line with [PEP 566](https://peps.python.org/pep-0566/).

This similarly updates how sphinx grabs the version, in line with their suggestions as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,9 +35,6 @@ docs/tutorials/assets
 
 *.mypy_cache
 
-# created automatically by setuptools.scm, don't track
-src/plenoptic/version.py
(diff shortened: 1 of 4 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/314

PR #315, opened 2025-02-20
Updates steerable pyramid error message on shape mismatch

billbrod:
The error message when the steerable pyramid was passed an input with the wrong shape was confusing, hopefully this makes it clearer.

Also adds validation and casting for steerpyr's image shape: image shape must be an iterable of length two whose values are castable to ints.


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/315

PR #319, opened 2025-03-04
Adds doctests action, metric doctests

BalzaniEdoardo:
Added examples to the `perceptual_distance.py` module.

@billbrod EDIT: also:
- adds doctests to ci.yml and noxfile
- fixes failing synthesis doctests, which removes most of the print statements from geodesic and  eigendistortion, either removing them entirely or turning them into warnings
- also removes note about creating a new instance of synthesis object and then loading. hopefully that's already clear enough
- use [matplotlib's sphinx plot directive](https://matplotlib.org/stable/api/sphinxext_plot_directive_api.html) in some of the doctests. updates  configuration of that directive to show source/not show link by default (because we'll use the directive more in the doctests than the docs)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/319

PR #322, opened 2025-03-04
add docstrings to example images in data

sjvenditto:
Adds docstrings and examples for the five example images in plenoptic.data. Also changed the default parameter `as_gray` to `True` for the parrot image, which is a grayscale image.


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/322

billbrod (approved):
Thanks Sarah Jo!

PR #328, opened 2025-03-21
Fix deprecation warnings

### LINKED ISSUE #273: torch meshgrid behavior changing
With torch version 2.1, calling torch.meshgrid raises the following warning: "torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument". See [meshgrid docs](https://pytorch.org/docs/stable/generated/torch.meshgrid.html) for more info.

we only call torch.meshgrid in a handful of places, so should be easy enough to fix.

billbrod:
Fix two deprecation warnings:
- imageio API change. I believe this change doesn't affect me (after looking through [their migration guide](https://imageio.readthedocs.io/en/stable/reference/userapi.html#migrating-to-the-v3-api)), so this just explicitly imports `imageio.v3` to suppress the warning.
- torch.meshgrid change, see [docs](https://pytorch.org/docs/stable/generated/torch.meshgrid.html), [issue](https://github.com/pytorch/pytorch/issues/50276). torch's meshgrid default behavior was backwards from that of numpy. they are planning on changing default behavior to match that of numpy. now, I explicitly set indexing arg for all calls to either numpy or torch meshgrid. I had ran into this at one point when creating the `polar_radius` / `polar_angle` functions and had a comment to the effect of "why is this different". Removed that comment and reverted the call to look the standard way.

closes #273

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -17,7 +17,7 @@ import plenoptic as po
 import torch
 import imageio
 
-img = torch.tensor(imageio.imread('data/einstein.pgm'), dtype=torch.float32).unsqueeze(0).unsqueeze(0)
+img = po.data.einstein()
 model = SomeModel()
 met = po.synth.Metamer(img, model)
 # this raises an error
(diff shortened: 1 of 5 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/328

PR #330, opened 2025-03-27
Small fixes to geodesic

billbrod:
As I was porting over recent changes to https://github.com/plenoptic-org/geodesics, realized there were several small issues:
- forgot to remove the `n_steps` arg from the `setup()` signature (should use `self.n_steps`, which is set at initialization)
- forgot to remove `optimizer` arg from `synthesize()` call signature
- the step_energy calculation shouldn't collapse across channels, that can be useful to keep separate.


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/330

PR #333, opened 2025-04-01
Removes deprecated 1.1.0 objects: geodesic, SimpleMetamer, make_synthetic_stimuli

billbrod:
These have been marked as deprecated since 1.1.0, and so we're now removing them.

The list is:
- geodesics and its plotting functions (now in https://github.com/plenoptic-org/geodesics), because it's not robust enough
- the contents of `tools/straightness.py` (`make_straight_line`, `sample_brownian_bridge`, `deviation_from_line`, `translation_sequence`), which were only used for geodesics (also in https://github.com/plenoptic-org/geodesics)
- `SimpleMetamer`: not used in anything
- `make_synthetic_stimuli`: only used for testing, users are better off grabbing them from `pyrtools` when needed
- `minimum` and `maximum`: `einops.reduce` is better for this

And all tests related to them. `make_synthetic_stimuli` was used in some other tests, so replaced that with the 256x256 test images (einstein, curie, color_wheel, metal, nuts)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/333

PR #335, opened 2025-04-02
Fixes all warnings raised by pytest

billbrod:
This PR addresses all the warnings that were being raised by pytest. Previously, we would get a huge number of warnings in the pytest output, which made it annoying to actually find the relevant output when things were failing. With this, I have either fixed warnings or selectively ignored them with [pytest's warning capture capabilities](https://docs.pytest.org/en/stable/how-to/capture-warnings.html). Several of these are configured for all tests, and others happen at a per-test level. In general, it is preferable to write the test such that warnings are not raised (e.g., calling `model.eval()` before passing a model to a synthesis method).

Now, any warnings not explicitly ignored will result in the test failing.

Questions:
- ~Should I just move the warnings from `pyproject.toml` to the specific tests where they're relevant? As I look at this now, they aren't *that* general.~ I moved these to annotate the specific functions. For now, the only warnings in pyproject.toml are those that are temporary.
- ~Not sure the best way to handle doctests. Currently, I have doctest-specific filters (in the action) which ignore the DeprecationWarning for gaussian1d (which I raise), and, in gaussian2d, a UserWarning from pytorch about efficiency with same padding and even-sized kernels. I could rewrite the doctest for the second to avoid this, but I want the code to be as straightforward as possible there, so I think this is fine?~ 
    - ~I moved my DeprecationWarning into the pyproject.toml (because it's temporary), but kept the filter for torch's efficiency warning on the command line.~
    - actually, decided to decided to just ignore all warnings for doctests (pass `-W ignore` in the CI). anything relevant should come up in the regular tests.

Specific changes:
- across all tests, ignore the following plenoptic-thrown warnings: 
  - `Loss has converged`: MADCompetition, Metamer depending on how long synthesize is run for
  - `Image range falls outside`: SSIM, NLPD, MS-SSIM during synthesis
  -  `SSIM was designed for grayscale images`: SSIM, MS-SSIM on multi-channel images
  - `Looks like representation is image-like`: animate with models with 4d outputs
  - `Randomized SVD complete`: Eigendistortion when synthesize is run with `method='random'`
  - `You will need to call setup`: loading a synthesis object with scheduler or optimizer set.
  - `Validating whether model can work with coarse-to-fine`: metamer synthesis with PortillaSimoncelli
- ~coverage warning during pytest was because I had `--cov=plenoptic` instead of `--cov=src`~
   - that "fix" actually broke the interaction between pytest-cov and pytest-xdist, such that I was getting a reported coverage of zero. I think it's related to [this note in the docs](https://pytest-cov.readthedocs.io/en/latest/plugins.html): pytest can't control the order that the plugins are loaded in. locally, xdist was loaded before cov, whereas the CI had the reverse (at least, that's the order they were reported on the `plugins:` line of the pytest output). setting `--cov=src` worked locally (coverage reported as expected), but broke the CI (coverage reported as 0). None of the suggested fixes either there on the github issues fixed this. reverting to `--cov=plenoptic`  means that both local and CI report coverage correctly. I do get the `CoverageWarning`, but it doesn't seem to affect anything and, because it's raised after the tests, not in them, they don't affect my intended behavior of "if tests raise a warning, fail"
- Invalid escape sequence because of regex confusion (now explicitly checked for with [ruff](https://docs.astral.sh/ruff/rules/invalid-escape-sequence/))
- SSIM, MS-SSIM, and NLPD image range warnings have been made more general: don't specify the image range, just that it's outside [0, 1]. That way, only the first one will be shown, the rest will be suppressed.
- The included torch Modules without trainable parameters are set to `eval` mode at the end of initialization: `LaplacianPyramid`, `SteerablePyramidFreq`, `PortillaSimoncelli`
- Replace [numpy.in1d](https://numpy.org/doc/stable/reference/generated/numpy.in1d.html) with [numpy.isin](https://numpy.org/doc/stable/reference/generated/numpy.isin.html#numpy.isin) (numpy 2.0 DeprecationWarning)
- `SteerablePyramidFreq.convert_tensor_to_pyr` should explicitly call `.real` on the residuals (instead of converting to `torch.float`, which raises a warning). When the `is_complex=True` and `convert_pyr_to_tensor(split_complex=False)` then the output tensor is complex-valued, but the residuals are actually real-valued, so their imaginary components are always zero. 
  -  adds a test that verifies this
- In `Eigendistortion._indexer`, replace `int(np.where(...))`  (which raises a DeprecationWarning) with `torch.where(...).item()`
- `upsample_convolve` no longer represents the padding info as numpy arrays, but torch tensors (avoids a DeprecationWarning)
- `imshow`, `pyrshow`, `animshow` now all raise more explicit ValueError if `channel_idx` or `batch_idx` are not None or int.
- `plot_representation` figsize arg defaults to None. If both ax and figsize are None, then we figsize to a reasonable value. If both are not-None, we raise a ValueError. Adds tests for this. (Removes warning we were throwing unnecessarily)
- Changes some generic Exceptions to more specific types (ValueError, TypeError)
- All models used in tests now call `model.eval()`
- Removes `Identity` model from conftest.py (duplicate of `naive.Identity`)
- More careful about closing of matplotlib figures during tests, to avoid warning about having too many open figures
- all `MADCompetition` instances in test explicitly set `metric_tradeoff_lambda` (to avoid warning we were throwing)
- Removes some unnecessary `mpl.use("agg")` in tests files where there was no animating happening
- Generally, set `cache_filt=True` when `pretrained=True` for `frontend` models in tests to avoid warning we throw. Exception is when we explicitly want to check the behavior, and then we ignore the warning from that specific test.
- For polar angle image tests, use `torch.argmin` and `torch.unravel_index` instead of numpy's versions, to avoid numpy-thrown DeprecationWarnings.

Unrelated to the above:
- needed to update `setuptools` in the Jenkins run, because of [this issue](https://github.com/scipy/scipy/issues/16726)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/335

BalzaniEdoardo:
Hi @billbrod,
> - Should I just move the warnings from `pyproject.toml` to the specific tests where they're relevant? As I look at this now, they aren't *that* general.

For your first question, I'd prefer more targeted configs then a global ignore in the `pyproject.toml`.  NeMoS also have global waning filters in the pyproject but I want to be more specific.
One way to keep the code a bit cleaner is registering new markers
```toml
# ..rest of pyproject.toml
# ...add to [tool.pytest.ini_options]
markers = [
  "ignore_loss_at_convergence: used to ignore the 'Loss has converged' warning"
]
```

```python
# ... rest of content of conftest.py
def pytest_collection_modifyitems(config, items):
    for item in items:
        if 'ignore_loss_at_convergence' in item.keywords:
            print(f"Applying warning filter to {item.nodeid}")
            item.add_marker(
                pytest.mark.filterwarnings("ignore:Loss has converged:UserWarning")
            )
```

Then in the test file

```python
@ignore_loss_at_convergence
def test_my_test_func():
...
```

> - Not sure the best way to handle doctests. Currently, I have doctest-specific filters (in the action) which ignore the DeprecationWarning for gaussian1d (which I raise), and, in gaussian2d, a UserWarning from pytorch about efficiency with same padding and even-sized kernels. I could rewrite the doctest for the second to avoid this, but I want the code to be as straightforward as possible there, so I think this is fine?

For this question, what is the issue with setting up a filter in the pyproject.toml if that's your deprecation? 

```
[tool.pytest.ini_options]
# ...
filterwarnings = [
  # other filters
  "ignore:gaussian1d.* is deprecated:DeprecationWarning",
]

```

billbrod:
@BalzaniEdoardo , I agree with making the ignore filter more specific, I was just being lazy. Changing that now. I think registering the additional marker is overkill, so I just added them like normal.

Moving the DeprecationWarning filter to pyproject.toml also makes sense.

I think, moving forward, I'll note that pyproject.toml should only include these temporary filters (that way they're all in one place an easy to find), whereas everything else should be test-specific.

PR #340, opened 2025-04-10
Improvements to center_crop, simple models

billbrod:
This PR makes several small quality-of-life changes:
- Makes center_crop better behaved. We now validate the input args and raise informative ValueErrors if e.g., you want to crop the center 200 pixels of a 100x100 image
- finishes adding type-hinting and doctests to models/naive.py and models/frontend.py
- better validation of inputs to `circular_gaussian2d`, raising more informative error messages when things go wrong

### DIFF
--- a/src/plenoptic/simulate/models/frontend.py
+++ b/src/plenoptic/simulate/models/frontend.py
@@ -1,13 +1,16 @@
 """
-Model architectures in this file are found in [1]_, [2]_. `frontend.OnOff()` has
-optional pretrained filters that were reverse-engineered from a previously-trained model
-and should be used at your own discretion.
+Model architectures in this file are found in [1]_, [2]_, pretrained parameters from
+[3]_.
 
 References
 ----------
 .. [1] A Berardino, J Ballé, V Laparra, EP Simoncelli, Eigen-distortions of hierarchical
     representations, NeurIPS 2017; https://arxiv.org/abs/1710.02266
 .. [2] https://www.cns.nyu.edu/~lcv/eigendistortions/ModelsIQA.html
+.. [3] A Berardino, Hierarchically normalized models of visual distortion
+   sensitivity: Physiology, perception, and application; Ph.D. Thesis,
+   2018; https://www.cns.nyu.edu/pub/lcv/berardino-phd.pdf
+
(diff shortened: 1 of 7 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/340

src/plenoptic/simulate/models/naive.py
```diff
@@ -284,39 +345,42 @@ def __init__(
...
+            on_center = on_center.repeat(out_channels)
+        if len(on_center) != out_channels:
+            raise ValueError("len(on_center) must equal out_channels")
         self.on_center = on_center
 
-        if isinstance(kernel_size, int):
-            kernel_size = (kernel_size, kernel_size)
-        self.kernel_size = kernel_size
+        if amplitude_ratio < 1.0:
+            raise ValueError("amplitude_ratio must at least be 1.")
```
sjvenditto, line 381: This error is not being tested, and more generally I'm not seeing tests for setting `amplitude_ratio`?

sjvenditto (changes requested):
The docs changes all look good and are clear! I only had one minor comment about `on_center` flag when using the pretrained parameters, and I didn't see any tests for the `amplitude_ratio` parameter for the `CenterSurround` naive model

src/plenoptic/simulate/models/frontend.py
```diff
@@ -96,6 +105,12 @@ def __init__(
                     "pretrained is True but cache_filt is False. Set cache_filt to "
                     "True for efficiency unless you are fine-tuning."
                 )
+            if not on_center:
```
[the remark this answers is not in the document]
billbrod: not that I can think of, though it might be interesting just to see what the "inverse" model looks like. I don't like silently changing inputs on users, so I was deciding between doing this and raising an exception, and decided a warning was a reasonable tradeoff

src/plenoptic/simulate/models/naive.py
billbrod: I was just being lazy....

sjvenditto (approved):
Looks good!

PR #341, opened 2025-04-11
Remove dimensionality restrictions

billbrod:
Previously, I had hard restrictions on dimensionality: plenoptic synthesis methods only worked on 4d inputs and required models to have 3d or 4d outputs. After talking with several folks, I realized this doesn't make sense -- it is common to represent videos as 5d tensors and audio as 3d tensors, and there's no reason why you couldn't collapse across the channel dimension.

Furthermore, `Metamer` required that the synthesized `metamer` tensor had the same shape as the target image. This made the "mixing texture" example in the Portilla-Simoncelli notebook and the (forthcoming) style transfer example more complex for no real gain.

This PR removes both of those restrictions; now, `validate_model` and `validate_input` raise warnings when those are not the case, saying that plenoptic has mainly been tested with 4d inputs and 3d/4d outputs, and telling folks to raise an issue if they have trouble. The docs have been similarly updated.

As of now, we only test that we can initialize and start synthesis for 1 to 5d model outputs and 2 to 5d inputs (except Eigendistortion, which tests 1, 3, 4, 5d -- because we can only work with a single batch, we want that dummy dimension). We *do not* test that we get anything reasonable. Later, we will add examples with video and audio models.

The models and metrics still all operate on 4d image tensors (and say as much in their docstrings), and the behavior of the display functions are likewise unchanged.

Additionally:
- Updated some synthesis object docstrings that were out of date, still referring to "limitations in pickle" for loss functions / metrics
- The two eigendistortion notebooks have been updated to use the torchvision feature extractor to grab intermediate deepnet layer activations, which simplify them
- The `02_Eigendistortion` notebook first part has been reworked so that the simple example is 1d
- Eigendistortion now has a private `_image_shape` attribute, which it uses to reshape its outputs, rather than four public attributes (`n_batch`, etc)
- `tools/validate.py` functions now have examples


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/341

examples/06_Metamer.ipynb
```diff
@@ -92,7 +92,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Like all of our models, when this is called on the image, it returns a 3d or 4d tensor (in this case, 4d). This representation is what the `Metamer` class will try to match. "
+    "When this model called on the image, it returns a 4d tensor. This representation is what the `Metamer` class will try to match. "
```
sjvenditto, line 95:
```suggestion
    "When this model is called on the image, it returns a 4d tensor. This representation is what the `Metamer` class will try to match. "
```

src/plenoptic/synthesize/metamer.py
```diff
@@ -189,14 +186,14 @@ def setup(
...
-                        ", n_channels, im_height, im_width]) but got "
-                        f"{initial_image.size()}"
-                    )
+                validate_input(initial_image, allowed_range=self.allowed_range)
                 if initial_image.size() != self.image.size():
-                    raise ValueError("initial_image and image must be same size!")
+                    warnings.warn(
+                        "initial_image and image are different sizes! This "
+                        "has not been tested as much, open an issue if you "
+                        "problems! https://github.com/plenoptic-org/plenoptic/"
```
sjvenditto, line 194:
```suggestion
                        "have any problems! https://github.com/plenoptic-org/plenoptic/"
```

sjvenditto (changes requested):
Looks good! Only found a couple typos

sjvenditto (approved):

PR #345, opened 2025-04-23
Fix malformed sphinx directives

billbrod:
I add a couple of `..plot::`, which don't render correctly. They should look like `.. plot::` (note the space).

This fixes that and then adds a small script with accompanying pre-commit hook which checks for anything that matches the regex `\.\.[a-z]+::` and raises an error if so. Hopefully this will catch that in the future. The script is intended to run on only python, markdown, and rST; that constraint is handled by pre-commit.


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/345

PR #347, opened 2025-05-02
Hotfix: fix eigendistortion.to()

billbrod:
Issue with calling `eigendistortion.to` with the frontend models (and similar).

updates `test_to` to check for this, and updates metamer tests to make sure we similarly check for it.

Additionally, allow user to set atol/rtol used when checking tensor equality during load. When trying to save/load on two machines, you can end up with very small differences that are not easily resolvable. (I ran into this on my work laptop and linux workstation, which both had the same versions of plenoptic and torch, and the problem didn't depend on GPU vs CPU; my guess is that it was something related to CUDA or BLAS). For example:

```python
# on my work laptop
img = po.data.einstein()
model = po.simul.CenterSurround(31, cache_filt=True, pad_mode='circular')
eig = po.synth.Eigendistortion(img, model)
eig.synthesize(5)
eig.save('test.pt')

# on my linux workstation
img = po.data.einstein()
model = po.simul.CenterSurround(31, cache_filt=True, pad_mode='circular')
eig = po.synth.Eigendistortion(img, model)
# FAIL
eig.load('test.pt')
# PASS
eig.load('test.pt', tensor_equality_atol=1e-7)
```

Obviously, this can be abused to avoid the load checks at all, but I think that's on the user.

### DIFF
--- a/src/plenoptic/synthesize/eigendistortion.py
+++ b/src/plenoptic/synthesize/eigendistortion.py
@@ -574,7 +574,7 @@ class Eigendistortion(Synthesis):
             pass to ``torch.device``
         tensor_equality_atol :
             Absolute tolerance to use when checking for tensor equality during load,
-            passed to :func:`torch.allclose`. It may be necessary to decrease if you are
+            passed to :func:`torch.allclose`. It may be necessary to increase if you are
             saving and loading on two machines with torch built by different cuda
             versions. Be careful when changing this! See
             :class:`torch.finfo<torch.torch.finfo>` for more details about floating
@@ -583,7 +583,7 @@ class Eigendistortion(Synthesis):
             dealing with a numerical issue.
         tensor_equality_rtol :
             Relative tolerance to use when checking for tensor equality during load,
-            passed to :func:`torch.allclose`. It may be necessary to decrease if you are
+            passed to :func:`torch.allclose`. It may be necessary to increase if you are
             saving and loading on two machines with torch built by different cuda
             versions. Be careful when changing this! See
             :class:`torch.finfo<torch.torch.finfo>` for more details about floating
--- a/src/plenoptic/synthesize/eigendistortion.py
+++ b/src/plenoptic/synthesize/eigendistortion.py
@@ -532,16 +532,16 @@ class Eigendistortion(Synthesis):
             "_representation_flat",
         ]
         super().to(*args, attrs=attrs, **kwargs)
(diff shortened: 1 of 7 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/347

src/plenoptic/synthesize/eigendistortion.py
```diff
@@ -570,6 +572,16 @@ def load(
...
+        tensor_equality_atol :
+            Absolute tolerance to use when checking for tensor equality during load,
+            passed to :func:`torch.allclose`. It may be necessary to decrease if you are
+            saving and loading on two machines with torch built by different cuda
+            versions. Be careful when changing this!
+        tensor_equality_rtol :
+            Relative tolerance to use when checking for tensor equality during load,
+            passed to :func:`torch.allclose`. It may be necessary to decrease if you are
+            saving and loading on two machines with torch built by different cuda
+            versions. Be careful when changing this!
```
BalzaniEdoardo, line 584: Should you advise against going above the float32 numerical precision of more than one order of magnitude? I think that would be rarely necessary

BalzaniEdoardo (approved):
This looks necessary and I agree with the implementation

src/plenoptic/synthesize/synthesis.py
```diff
@@ -159,6 +161,24 @@ def load(
...
+            passed to :func:`torch.allclose`. It may be necessary to decrease if you are
+            saving and loading on two machines with torch built by different cuda
+            versions. Be careful when changing this! See
+            :class:`torch.finfo<torch.torch.finfo>` for more details about floating
+            point precision of different data types (especially, ``eps``); if you have
+            to increase this by more than 1 or 2 decades, then you are probably not
+            dealing with a numerical issue.
+        tensor_equality_rtol :
+            Relative tolerance to use when checking for tensor equality during load,
+            passed to :func:`torch.allclose`. It may be necessary to decrease if you are
```
sjvenditto, line 175:
```suggestion
            passed to :func:`torch.allclose`. It may be necessary to increase if you are
```

billbrod:
> Looks good, my only comment is I think you put "decrease" in the docstrings when I think you meant "increase"?

Whoops, good catch!

src/plenoptic/synthesize/eigendistortion.py
billbrod: I added a more informative description

PR #349, opened 2025-05-12
Starts fixing docstrings, sphinx warnings

billbrod:
This PR fixes the docstrings of the contents of most of the package: `data/`, `metric/`, `simulate/`, and `synthesize/` (everything but `tools/`).

"Fix" here means:
- Adds ruff rules related to docstrings: "D", "DOC201", "DOC202", "DOC402", "DOC403", "DOC501", "ANN"
- Adds numpydoc linting: all but SA01 (require `See Also` section, don't want to require for everything), PR04 (since we're using type annotations, don't need to describe type in docstring), and EX01 (require `Examples` section -- will remove this eventually)
- Adds small scripts to check docstrings for markdown or what looked like malformed sphinx directives (missing space)
- Adds the above checks to pre-commit config and noxfile
- Fixed the latex rendering and use intersphinx to link directly to relevant functions (in plenoptic or pytorch)

Additionally, some changes in sphinx conf to reduce warnings:
  - Build should ignore the pyrtools-thrown warning about dpi
  - Don't show inherited methods/attributes from `torch.nn.Module` (but do show the inherited ones from plenoptic objects)
  - Clean up extensions a bit


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/349

billbrod:
The failing checks are because:
- http://www.cvrl.org/cones.htm is down. This has happened before, intermittently, so should consider switching to the web archive version of the site
- pre-commit.ci is running on all the files in `tools/` (which I haven't finished updating). I thought it would only run on changed files, so not sure what's going on. Regardless, I ran pre-commit on all the changed files locally and they pass, so I think I will just bypass that check (next PR will fix the failing checks on those files, so this is temporary.)

src/plenoptic/simulate/models/naive.py
```diff
@@ -276,10 +297,13 @@ def forward(self, x: Tensor, **conv2d_kwargs) -> Tensor:
...
-    on-center/off-surround, or vice versa.
+    r"""
+    Center-Surround, Difference of Gaussians (DoG) filter model.
+
+    Can be either on-center/off-surround, or vice versa.
 
     Filter is constructed as:
+
     .. math::
         f &= amplitude_ratio * center - surround \\
```
sjvenditto, line 308:
```suggestion
        f &= amplitude\_ratio * center - surround \\
```

src/plenoptic/synthesize/eigendistortion.py
```diff
@@ -69,38 +95,30 @@ def fisher_info_matrix_eigenvalue(
...
-    explain perceptual sensitivity in humans. It estimates eigenvectors of the FIM.
-    A model, :math:`y = f(x)`, is a deterministic (and differentiable)
-    mapping from the input pixels :math:`x \in \mathbb{R}^n` to a mean output
-    response vector :math:`y\in \mathbb{R}^m`, where we assume additive white
+    explain perceptual sensitivity in humans. It estimates eigenvectors of the Fisher
+    Information Matrix. A model, :math:`y = f(x)`, is a deterministic (and
+    differentiable) mapping from the input pixels :math:`x \in \mathbb{R}^n` to a mean
+    output response vector :math:`y\in \mathbb{R}^m`, where we assume additive white
     Gaussian noise in the response space.
     The Jacobian matrix at x is:
```
sjvenditto, line 123:
```suggestion
    The Jacobian matrix at :math:`x` is:
```
(if you want to be consistent with the "x" in line 131)
there's also an "f" in 126 and an "x" in 127 not in math mode.

src/plenoptic/synthesize/eigendistortion.py
```diff
@@ -590,13 +647,27 @@ def load(
...
-            any additional kwargs will be added to ``pickle_module.load`` via
-            ``torch.load``, see that function's docstring for details.
+        **pickle_load_args
+            Any additional kwargs will be added to ``pickle_module.load`` via
+            :func:`torch.load`, see that function's docstring for details.
+
+        Raises
+        ------
+        ValueError
+            If :func:`setup` or :func:`synthesize` has been called before this call
```
sjvenditto, line 657: `setup` is a dead link. Is it because it doesn't exist as a method for eigendistortion?

src/plenoptic/synthesize/eigendistortion.py
```diff
@@ -187,21 +208,33 @@ def synthesize(
             computation. Ignored for other methods.
         p
             Oversampling parameter for randomized SVD. k+p vectors will be sampled,
-            and k will be returned. See docstring of ``_synthesize_randomized_svd``
+            and k will be returned. See docstring of :meth:`_synthesize_randomized_svd`
```
sjvenditto, line 211: this method is a dead link, I believe because docstrings are not created for private methods?

src/plenoptic/synthesize/mad_competition.py
```diff
@@ -1207,76 +1301,88 @@ def animate(
     ],
     width_ratios: dict[str, float] = {},
 ) -> mpl.animation.FuncAnimation:
```
sjvenditto, line 1303: In the docstring for this (and related) functions, the `FuncAnimation` output type does not link to matplotlib documentation. however, other external output types do, such as `Tensor`. Not sure why this is the case

sjvenditto (commented):
In general, everything looks good. Everything in the built docs appears to be formatted correctly. I only found a couple small issues with latex formatting (valid formatting so it wouldn't be caught, but made an unintended subscript)

The only other thing is that some of the methods / types come out to be unlinked. It's usually because it's trying to link a method that doesn't have an associated docstring (e.g. links to private methods). This won't throw an error because there isn't a link at all, so it's not "invalid". I've pointed out a few instances where I caught this in the comments. I don't know how to systematically test or check for this though.

docs/conf.py
```diff
@@ -89,10 +92,48 @@
...
+def find_module(obj):
+    members = inspect.getmembers(obj, is_interesting)
+    to_return = []
+    for name, mem in set(members):
+        if inspect.ismodule(mem):
+            # go through the modules recursively
+            to_return.extend(find_module(mem))
+        else:
+            # if they inherit torch Module
+            if torch.nn.Module in mem.__bases__:
```
BalzaniEdoardo, line 126:
```suggestion
            if issubclass(mem, torch.nn.Module):
```

This would work for nested inheritance too

docs/conf.py
```diff
@@ -204,6 +240,7 @@
 myst_enable_extensions = [
     "colon_fence",
     "dollarmath",
+    "fieldlist",
```
BalzaniEdoardo, line 243: what is this for?

tests/check_docstrings.py
BalzaniEdoardo, line 1: Looks like something that could be implemented as a `ast.NodeVisitor` instead of traversing the tree and do a try/except.

BalzaniEdoardo: you need to define a visit for the nodes that have docstrings, [from the ast docs](https://docs.python.org/3/library/ast.html)

> ast.get_docstring(node, clean=True)
>    Return the docstring of the given node (which must be a [FunctionDef](https://docs.python.org/3/library/ast.html#ast.FunctionDef), [AsyncFunctionDef](https://docs.python.org/3/library/ast.html#ast.AsyncFunctionDef), [ClassDef](https://docs.python.org/3/library/ast.html#ast.ClassDef), or [Module](https://docs.python.org/3/library/ast.html#ast.Module) node), or None if it has no docstring. If clean is true, clean up the docstring’s indentation with [inspect.cleandoc()](https://docs.python.org/3/library/inspect.html#inspect.cleandoc).


```python
class DocstringVisitor(ast.NodeVisitor):
    def __init__(self, path):
        self.path = path
        self.links = []
        self.backticks = []

    def check_docstring(self, node, name):
        # ... implement logic and store output
        pass
        
    def visit_FunctionDef(self, node):
        self.check_docstring(node, node.name)
        self.generic_visit(node)

    def visit_AsyncFunctionDef(self, node):
        self.check_docstring(node, node.name)
        self.generic_visit(node)

    def visit_ClassDef(self, node):
        self.check_docstring(node, node.name)
        self.generic_visit(node)
  
    def visit_Module(self, node):
        self.check_docstring(node, node.name)
        self.generic_visit(node)
```

BalzaniEdoardo (approved):

docs/conf.py
billbrod: https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#field-lists

billbrod: it's actually not necessary anymore -- I was originally trying to use myst for the docstrings, in which case it's necessary, but ended up just sticking with default sphinx / rest.

so I'll remove it

tests/check_docstrings.py
billbrod: oh that is nice. I don't think I'll make this change right now, since I need to finish up my work for VSS. Can you create an issue describing this? then I can return to it if I come back and change this logic.

src/plenoptic/synthesize/eigendistortion.py
billbrod: Yes, I think in general I should probably not be referring to hidden methods in public docstrings. I'll return to that when I do a bigger doc restructure

src/plenoptic/synthesize/mad_competition.py
billbrod: yep, I had set up intersphinx for torch but not matplotlib. I'll do that now

billbrod:
Also, following Edoardo's sleuthing, I added `ELLIPSIS` as a doctest arg into my pyproject.toml and removed it from the examples sections. The exception is `tools/validate.py`, which I am not modifying in this PR and will update in a separate one.

PR #351, opened 2025-05-15
corrects even/odd in steerpyr docstring

billbrod:
@Bichidian noticed that they were swapped.

The real-valued is odd symmetric (like sine), the imaginary-valued is even symmetric (like cosine).


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/351

PR #355, opened 2025-05-28
Fix PortillaSimoncelli update_plot on GPU

billbrod:
Make sure that `PortillaSimoncelli.update_plot` (and thus `animate`) works properly on GPU, by using our `to_numpy` function.

Updates test to properly check this, and then updates some other update_plot tests to check for GPU compatibility.


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/355

PR #356, opened 2025-06-02
More docstrings!

billbrod:
This PR continues the work started in #349: now all docstrings should pass validation.

This also:
- adds proper cross-references to pyrtools
- corrects the steer pyr docstring (#351 wasn't right, the parity of the filter depends on the order of the pyramid)
- adds some more validation to check for missed cross-references or None/False/True that should be wrapped in double backticks
- `animshow` now supports `zoom=None`, just like `imshow`.
- adds tests for `animshow` / `imshow` `zoom` arg.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -101,6 +101,10 @@ napoleon_use_rtype = False
 
 numfig = True
 
+# SPHINX CROSS REFERENCES
+
+add_function_parentheses = False
+
 # numpydoc
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -64,6 +64,7 @@ intersphinx_mapping = {
     "torch": ("https://pytorch.org/docs/stable/", None),
     "matplotlib": ("https://matplotlib.org/stable/", None),
     "numpy": ("https://numpy.org/doc/stable/", None),
+    "pyrtools": ("https://pyrtools.readthedocs.io/en/latest", None),
 }
 
 # Add any paths that contain templates here, relative to this directory.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ docs = [
 ]
 
 dev = [
-    "pytest>=5.1.2,<8.4",
+    "pytest>=5.1.2",
     'pytest-cov',
     'pytest-xdist',
     "pooch>=1.2.0",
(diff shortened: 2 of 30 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/356

billbrod:
For now, I'm pinning `pytest<8.4`, because I am running into the issue described https://github.com/pytest-dev/pytest-cov/issues/693.

sjvenditto (approved):
Look's good! I looked through the built docs and only noticed a couple small things in the tools module.

BalzaniEdoardo:
in the docstring of ``ssim_map`` reference number 3 doesn't appear in the docstring text.

BalzaniEdoardo:
for the math in this docstrings, I suggest to wrap the variables (amplitude, center), in `\text` otherwise the spacing between letters is off.
https://docs.plenoptic.org/docs/branch/main/api/plenoptic.simulate.models.html#raae9b18084a5-1

```
    .. math::
        f &= amplitude\_ratio * center - surround \\
        f &= f/f.sum()
```

BalzaniEdoardo:
`plenoptic.imshow()` is not linked, should it be?

https://docs.plenoptic.org/docs/branch/main/api/plenoptic.synthesize.html#plenoptic.synthesize.mad_competition.MADCompetition.initial_image

BalzaniEdoardo:
> `plenoptic.imshow()` is not linked, should it be? in the`zoom` parameter description
> 
> https://docs.plenoptic.org/docs/branch/main/api/plenoptic.synthesize.html#plenoptic.synthesize.mad_competition.MADCompetition.initial_image

elsewhere too

BalzaniEdoardo (approved):
I left comment for all I was able to find.

billbrod:
> In the references after this paragraph
> 
> https://docs.plenoptic.org/docs/branch/main/api/plenoptic.simulate.models.html#plenoptic.simulate.models.frontend.LinearNonlinear.center_surround
> 
> You have `(1,2)` appearing before the refs. In the docstrings the reference is correct, probably there is some interplay with the module level references. I don't know how to fix this

This is because the references are cited twice in the `LinearNonlinear` init docstring, so this is the proper behavior.

Made the other changes.

PR #359, opened 2025-06-25
doc string examples

wulfdewolf:
Added doc string examples to:

- `metric/perceptual_distance.py`
- `simulate/canonical_computations/non_linearities.py`
- `simulate/models/portilla_simoncelli.py`

All other files in `metric/` and `simulate/` were good on docstrings already.
(ignoring functions that start with `_`)

### DIFF
--- a/src/plenoptic/simulate/models/portilla_simoncelli.py
+++ b/src/plenoptic/simulate/models/portilla_simoncelli.py
@@ -659,24 +659,50 @@ class PortillaSimoncelli(nn.Module):
         --------
         >>> import plenoptic as po
         >>> img = po.data.curie()
-        >>> portilla_simoncelli_model = po.simul.PortillaSimoncelli(img.shape[2:])
+        >>> portilla_simoncelli_model = po.simul.PortillaSimoncelli(
+        ...     img.shape[2:], n_scales=3
+        ... )
         >>> representation_tensor = portilla_simoncelli_model(img)
         >>> representation_dict = portilla_simoncelli_model.convert_to_dict(
         ...     representation_tensor
         ... )
-        >>> for k, v in representation_dict.items():
-        ...     print(k, v.shape)
-        pixel_statistics torch.Size([1, 1, 6])
-        auto_correlation_magnitude torch.Size([1, 1, 9, 9, 4, 4])
-        skew_reconstructed torch.Size([1, 1, 5])
-        kurtosis_reconstructed torch.Size([1, 1, 5])
-        auto_correlation_reconstructed torch.Size([1, 1, 9, 9, 5])
-        std_reconstructed torch.Size([1, 1, 5])
-        cross_orientation_correlation_magnitude torch.Size([1, 1, 4, 4, 4])
-        magnitude_std torch.Size([1, 1, 4, 4])
-        cross_scale_correlation_magnitude torch.Size([1, 1, 4, 4, 3])
-        cross_scale_correlation_real torch.Size([1, 1, 4, 8, 3])
-        var_highpass_residual torch.Size([1, 1, 1])
+        >>> # We will go through and examine each of these keys individually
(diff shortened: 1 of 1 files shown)


src/plenoptic/simulate/canonical_computations/non_linearities.py
```diff
@@ -90,6 +102,21 @@ def polar_to_rectangular_dict(
...
+
+        >>> import plenoptic as po
+        >>> img = po.data.einstein()
+        >>> spyr = po.simul.SteerablePyramidFreq(img.shape[-2:])
+        >>> coeffs = spyr(img)
+        >>> energy, state = po.simul.non_linearities.rectangular_to_polar_dict(
+        ...     coeffs, residuals=True
+        ... )
+        >>> coeffs = po.simul.non_linearities.polar_to_rectangular_dict(energy, state)
+        >>> po.pyrshow(coeffs)
```
billbrod, line 118: could maybe show that these are identical to the original coeffs? but that would look fairly busy, what do you think?

src/plenoptic/simulate/canonical_computations/non_linearities.py
```diff
@@ -246,6 +290,18 @@ def local_gain_control_dict(
...
+    Examples
+    --------
+    .. plot::
+
+        >>> import plenoptic as po
+        >>> img = po.data.einstein()
+        >>> spyr = po.simul.SteerablePyramidFreq(img.shape[-2:], is_complex=True)
+        >>> coeffs = spyr(img)
+        >>> energy, state = po.simul.non_linearities.local_gain_control_dict(coeffs)
+        >>> po.pyrshow(energy, col_wrap=4)
```
billbrod, line 303:
```suggestion
        >>> po.pyrshow(energy)
```

src/plenoptic/simulate/canonical_computations/non_linearities.py
```diff
@@ -293,6 +349,18 @@ def local_gain_release_dict(energy: dict, state: dict, residuals: bool = True) -
...
     polar_to_rectangular_dict
         The analogous function for complex-valued signals.
+
+    Examples
+    --------
+    .. plot::
+
+        >>> import plenoptic as po
+        >>> img = po.data.einstein()
+        >>> spyr = po.simul.SteerablePyramidFreq(img.shape[-2:], is_complex=True)
```
billbrod, line 359:
```suggestion
        >>> spyr = po.simul.SteerablePyramidFreq(img.shape[-2:])
```

billbrod (changes requested):
Thanks Wolf! Generally, just visualizing the output of these functions is helpful (maybe with some titles). I put specific suggestions in here, and make sure to pull from main as well.

From your examples, it looks like it was unclear to you which functions should accept real-valued and which complex-valued tensors. I'm inclined then to make them raise exceptions if they receive the wrong dtype? It's not *impossible* to perform the computations on the other type, but it's not how they're intended to use and it doesn't really make sense.

- For the examples in `non_linearities.py` where you're using the SteerablePyramid, set `height=3` so we don't have too many plots
- Reviewing this made me realize that there's no reason for `polar_to_rectangular_dict` to have a `residuals` arg, should just grab them if they're present. i pushed a commit making that change

billbrod (changes requested):
Thanks for putting these together Wolf! I've gone through and left suggestions where necessary.

Several of these are changing the call from `portilla_simoncelli_model.forward(img)` to `portilla_simoncelli_model(img)` -- for torch modules, `forward` and `__call__` are the same, but using `__call__` is more idiomatic.

billbrod (approved):
Great, thanks for all your work on this Wolf! I'm going to push a couple of quick small changes and then merge:
- undoing something left over from my commits that aren't needed anymore
- testing a change I made in `signal.steer`
- adding some comments explaining the shapes in the `PortillaSimoncelli.convert_to_dict` doctest.

PR #364, opened 2025-07-23
docstrings for stats functions

wulfdewolf:


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/364

src/plenoptic/tools/stats.py
```diff
@@ -32,6 +32,42 @@ def variance(
...
+    >>> v
+    tensor(1.6667)
+
+    If you want to compute along a specific dimension, you can specify it:
+
+    >>> v = variance(x, dim=0)
+    >>> v
+    tensor([1., 1., 1.])
+
+    This function differs from ``torch.var`` in that it does not apply a correction:
```
billbrod, line 65:
```suggestion
    This function differs from the defaults of :func:`torch.var` in that it does not apply a correction:
```
get that intersphinx link

billbrod (changes requested):
Thanks for this Wolf! I think it might be helpful to show these computed on actual probability distributions, which we can plot and will help make it clearer how to interpret the values. I like having each of them use the same distributions. If we wanted to be really fancy, could have 2x2x2 to show two different values for variance, skew, and kurtosis, but that might be overwhelming. Maybe have each of them show two distributions, a standard Gaussian and then another distribution where the moment in question has changed but the other two are the same? What do you think?

wulfdewolf:
Some comments after speaking in person:
- we want `sharex=True` for all of the plots, it'll be more clear
- add a label indicating the left one is a standard Gaussian
- if reasonable, add a label indicating what the other distribution is (more skewed, more kurtosis, etc.)
- add a plot with less kurtosis

billbrod (approved):
This looks great, thanks Wolf! I'm just going to pull in main and add cross-refs to skew/kurtosis and then this is good.

PR #366, opened 2025-07-25
Regression tests

billbrod:
This PR moves the long-running syntheses out of the `Portilla_Simoncelli` and `Demo_Eigendistortion` tutorials and into `tests/test_uploaded_files.py`. That is, they are now regression tests, and the tutorials download and examine their output. As part of this, added additional texture synthesis examples, so users can download others (but see #365 -- the quality of some of the metamers should be improved).

This also adds a new `check_regression_tests.py` linting script, which checks that the calls to `setup` and `synthesize` that happen in the test file are shown in an admonition in the notebook (whether everything else is equal is validated during `load`). This behavior is explained in `CONTRIBUTING`

By default, `pytest` will skip the regression tests; to run the environment variable `RUN_REGRESSION_SYNTH=1` must be set (and a GPU available). Jenkins will run these tests as part of every PR and weekly, and they now take ~40 minutes. The docs build now takes ~15 minutes.

Importantly, see #368: I am unable to get these longer-running regression tests to produce identical output on different devices. I think this is due to CUDA version and driver version, though I am not sure. even when doing everything explained in the torch docs to guarantee reproducibility, I still end up with slightly different outputs from matrix multiplication, which then leads to slight differences in the gradient that accumulate. Should eventually see if I can track this down, but for now, this means that we need to use the outputs of Jenkins to test against.

Other changes:
- we still test the notebooks on the cpu (using the github runner), but this simplifies it: no longer user `papermill`, call `jupytext --execute` directly
- updates several of the other notebooks to use gpu to speed up the docs build. the goal is to have each notebook be under 2 minutes.
- tries to explain all this in `CONTRIBUTING`
- adds [`sphinx_togglebutton`](https://sphinx-togglebutton.readthedocs.io/) so that we can actually toggle our admonitions; I had previously added `:class: dropdown` to a couple admonitions but they weren't doing anything
- `MAD_Competition_1`: in the "full images" section, minimizing the L1 norm requires many more iterations than the other variants of that MAD Competition. We increased the learning rate a bit (didn't seem to do much) and shortened the duration; it's current duration is long enough to get the gist, but added a note explaining it really should be run for longer.
- some of the new tests use a second GPU, if available. relatedly, our Jenkins tests now use 2 GPUs (and docs switches back to 1).
- using [`pytest-order`](https://pytest-order.readthedocs.io/en/stable/usage.html), the tests in `test_uploaded_files.py` get started first, since they take the longest.
- we also use [`pytest-xdist`'s](https://pytest-xdist.readthedocs.io/en/stable/distribution.html) `--dist worksteal` option, which allows runners which finish their assigned tests to grab tests from other workers, which should help with the long-running tests.
- replace `torch.as_tensor` with `torch.stack` in `_ssim_parts`, which was leading to a warning in torch 2.8.0
- adds PR template
- Jenkins will now abort previous build when pushed to a PR with a build already running.
- small change to `synthesis.load` that should make it more amenable to user-defined models (whose exact module path may vary), we only check the final name. Thus, `po.simul.PortillaSimoncelli` will match `PortillaSimoncelli` defined in a script (we elsewhere require that their input/output behavior is identical). This is used for the variants of the PortillaSimoncelli model where we remove statistics, since that requires a new class, which is defined either in the test file or the notebook directly. also adds tests for this


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/366

linting/check_regression_tests.py
```diff
@@ -0,0 +1,88 @@
...
+with open(sys.argv[1]) as f:
+    tests = ast.parse(f.read())
+
+tests = [
+    c
+    for c in tests.body
+    if isinstance(c, ast.ClassDef) and c.name == "TestTutorialNotebooks"
+]
+assert len(tests) == 1
+tests = {b.name: b for b in tests[0].body}
```
BalzaniEdoardo, line 17: I would use a node visitor class form ast to avoid a lot of nested list comprehensions, but up to you

BalzaniEdoardo (changes requested):
This looks mostly good to me.
I just have a concern on the way the long tests are scheduled. On top of that I don't understand how xdist local groups work. 

I see that you are marking some tests with a `gpu-0/1` group tag , and the tests use a specific device. Are the processes split into two groups because there are two tags? And are the untagged tests randomly assigned to different processes?

BalzaniEdoardo (approved):
LGTM

PR #375, opened 2025-09-15
Bugfix: fixes `MetamerCTF.to` when called mid-synthesize

billbrod:

With `MetamerCTF`, if `synthesize` has been called at least once and we have not finished moving through all scales, the attribute `_ctf_target_representation` will be a Tensor that gets passed to `objective_function` during the optimization step. In other situations (before `synthesize` has been called and after we're optimizing all scales at once), its value is None.

This PR makes sure to call `_ctf_target_representation.to` if it is not None.

**Describe changes**

- Adds tests for switching device in the middle of synthesis for all synthesis methods.

**Checklist**


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/375

PR #381, opened 2025-10-20
Updates Portilla-Simoncelli synthesis in docs/tests

### LINKED ISSUE #365: PortillaSimoncelli metamer quality
As part of `regression_tests` branch (#366 ), I moved a lot of PortillaSimoncelli texture model synthesis from the notebook into `tests/test_uploaded_files.py`, where they function as regression tests (and some of their outputs are examined in the notebook / docs). However, some of their quality is not great. 

I have a notebook for examining these, which I can send to anyone if they're interested in working on this. Eventually, will make it easier to see the outputs of all these syntheses (with corresponding arguments) in the docs.

## `test_ps_basic_synthesis`
- figures 12a, 12b, 12e
- figures 14d, 15e, 15f, 16a, 16b, 16c, 16f

The first set are all "artificial textures" (per the paper), different from most of the images in the paper, and look like they haven't converged. ("synthesis success" for most of these metamers generally means a loss of about 0.1, with absolute values of all values in the representation error plot being at 0.5 or less)

The second set seem to have separate dark and light splotches (not seen in the original paper), even though they all meet the definition of "success" given above. Current hypothesis: the marginal pixel stats (especially mean and variance) aren't being given enough weight in plenoptic's optimization (compared to the matlab code). To fix this, will try multiplying them by a large number to see if that changes perceptual quality. @ershook also noticed this at one point, and found that switching the loss from `l2_norm` to `mse` seemed to help, though to my eye, the resulting metamers have other issues. @gabrielyancy99 has noticed issues removing high frequencies in texture metamers as well (when initializing with high-amplitude noise, instead of the lower range we typically use), which might be related.

Similarly, when initializing with uniform noise with a big range (rather than the narrow range of mean+/-0.05 we use in the notebook), there's a decent amount of visible high frequency noise remaining in the images. May also be able to fix by multiplying the residual highpass by something large.

## `test_ps_remove`: 
- figures 3a: all stats included having trouble converging, reduce learning rate?
- figure 3a and 3b: without marginal stats, should probably also remove the range penalty? the version in the paper looks "washed out" / low contrast.
- figure 4b: both look like they haven't quite converged.
- figure 5a: stats removed looks like it hits a minimum and then jumps back up. still finds a decent solution
- figure 8b: has the same "blotchy luminance" issue discussed above

## `test_ps_mixture`

all of them have representation error plot values of ~0.1 or so, could probably be lower.

billbrod:

Following the investigation in #365 and the speed-up in #377, this PR updates the synthesis used for the Portilla-Simoncelli in the regression tests (and thus the images / details in the docs). We also split up the PS notebook into many smaller notebooks and add a new notebook explaining some of the details of the optimization.

Closes #365. With this PR, I believe our examples are good enough: the quality is better and the synthesis faster than the matlab code.

- Most of this is user-facing docs. I wanted to make it clear that: this is the suggested way of running Portilla-Simoncelli texture synthesis, and I think it should work in a wide variety of cases. I wanted to avoid overwhelming users with info, but provide avenues for learning more if they wanted. Have I succeeded in that?
- In particular, are the optimization details and the new section of the limitations page (under "usage limitations") clear? Do they belong in their current locations? 
- The new `portilla_simoncelli_loss_factory` function, which creates the custom loss function to use for PS metamer synthesis, was the best way I could come up to simplify the use of that custom loss. Is it clear? both in what it's doing and in how to use it.
- ~What should I do with my breaking save/load change? I think I should raise a warning, rather than an Error, for now (with a future warning).~
    - ~when loading a new save with an old plenoptic version, it will raise a fairly opaque error (`KeyError: '2*_target_representation'`). I don't think there's any way I can fix that, right? other than making something in my docs that describes what's going on?~
    - ~when loading an old save with a new version, what should I do? raise a warning that says, "this save is from an older version of plenoptic and I can't check your loss function. Continuing anyway, but this will raise an error in the future." and point to the docs?~
    - For now, we raise a FutureWarning about this issue and have added a note in the docs.

**Describe changes**

- Fix for #380: `workflows/check_version.yml` needs to check out the repo so that it can find the Dockerfile
- Updates how we check notebook synthesis against regression test synthesis, updating `CONTRIBUTING.md` and `linting/check_regression_tests.py`. Now allows for a bit more flexibility in the comparison.
- Adds a "loss factory" for PS metamer synthesis, which returns the custom loss I recommend folks use. I discuss a little bit about how you could do something similar in the docs, though I think, unless you're trying to preserve backwards compatibility (like we are here), there are probably better ways to accomplish this.
- Adds some notes to the Tips+Tricks docs page with lessons learned from this process.
- Splits `docs/tutorials/models/Portilla_Simoncelli.md` into notebooks found in `docs/tutorials/models/portilla_simoncelli`: index, "what's a texture?", basic usage, examples from different types of textures, extensions (texture mixtures and extrapolation), limitations, understanding the stats, differences with matlab, and (new) optimization details.
   - With the exception of index, optimization details, and limitations, the contents are largely the same as before, just with some tweaking.
   - The limitations page has some new discussion of PS model input size restrictions.
- Bugfix: when saving/loading metamer, we use to call the loss function on two images in order to get a comparison value. we now call it on two different representations (i.e., model outputs, not inputs), which is what the loss is actually called on during synthesis. **This is a breaking change**, but is necessary to save the custom loss I now use for PS texture synthesis.
- Adds a new loss factory, `groupwise_relative_l2_norm_factory.` this isn't used in this PR (I wrote it because I thought it might do better than the PS-specific loss factory, but it didn't, see #365), but I do think it's generally useful, so I'm including it and I mention it in the docs.

**Checklist**

### DIFF
--- a/docs/tutorials/models/portilla_simoncelli/ps_intro.md
+++ b/docs/tutorials/models/portilla_simoncelli/ps_intro.md
@@ -53,9 +53,7 @@ IMG_PATH = fetch_data("portilla_simoncelli_images.tar.gz")
 plt.rcParams["figure.dpi"] = 72
 ```
 
-The simplest definition is a repeating visual pattern. Textures encompass a wide variety of images, including natural patterns such as bark or fur, artificial ones such as brick, and computer-generated ones such as the Julesz patterns ([Julesz 1978](https://link.springer.com/article/10.1007/BF00336998), [Yellot 1993](https://opg.optica.org/josaa/abstract.cfm?uri=josaa-10-5-777)). Below we load some examples.
-
-The Portilla-Simoncelli model was developed to measure the statistical properties of visual textures. Metamer synthesis was used (and can be used) in conjunction with the Portilla-Simoncelli texture model to demonstrate the necessity of different properties of the visual texture. Throughout the notebooks in this section, we will use some of these example textures to demonstrate aspects of the Portilla-Simoncelli model.
+The simplest definition of a texture is a repeating visual pattern. Textures encompass a wide variety of images, including natural patterns such as bark or fur, artificial ones such as brick, and computer-generated ones such as the Julesz patterns ([Julesz 1978](https://link.springer.com/article/10.1007/BF00336998), [Yellot 1993](https://opg.optica.org/josaa/abstract.cfm?uri=josaa-10-5-777)). Below we load some examples.
 
 ```{code-cell} ipython3
 # Load and display a set of visual textures
@@ -95,6 +93,14 @@ display_images(im_files, "Artificial textures")
 im_files = [IMG_PATH / f"fig{num}.jpg" for num in hand_drawn]
 display_images(im_files, "Hand-drawn / computer-generated textures")
 ```
+
+Why do vision scientists care about visual textures? Quoting from the beginning of {cite:alp}`Portilla2000-param-textur`:
+
+> Vision is the process of extracting information from the images that enter the eye. The set of all possible images is vast, and yet only a small fraction of these are likely to be encountered in a natural setting... Nevertheless, it has proven difficult to characterize this set of "natural" images, using either deterministic or statistical models. The class of images that we commonly call "visual texture" seems most amenable to statistical modeling. Loosely speaking, texture images are spatially homogeneous and consist of repeated elements, often subject to some randomization in their location, size, color, orientation, etc. Julesz pioneered the statistical characterization of textures by hypothesizing that the Nth-order joint empirical densities of image pixels (for some unspecified N), could be used to partition textures into classes that are preattentively indistinguishable to a human observer (Julesz, 1962).
+
+The phrase "indistinguishable to a human observer" in that last sentence should grab your attention. As discussed in the [Metamer tutorial](metamer-nb), this is the same idea as perceptual metamers!
+
+In the Portilla-Simoncelli paper, the authors developed a model to measure the statistical properties of visual textures and then used metamer synthesis was used in conjunction with this model to demonstrate the necessity of different properties of the visual texture. Throughout the rest of the notebooks in this section, we will use some of these example textures to demonstrate aspects of the Portilla-Simoncelli model.
 
 ## Further reading
 


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/381

docs/tutorials/models/portilla_simoncelli/ps_basic_synthesis.md
```diff
@@ -0,0 +1,192 @@
...
+# so that relative sizes of axes created by po.imshow and others look right
+plt.rcParams["figure.dpi"] = 72
+
+# set seed for reproducibility
+po.tools.set_seed(1)
+```
+
+## A quick reminder of what metamers are and why we are calculating them
+
+The primary reason that the original {cite:alp}`Portilla2000-param-textur` paper developed the metamer procedure was to assess whether the model's understanding of textures matches that of humans. While developing the model, the authors originally evaluated it by performing texture classification on a then-standard dataset (i.e., "is this a piece of fur or a patch of grass?"). The model aced the test, with 100% accuracy. After an initial moment of elation, the authors decided to double-check and performed the same evaluation with a far simpler model, which used the steerable pyramid to compute oriented energy (the first stage of the model described here). That model also classified the textures with 100% accuracy. The authors interpreted this as their evaluation being too easy, and sought a method that would allow them to determine whether their model better matched human texture perception.
```
BalzaniEdoardo: this is a very good description. The only thing I would do, is move the definition of metamers early in the paragraph. Start with something like

"Model metamers are images... " and then describe the motivation behind the technique: "The primary reason that the original Portilla and Simoncelli, 2000 paper developed this metamer...".

docs/tutorials/models/portilla_simoncelli/ps_basic_synthesis.md
```diff
@@ -0,0 +1,192 @@
...
+
+- {func}`~plenoptic.simulate.models.portilla_simoncelli.PortillaSimoncelli.convert_to_dict` will turn the representation tensor into a dictionary with informative keys:
+
+```{code-cell} ipython3
+rep_dict = model.convert_to_dict(representation)
+for k, v in rep_dict.items():
+    print(k, v.shape)
+```
+
+    The statistics have also been reshaped so that e.g., the `auto_correlation_magnitude` representation has shape `(batch, channel, spatial_corr_width, spatial_corr_width, n_orientations, n_scales)`, which might make understanding the values easier.
```
BalzaniEdoardo: This renders weirdly, I suspect it is the initial tab.
```suggestion
The statistics have also been reshaped so that e.g., the `auto_correlation_magnitude` representation has shape `(batch, channel, spatial_corr_width, spatial_corr_width, n_orientations, n_scales)`, which might make understanding the values easier.
```

docs/tutorials/models/portilla_simoncelli/ps_optimization.md
```diff
@@ -0,0 +1,114 @@
...
+
+In order to make plenoptic's synthesis perform similarly, we need to make this statistic more important than the others. We do this by using a custom loss function which reweights the representation of each image before computing the L2-norm of the difference these representations. We do this using the {func}`~plenoptic.tools.optim.portilla_simoncelli_loss_factory` function, which takes in an instance of the {class}`~plenoptic.simulate.models.portilla_simoncelli.PortillaSimoncelli` model and an image, and returns a loss function. (This is necessary because the shape of the model output depends on the parameters used to initialize it.)
```
BalzaniEdoardo:
```suggestion
In order to make plenoptic's synthesis perform similarly, we need to make this statistic more important than the others. We do this by using a custom loss function which reweights the representation of each image before computing the L2-norm of the difference these representations, see `~plenoptic.tools.optim.portilla_simoncelli_loss_factory` for more details. 
```

I would trim this down

docs/tutorials/models/portilla_simoncelli/ps_matlab_differences.md
```diff
@@ -0,0 +1,386 @@
...
+
+    Why does this matter? We have qualitatively reproduced the results but cannot guarantee exact reproducibility. This means that, in general, metamers synthesized by the two versions will differ, though we believe that the metamers synthesized by plenoptic are better (see [](ps-optimization) for more discussion).
+
+    In particular, in order for synthesis to find good results, we use a custom loss function that reweights the model's output. This is described in more detail in the [](ps-optimization) notebook, but in short: we upweight the variance of the highpass residuals (which are matched very quickly with very high precision by the matlab algorithm) and remove the pixel minimum and maximum from the gradient computation (the min and max have very strange gradients and are better captured through the range penalty that is built into the {class}`~plenoptic.synthesize.metamer.Metamer` class).
```
BalzaniEdoardo: instead if "strange gradients" I would say, they are non-differentiable.

src/plenoptic/tools/optim.py
```diff
@@ -159,3 +165,86 @@ def penalize_range(
...
+    Parameters
+    ----------
+    model
+        An instantiated
+        :class:`~plenoptic.simulate.models.portilla_simoncelli.PortillaSimoncelli`
+        model.
+    image
+        The target image for metamer synthesis, or an image with the same shape, dtype,
+        and device.
+    minmax_weight
```
BalzaniEdoardo: is there any other valid alternative? if not, I would remove the argument and always set to zero

BalzaniEdoardo (changes requested):
At an high-level, I think this is a lot of content. While all of it is valuable, currently it may feel overwhelming and hard to navigate. In particular, the current structure forces readers to jump between sections to get a complete picture of any single aspect. You could think of grouping your notes even further, to help out users focusing on specific session of interest:

**Getting Started**
- Model basics + How to synthesize metamers (merge theory intro with practical usage in one notebook)

**Understanding the Model**
- Model statistics + Example texture synthesis (combine theory with concrete examples)

**Technical Details** (all in one place)
- Optimization details
- Model limitations
- Differences with MATLAB implementation

**Advanced Applications**
- Other applications of the model

This way:
- Someone new gets theory + practice together immediately
- Someone familiar with theory can skip straight to "Technical Details"
- MATLAB users can go directly to the comparison
- The "Advanced Applications" stands alone for experienced users

The decision tree in the Portilla-Simoncelli section index would be also easier to parse: "New to texture models? Start with Getting Started. Familiar with Portilla-Simoncelli? Jump to Technical Details. Coming from MATLAB? See the comparison notebook."

docs/tutorials/models/portilla_simoncelli/ps_examples.md
```diff
@@ -0,0 +1,397 @@
...
+
+(See figure 13 of {cite:alp}`Portilla2000-param-textur`.)
+
+Excerpt from paper: _"Figure 13 shows two pairs of counterexamples that have been used to refute the Julesz conjecture. \[13a and 13b were \] originally created by Julesz et al. (1978): they have identical third-order pixel statistics, but are easily discriminated by human observers. Our model succeeds, in that it can reproduce the visual appearance of either of these textures. In particular, we have seen that the strongest statistical difference arises in the magnitude correlation statistics. The rightmost pair were constructed by Yellott (1993), to have identical sample autocorrelation. Again, our model does not confuse these, and can reproduce the visual appearance of either one."_
```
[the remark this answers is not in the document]
billbrod: added some more text here, let me know if it's enoguh

docs/tutorials/models/portilla_simoncelli/ps_limitations.md
```diff
@@ -0,0 +1,185 @@
...
+
+However, the above demonstrates that it is difficult to predict what a Portilla-Simoncelli metamer will look like for a **non-texture** target image. Therefore, one must be careful to ensure that their optimization procedure has converged to a low enough loss to be interpretable and that optimization has not failed because e.g., it got stuck in a local minimum. See [](ps-optimization) for further discussion.
+
+:::
+
+```{code-cell} ipython3
+po.imshow(
+    [met.image, met.metamer],
+    title=["Target image", "Synthesized Metamer"],
+    vrange="auto1",
```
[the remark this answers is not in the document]
billbrod: I'm not sure what you mean here -- this notebook just has the picture of Einstein and its PS model metamer. there is no brick wall texture here (or mention of it, as far as I can find)

docs/tutorials/models/portilla_simoncelli/ps_matlab_differences.md
```diff
@@ -0,0 +1,386 @@
...
+5) **Statistics not in paper**: The Matlab code outputs the mean magnitude of each band and cross-orientation real correlations, but these are not enumerated in the paper. These statistics are removed in `plenoptic`. See the next section for some more detail about the magnitude means.
+
+Note: This can be understood by thinking of $A_{i,0}$, the autocorrelation of every pixel and the pixel $i$ to their right. Computing this auto-covariance involves adding together all the products $I_{x,y}*I_{x+i,y}$ for every x and y in the image. But this is equivalent to computing $A_{-i,0}$, because every pair of two neighbors $i$ to the right $I_{x,y}*I_{x+i,y}$ is also a pair of neighbors $i$ to the left, $I_{x+i,y}*I_{(x+i)-i,y}=I_{x+i,y}*I_{x,y}$. So, any opposite displacements around the central element in the auto-covariance matrix will have the same value.
```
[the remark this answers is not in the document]
billbrod: moved to make it more obvious, hopefully

docs/tutorials/models/portilla_simoncelli/ps_matlab_differences.md
```diff
@@ -0,0 +1,386 @@
...
+# Sum cross-scale phase statistics
+phase_statistics_num = torch.sum(
+    ~torch.isnan(stats_dict["cross_scale_correlation_real"])
+)
+print(f"Phase statistics: {phase_statistics_num} parameters, compared to 96 in paper")
+```
+
+## Magnitude means
+
+The mean of each magnitude band are slightly different from the redundant statistics discussed in the previous section. Each of those statistics are exactly redundant, e.g., the center value of an autocorrelation matrix will always be 1. They thus cannot include any additional information. However, the magnitude means are only *approximately* redundant and thus could improve the texture representation. The authors excluded these values because they did not seem to be necessary: the magnitude means are constrained by the other statistics (though not perfectly), and thus including them does not improve the visual quality of the synthesized textures.
```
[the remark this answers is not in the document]
billbrod: how's this?

BalzaniEdoardo:
> @BalzaniEdoardo I think I've gone ahead and addressed all your feedback.
> 
> In your [first comment](https://github.com/plenoptic-org/plenoptic/pull/381#pullrequestreview-3500589959), it sounded like you were proposing I combine the notebooks back into fewer larger notebooks. I'm not sure I wanto do that -- part of the goal here was to make the notebooks smaller, because previously they were all in one giant notebook. I did arrange them on the index page like you suggested, what do you think? I'd like a second opinion on whether I should combine the notebooks again -- @ershook?

no, I meant to keep stuff separated but add levels in the site tree (like group notebooks into categories):
background:
 - notebooks 1,2,3,4... etc
optimization:
 - notebook 5,6,7 etc

to form a nested navigation bar instead of a flat overwhelming one with a lot of notebook at the same level.

> In addition to your comments on the docs, I:
> 
>     * added the FutureWarning about loading, and added an explanation in the docs. I think, in the future, I will add the ability for users to skip these checks ([Add `skip_checks` to Synthesis load methods #384](https://github.com/plenoptic-org/plenoptic/issues/384)), if they're certain, but that's out of scope for this PR.

sounds good! 

>     * `portilla_simoncelli_loss_factory`:
>       
>       * will raise warnings/errors if the model's output looks unexpected.
>       * has now a single argument, `reweighting_dict`, which users can use to further change the weighting on the stats.

I like this

>     * I added a new loss factory, `groupwise_relative_l2_norm_factory`. this isn't used in this PR (I wrote it because I thought it might do better than the PS-specific loss factory, but it didn't), but I _do_ think it's useful, so I'm including it and I mention it in the docs.

I agree it is a good addition.

billbrod:
@ershook I made some small changes to the PS index and intro pages in response to your suggestions, take a look and let me know what you think!

BalzaniEdoardo:
> > no, I meant to keep stuff separated but add levels in the site tree (like group notebooks into categories):
> > background:
> > notebooks 1,2,3,4... etc
> > optimization:
> > notebook 5,6,7 etc
> > to form a nested navigation bar instead of a flat overwhelming one with a lot of notebook at the same level.
> 
> I see. That is what I [tried to do](https://docs.plenoptic.org/docs/pulls/381/tutorials/models/portilla_simoncelli/ps_index.html). Is that what you were thinking, or something more?

Yes, breaking down the navi in those same sections. if you look at navi in side bar, all section are listed at the same level. Tomorrow I can show you what I mean if this is not clear!

PR #383, opened 2025-12-01
Add regularization to synthesis methods via `penalty_function`

### LINKED ISSUE #363: Add regularization functionality to synthesis methods
Issue to track addition of new regularization feature.

**Goal:** Add regularization functionality to constrain image synthesis (e.g. constrain the amplitude spectra, or the pixel histogram).

**Approach:**  For classes that optimize a model-based `loss_function()` (i.e. metamer and mad competition), we add a separate `penalty_function()` implementing the constraint. `penalty_function()` can be passed in class initialization. It takes the synthesized image as input and returns a penalization term.

**Default behavior:** Default behavior is the same as before.  The default `penalty_function()` is `penalize_range`, which was used to constrain the pixel range before this feature.

**Other changes:** We also add a new module, `plenoptic.tools.regularization` where to have standard functions for regularization. `penalize_range` is moved to the `regularization` module.

dherrera1911:

This PR adds basic functionality to include regularization in the synthesis objects `Metamer` and `MADCompetition`

This closes Issue #363

- Should we add a test for the case when an invalid `penalty_function` is passed? E.g. requiring more than one input
- Any missing tests for the functionality? Currently only save/load tests were added, I think.
- There is a need for an example of the new functionality in the documentation. Should this be done in this PR, or in a separate one? Can use feedback of where to add this, and what new `penalty_function` to use as an example.
- I need to check whether other places of current documentation need adjusting. LMK if you already know some.

**Describe changes**

- Removed old parameters `range_penalty_lambda` and `allowed_range` as inputs to `Metamer` and `MADCompetition`
- Added parameters `penalty_function` and `penalty_lambda`
- `penalty_function` defaults to the old behavior
- Added tests for save/load of new parameters
- Hard-coded `allowed_range` in `validate_input`
- Small changes to the tests, to pass without `allowed_range` in some cases
- Created new module `po.tools.regularization`, moved `penalize_range` there from `po.tools.optim`

**Checklist**

- [X ] I have described the changes in this PR, following the template above.
- [ X] I have added any necessary tests.
- [ ] I have added any necessary documentation. This includes docstrings, updates to existing files found in `docs/`, or (for large changes) adding new files to the `docs/` folder.


dherrera1911:
I revised the `allowed_range` thing. To achieve what I interpreted as the desired behavior I:
- Removed `allowed_range` from `validate_input`.
- Added a bool `check_range` to `validate_input`. When `True` (the default), a `UserWarning` is raised if the input is outside of the `(0, 1)`  range. Synthesis classes check the range.
-  I set the `PortillaSimoncelli` class to not check the range. Many tests from `test_models.py` failed otherwise.

@billbrod verify that this is the behavior you had in mind.

I'll work on the other pending issues next.

dherrera1911:
@billbrod, about the question of how to input the `penalty_lambda`, I'm inclined to keep it in the main class, although I can be convinced otherwise.

From my short experience working on this problem, it seams that the penalty weight has a significant effect on the synthesis results for some penalties (e.g. when I tried to do pixel histogram matching). Thus, I think that tweaking the `penalty_lambda` will be common for users of regularization. If this is the case, I weight ease of use more than making the class cleaner. It might also allow the user to easily do things like start with a stronger penalty_lambda and then reduce it later in the synthesis.

I'm happy to discuss it more offline.

ershook:
> > @billbrod, about the question of how to input the `penalty_lambda`, I'm inclined to keep it in the main class, although I can be convinced otherwise.
> 
> Another possibility is to make the penalty functions factories, like the new `portilla_simoncelli_loss_factory` in #381: that is, it's a function that returns another function. If one of its arguments is the lambda, then the Metamer init would look like:
> 
> ```python
>     def __init__(
>         self,
>         image: Tensor,
>         model: torch.nn.Module,
>         loss_function: Callable[[Tensor, Tensor], Tensor] = optim.mse,
>         range_penalty:  Callable[[Tensor], Tensor] = penalize_range_factory(allowed_range=(0, 1), penalty_lambda=.1),
>     ):
> ```
> 
> `penalize_range_factory` would be a function with two arguments, `allowed_range` and `penalty_lambda`, and it returns a function that just accepts a single argument (the image) and returns a scalar.
> 
> This accomplishes what I wanted from using lambda/partials, but is, I think, more accessible. What do you think? @ershook ?
> 
> > From my short experience working on this problem, it seams that the penalty weight has a significant effect on the synthesis results for some penalties (e.g. when I tried to do pixel histogram matching). Thus, I think that tweaking the `penalty_lambda` will be common for users of regularization. If this is the case, I weight ease of use more than making the class cleaner.
> 
> I think you're right that tweaking the lambda will be necessary for the more complex penalties (or at least, it will be for us as we try to understand them better). But I _don't_ think tweaking lambda will be very necessary if you're using the default penalty. So the case where you need to that tweaking is one in which you're already setting the penalty function, so I don't think it's more complicated to tweak an input to that function. That is, I don't think doing
> 
> ```python
> for l in lambda_vals:
>     Metamer(img, model, penalty_function=my_cool_penalty(penalty_lambda=l))
> ```
> 
> (using function factories) is that different from
> 
> ```python
> for l in lambda_vals:
>     Metamer(img, model, penalty_function=my_cool_penalty, penalty_lambda=l)
> ```
> 
>  > It might also allow the user to easily do things like start with a stronger penalty_lambda and then reduce it later in the synthesis.
> 
> I don't think any of the setups we're discussing allow this, actually. In all of them, the only way to do this would be to initialize a new `Metamer` object and initialize it with the `metamer` attribute from the previous object. For now, I don't think it's worth the complexity to make this easier. But if we do want to look into it, we should look into how torch's learning rate schedulers work (and we do support these, for changing the optimizer's learning rate)
> 
> > I'm happy to discuss it more offline.
> 
> Same. If you don't find my argument for using loss factories convincing, let's find a time to chat.

I like the idea of having the range penalty be implemented the same as the  loss factory  for consistency sake. I also think it is more accessible @billbrod.

dherrera1911:
@billbrod the contribution is getting tighter. There's a couple of things in the TODO list still:
- [ ]: Write a function to validate `penalty_function`
- [ ]: Write some new tests for the `penalty_function` functionality
- [ ]: Maybe sort out issue of saving/loading function names and `functools.partial` (or leave for another smaller PR)
- [ ]: Add explanation to the docs of how to use new functionality

I think before doing the above, we should decide  what to do with `penalty_lambda`. I am still in favor of keeping `penalty_lambda` as is.

You mentioned that simplifying the Metamers & MADCompetition docstring is a motivation for removing `penalty_lambda`. But having a `penalty_function` factory would make the documentation so much more complicated somewhere else in the package! And probably add extra complexity to the testing suite too.

Also, I think of tweaking the `penalty_lambda` as potentially being more common than you suggest. E.g. different user-defined models will have different magnitudes of the loss, and it might be necessary to adjust `penalty_lambda` for each case. Also, while most users will likely just use the default regularization, if plenoptic offers a nice suite of regularization functions, it might become a more frequently used functionality with non-savvy users.

We should chat to make a decision.

**Small things to solve:**
- Have `PortillaSimoncelli` not raise range warning on forward pass. I think this is what makes jenkins fail right now. Also the new branch might be raising lots of warnings in PS notebooks.
- Use of `clamp()` in `MADCompetition` setup

billbrod:
@dherrera1911 after discussing this with @BalzaniEdoardo and @gviejo , I think you're correct: the increased usability of including `penalty_lambda` as an explicit argument for Metamer and MADCompetition is more important than the simplicity + consistency of using function factories.

docs/tutorials/intro/Metamer.md
```diff
@@ -209,6 +209,57 @@ met.synthesize()
...
+that control other properties of the synthesized metamer.
+A simple example can be a modified range penalty that uses a different allowed range.
+For this, we can use `functools.partial` to create a new function that penalizes pixels
+outside a different range, as shown below.
+
+```{code-cell} ipython3
+import functools
+
+# Create custom_penalty function, that penalizes pixels outside of [0.2, 0.8] range
+custom_penalty = functools.partial(
```
billbrod: see top-level comment: this should be a def, not a partial

docs/tutorials/intro/Metamer.md
```diff
@@ -209,6 +209,57 @@ met.synthesize()
...
+print(
+    "Range of synthesized metamer: "
+    f" {met.metamer.min().item():.3f} to {met.metamer.max().item():.3f}"
+)
+```
+
+The {class}`Metamer <plenoptic.synthesize.metamer.Metamer>` class also has a
+`penalty_lambda` argument, that weighs the contribution of the penalty function
+to optimization. The `penalty_function` argument is also present in the
+{class}`MADCompetition <plenoptic.synthesize.mad_competition.MADCompetition>` class.
```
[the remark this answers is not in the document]
billbrod: this doc page is about metamer. 

instead, in the section where you set `penalty_function` in the MAD Competition doc, add a sentence like: "for more details about `penalty_function`, see [here]` and link to this section

src/plenoptic/simulate/models/portilla_simoncelli.py
```diff
@@ -394,7 +393,8 @@ def forward(self, image: Tensor, scales: list[SCALES_TYPE] | None = None) -> Ten
         >>> representation_tensor.shape
         torch.Size([1, 1, 1046])
         """
-        validate_input(image)
+        if image.ndimension() != 4:
```
billbrod: actually, looks like steerable pyramid already does this check. so instead, remove this check here and replace the assert in steerable pyramid's forward (line 443) with this if/raise.

(and I added #401, which I think should be possible)

src/plenoptic/synthesize/mad_competition.py
```diff
@@ -59,11 +64,11 @@ class MADCompetition(OptimizedSynthesis):
...
         ``optimized_metric`` loss. If ``None``, we pick a value so the two
         initial losses are approximately equal in magnitude.
-    range_penalty_lambda
-        Lambda to multiply by range penalty and add to loss.
-    allowed_range
-        Range (inclusive) of allowed pixel values. Any values outside this
-        range will be penalized.
+    penalty_function
+        A penalty function applied to the synthesized image, that returns a scalar
+        penalty to be minimized. It constrains specific properties of the metamer.
```
billbrod:
```suggestion
        penalty to be minimized.
```

src/plenoptic/synthesize/metamer.py
```diff
@@ -830,12 +834,11 @@ class MetamerCTF(Metamer):
...
         in order to determine their loss.
-    range_penalty_lambda
-        Strength of the regularizer that enforces the allowed_range. Must be
-        non-negative.
-    allowed_range
-        Range (inclusive) of allowed pixel values. Any values outside this
-        range will be penalized.
+    penalty_function
+        A penalty function applied to the metamer, that returns a penalty
+        scalar to be minimized. It constrains specific properties of the metamer.
```
billbrod:
```suggestion
        scalar to be minimized. 
```

src/plenoptic/synthesize/synthesis.py
```diff
@@ -274,13 +286,33 @@ def load(
                     "saved object futureproof and avoid this warning.",
                     category=FutureWarning,
                 )
-            else:
-                init_not_save_str = "\n ".join(
-                    [f"{k}: {getattr(self, k)}" for k in init_not_save]
+            penalty_missing = {"penalty_function", "_penalty_lambda"} & init_not_save
```
billbrod: `&` is intersect for sets? did not realize that. I would prefer the use of the `intersection` method, which is easier to read.

src/plenoptic/synthesize/synthesis.py
```diff
@@ -274,13 +286,33 @@ def load(
...
+            if penalty_missing:
+                # in PR #383, we added penalty_function and penalty_lambda attributes,
+                # which we'll handle for now, but warn about.
+                # Remove allowed_range and range_penalty_lambda so there's no extra key
+                # in saved dictionary
+                allowed_range = tmp_dict.pop("_allowed_range")
```
billbrod: not having a default here makes me worried, from a "what if I mess this up in the future" perspective. 

```suggestion
                allowed_range = tmp_dict.pop("_allowed_range", (0,1))
```

src/plenoptic/tools/validate.py
```diff
@@ -66,15 +62,11 @@ def validate_input(
     >>> import plenoptic as po
     >>> po.tools.validate.validate_input(po.data.einstein())
 
-    Intentionally fail:
+    Raise warning:
```
billbrod: instead, have this intentionally fail because of the dtype

billbrod (changes requested):
Thanks for your work Daniel!

Looking through this, I'm realizing I got confused as to where we ended up with partials. From what you have here, my understanding is, in order to change the allowed_range, users will have to either use partials or define their own custom penalty like

```python
def custom_penalty(x):
    return po.tools.regularization.penalize_range(x, allowed_range=(.2, .8))
```

If they use partials, they will be unable to save/load, right (because of the issue with `_get_name`)? In that case, I do not think we should tell people to use partials, since that's a weird sharp edge.

Instead, I think the docs / examples should show people how to write a custom penalty like above. Advantage here, is it should be easy to see that you can easily combine two penalties in that new function (and the docs should say so, explicitly).

The reason to use partials is if you're lazy (like me) and want to do it in the same line where you pass it to the init method: 
```python
po.synth.Metamer(
    img,
    model,
    penalty_function=functools.partial(po.tools.regularization.penalize_range, allowed_range=(0.0, 255.0)
)
```

And I don't think that's worth telling people about -- advanced users will realize they can do that based on the custom def example, and otherwise we should explain (or at least link) to what partials mean for non-advanced users

Other notes:
- I'm not sure what to do for the description of `penalty_function` in the docstrings. I don't think "It constrains certain properties of the metamer" is that informative, but without that sentence, it reads fairly abstract. What about something like "For example, this can penalize all values that lie outside of a specified range"?
- Need to add a paragraph to the `reproducibility.md` docs page about the change to penalty_function. should be pretty straightforward, more or less a copy of the existing paragraph under `FutureWarning in load in plenoptic 1.4`
- Metamer and MADCompetition's init docstrings should have examples setting penalty_function

For the failing PS synthesis tests -- yes, I'll go ahead and update these once we've otherwise finished with this PR.

src/plenoptic/synthesize/mad_competition.py
dherrera1911: For the docstring, what about:
```
penalty_function
        A function applied to the metamer during optimization, that returns
        a scalar penalty to be minimized. By penalizing certain properties of
        the image, like pixels values outside an allowed range, we can constrain
        those image properties.
```
I am not in love with the current description, so very open to suggestions.

docs/tutorials/intro/MAD_Competition_1.md
```diff
@@ -208,6 +213,11 @@ po.imshow(img, vrange=(0, 255), zoom=4);
 Now we'll do the same process of running synthesis and checking our loss as above:
 
 ```{code-cell} ipython3
+range_penalty_custom = functools.partial(
```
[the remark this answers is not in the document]
dherrera1911: Agree.

Originally I only passed the raw `penalty_function=po.tools.regularization.penalize_range`. What happened was that in the save/load tests, I was using `functools.partial` to generate different scenarios, but some things didn't work because of the name mismatch.

The function name validation does pass if two functions are defined with `functools.partial`, even if they are different functions. So, to make the tests run, I used `functools.partial` in the classes.

I will remove `functools.partial` from the main code, but then I'll also have to tweak how penalty functions are defined in the save/load tests.

src/plenoptic/synthesize/mad_competition.py
```diff
@@ -18,11 +19,15 @@
 from torch import Tensor
 from tqdm.auto import tqdm
 
-from ..tools import data, display, optim
+from ..tools import data, display, regularization
 from ..tools.convergence import loss_convergence
 from ..tools.validate import validate_input, validate_metric
 from .synthesis import OptimizedSynthesis
 
+default_penalty = functools.partial(
```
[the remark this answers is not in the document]
dherrera1911: As mentioned in a comment above (https://github.com/plenoptic-org/plenoptic/pull/383#discussion_r2734704531), `_get_name` actually passes if two functions are defined with `functools.partial`, even if they are different functions with different names. I added this here to help the save/load tests pass, but we should change this.

I can either, pass `po.tools.regularization.penalize_range` directly, or have a `def _default_penalty(image):` in here, which might help with save/load telling whether the default penalty was changed. I need to look more closely at the tests and save/load to figure out the options and possible cases.

docs/tutorials/intro/MAD_Competition_1.md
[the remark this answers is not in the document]
dherrera1911: Maybe I misremember some specifics.

In the tests, if two functions were different and produce different outputs, save/load always failed.

My recollection of the problem is that in some cases, we had the same function, but defined in different ways (e.g. `functools.partial` vs straight `regularization.penalize_range`), and that made the tests fail. Which I agree, was not necessarily bad from a user perspective, but added some wrinkles to the tests.

I'll look more closely at what was the issue and come back to you.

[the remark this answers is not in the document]
billbrod: Thanks for looking into this more. This is more evidence that the way I'm currently handling load is overly conservative -- can you add your long comment above to #384?

This is my understanding of how things work:
- during load, we check whether the function name (but not the module path) of penalty_function is identical. So e.g., they must both be called `penalize_range`, but can be defined in different places (within `load()`, in `po.tools.optim`, in `po.tools.regularization`, in the interpreter, etc.).
- the outputs of the saved and loading version must be identical. this catches whether `allowed_range` was the same in both cases.

Thus, we can have the default for `Metamer` be the standard `po.tools.regularization.penalize_range` (which thus defaults to `(0,1)`). Then, for backwards compatibility, when we load an older `Metamer` object, we locally (within `load()`, where you currently define the partial) define a new `penalize_range` that uses the saved `allowed_range`, so that would look like:

line 296 in synthesis.py:
```diff
- penalty_fn = functools.partial(
-                                penalize_range, allowed_range=allowed_range)
- 
- tmp_dict["penalty_function"] = (
-                                 _get_name(penalty_fn),
-                                 ("_image",),
-                                 penalty_fn(tmp_dict["_image"]),
- )
+ def penalize_range(x):
+     return regularization.penalize_range(x, allowed_range)
+ 
+ tmp_dict["penalty_function"] = (
+                                 _get_name(penalize_range),
+                                 ("_image",),
+                                 penalize_range(tmp_dict["_image"]),
+ )
```

(note this requires not importing `penalize_range` directly into `synthesis.py`, but importing `regularization`)

and this should have a comment pointing to this PR (and this comment thread)

And yes, eventually we would remove this, once we remove this backwards compatibility.

Does that all make sense? If it's not clear or if I've misunderstood something, let's hop on a call to talk it through.

Looking back at the relevant PR (#317), the reason I have the name check is because pytorch allows one optimizer to load the state dict of another:
> I was originally going to only raise a warning if the name was different and then e.g., have the optimizer load the state dict and try to continue. But pytorch let's you load another optimizer's state dict. So I can do `SGD.load_state_dict(Adam.state_dict())` without a problem, but then when I try to step the optimizer, I get a confusing error message (because it can't find the relevant information). To avoid this problem, I just decided that names have to be stable.

src/plenoptic/tools/validate.py
dherrera1911: You mean, change "Raise warning:" for "Intentionally fail because of the dtype:"?

PR #392, opened 2026-01-13
Adds doctests to `metamer.py`, `portilla_simoncelli.py`; make `pooch` required

billbrod:

Adds doctests to `metamer.py`, finishes adding them to `portilla_simoncelli.py`

The new examples can be found at [metamer](https://docs.plenoptic.org/docs/pulls/392/api/plenoptic.synthesize.html#module-plenoptic.synthesize.metamer) and [portilla_simoncelli](https://docs.plenoptic.org/docs/pulls/392/api/plenoptic.simulate.models.html#plenoptic.simulate.models.portilla_simoncelli.PortillaSimoncelli)

- How do these examples look?
- The main dilemma I faced with adding them to `metamer.py` is that almost all examples there involve calling `synthesize()`: it doesn't make sense to e.g., call `display_metamer` on a `Metamer` object if you haven't first synthesized the metamer. However, this makes all the tests take much longer! I could do what I do in the PortillaSimoncelli documentation (#381) and save a single object and then load it in for each example, but that's not how people will generally be using these functions. So I instead I have generally used a simple model (so each iteration is quick) and only synthesized for 5 iterations. Is that reasonable? Is there a better way to do this?

**Describe changes**

- Using pytest to run these tests in parallel is slower than in sequence, so change our run doctests CI to use `-n 0`. Haven't tracked down why this is.
    - Relatedly, add `OMP_NUM_THREADS=1` and `torch.set_num_threads(1)` to sphinx config (already in pytest config)
- There was an annoying issue with matplotlib figures persisting across doctests. The [matplotlib plot_directive](https://matplotlib.org/stable/api/sphinxext_plot_directive_api.html) has an option to fix this (`:context: reset`), but pytest ignores it. Thus some tests will fail depending on the tests run before them (e.g., `display_metamer` will fail if the tests from `MetamerCTF` are run beforehand, but not if run alone or with the `plot_*` and `animate` functions). I couldn't find an easy way to handle this (I tried writing a pytest fixture with teardown code that closed all open figures, but that didn't work), so for now I'm going to just try and write the tests to handle this situation correctly.
- Describes above in CONTRIBUTING
- Change default `PortillaSimoncelli.plot_representation` figsize, since I was always changing it in practice. Updates where this is called in docs / doctests
- Eventually, I'm going to have numpydoc's linter check for Examples. To that end, I've gone ahead and put `numpydoc ignore=EX01` on docstrings which won't have examples (modules and private / internal functions) in modules that have already had examples added elsewhere (i.e., everything in `simulate/`, `metric/`, `data/`, and `metamer.py`).
- Removes `doctest: +ELLIPSIS` from a couple doctests, because that option is in pyproject.toml and so redundant

**Checklist**

### DIFF
--- a/src/plenoptic/simulate/models/portilla_simoncelli.py
+++ b/src/plenoptic/simulate/models/portilla_simoncelli.py
@@ -100,11 +100,14 @@ class PortillaSimoncelli(nn.Module):
     --------
     Compute texture statistics of an image:
 
-    >>> import plenoptic as po
-    >>> img = po.data.reptile_skin()
-    >>> ps_model = po.simul.PortillaSimoncelli(img.shape[2:])
-    >>> ps_model(img)
-    tensor([[[0.4172, 0.0547, ..., 0.0048]]])
+    .. plot::
+      :context: reset
+
+       >>> import plenoptic as po
+       >>> img = po.data.reptile_skin()
+       >>> ps_model = po.simul.PortillaSimoncelli(img.shape[2:])
+       >>> ps_model(img)
+       tensor([[[0.4172, 0.0547, ..., 0.0048]]])
 
     Visualize texture statistics:
 


ershook:
I think these examples look great. I especially like how you showed the difference between objective function versus loss. Previously I had found confusing how to change optimizers and the example showing how to set the optimizer and scheduler would have solved this. Only note would be to maybe mention the fact that store_progress = True is useful later when you show the examples of saving out videos of the optimization process.

billbrod:
> Some general comments I have:
> 
> Some of the examples can be narrowed in scope, and don't need to show the whole pipeline, e.g.
>       - `load` doesn't need to also show saving first, maybe you can have some static file that is loaded in
>       - I don't think `setup` examples all need to call `synthesize`
>       - `objective_function` is an exception to this, where I think it relies on showing it's use at different points in the pipeline
> 
> I would avoid repeating the import strings within a function as it adds a lot of visual clutter, unless there's something that's really important to have a self-contained code block
> 
> you can also reduce `synthesize` calls by reusing a single model fit within a function
> 
> some examples are quite long with inline comments (e.g. in `objective_function`) that I think can be split up, with the comments outside of the code block to highlight what is being demonstrated (this doesn't need to happen for every inline comment, e.g. the comment that says the number of iterations isn't enough for convergence)

~I agree with all of these (and made the changes) except load not showing save. Like you said, there are multiple ways I can do this: either I show save or I load in an already existing file. For an already existing file, it would have to be either created elsewhere in tests (in which case the user can't easily reproduce the behavior locally and the tests might fail when parallelized or anything else that changes which tests are run or their order) or I'd have to download it using `fetch_data` (the way I do in the PortillaSimoncelli model notebooks). I don't like the download option, because I want the doctest to show users the way they will interact with the function. So I'm going to stick with this for now.~

~There's a similar issue with the plotting/animating functions: they all start with a brief initialization and synthesis, but could all just load something pre-stored. For similar reasons, I kept them as is, but I'm going to ask for opinions (on both) from pyOpenSci review (but we shouldn't wait for this PR).~

Actually, after writing the above and before posting, I checked what other libraries do in their doctests. [scipy's loadmat](https://docs.scipy.org/doc/scipy-1.16.2/reference/generated/scipy.io.loadmat.html) and [arviz's plot_forest](https://python.arviz.org/en/stable/api/generated/arviz.plot_forest.html#arviz.plot_forest) both load in / download example data to use. And, like you mentioned, it will reduce clutter and focus on the specific usage of that method, which is the goal of the examples section.

Okay, so I'm going to go through and remove even more synthesize calls. One note is that I think this means I should make pooch a required dependency, since it's needed to download examples. But pooch is light-weight, so I think that's fine.

PR #405, opened 2026-01-27
Updates jenkins torch version 2.9.1 -> 2.10.0

billbrod:

Updates `jenkins/Dockerfile` pytorch image to 2.10.0, most recent version on pypi.

Also updates the python version jenkins uses to 3.12 (from 3.11). This is not because of the pytorch package (which still supports back to 3.10), but the pytorch image, which is now [based on ubuntu 24.04](https://hub.docker.com/layers/pytorch/pytorch/2.10.0-cuda12.6-cudnn9-runtime/images/sha256-4ec910830ec6c3b837d5621b6f995a31638d3b0824bf0ed4fcaeba7dd4f89b48), which no longer includes the [python3.11](https://launchpad.net/ubuntu/+source/python3.11) in its default sources for apt (jammy jellyfish is 22.04 LTS, 24.04 LTS is noble numbat, which includes [python3.12](https://launchpad.net/ubuntu/+source/python3.12), for more on ubuntu naming conventions see [wikipedia](https://en.wikipedia.org/wiki/Ubuntu_version_history)).

Additionally, not sure why this happened now, but this upgrade also caused us to hit a known issue with the [cache](https://github.com/pytorch/pytorch/issues/140765), because our Jenkins CI run in containers. The fix is to set the `TORCHINDUCTOR_CACHE_DIR` to a specific path (I used `/tmp/torchinductor`, but specifics shouldn't matter).

**Checklist**

- [x] If a public new class or function was added: I have double-checked that it is present in the API docs, modifying `docs/api.rst` and `docs/api_modules.rst` to include it if necessary.

### DIFF
--- a/jenkins/Dockerfile
+++ b/jenkins/Dockerfile
@@ -1,4 +1,4 @@
-FROM pytorch/pytorch:2.11.0-cuda13.0-cudnn9-runtime
+FROM pytorch/pytorch:2.10.0-cuda12.8-cudnn9-runtime
 
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \


billbrod:
Docs build was failing with an add CUDA initialization error during the running of the matplotlib plot directives in the docstrings:


Turns out, it was an issue with parallelizing the sphinx build (probably could've been able to tell by the fact that the above error showed up 3 times). Changing `-j 4` back to `-j 1` fixes this problem, and the build is only a little slower (previously, 17 or 18 minutes, now 20), so we'll go with this for now.

billbrod:
pytorch 2.11.0 has been released, but it's not working right now on our Jenkins nodes (which have older V100s), so I'm sticking with 2.10.0 for the time being.

PR #408, opened 2026-01-28
Docstring examples for `tools/signal.py`

wulfdewolf:
This PR addresses part of #237  by adding docstring examples to the public functions in `tools/signal.py`.

- [x] rescale
- [x] rectangular_to_polar
- [x] polar_to_rectangular
- [x] make_disk
- [x] add_noise
- [x] modulate_phase
- [x] autocorrelation
- [x] center_crop
- [x] expand
- [x] shrink

### DIFF
--- a/src/plenoptic/tools/signal.py
+++ b/src/plenoptic/tools/signal.py
@@ -247,7 +247,7 @@ def polar_to_rectangular(amplitude: Tensor, phase: Tensor) -> Tensor:
     torch.Size([1, 1, 256, 256])
     >>> phase.shape
     torch.Size([1, 1, 256, 256])
-    >>> # and from those, we can use this function to recoverthe rectangular coordinates
+    >>> # from those, we can use this function to recover the rectangular coordinates
     >>> rectangular_coeff = po.tools.polar_to_rectangular(amplitude, phase)
     >>> rectangular_coeff.shape
     torch.Size([1, 1, 256, 256])
@@ -891,8 +891,8 @@ def shrink(x: Tensor, factor: int) -> Tensor:
       ... )
       <PyrFigure...>
 
-    You can undo a shrink using :func:`~plenoptic.tools.signal.expand`, but not exactly.
-    Shrinking discards information that can not be recovered:
+    You can invert ``shrink`` using :func:`~plenoptic.tools.signal.expand`, but the
+    inversion is not perfect; shrinking discards information that can not be recovered:
 
     .. plot::
       :context: close-figs
@@ -908,8 +908,9 @@ def shrink(x: Tensor, factor: int) -> Tensor:
       ... )
       <PyrFigure...>
 
-    Even in the opposite order, i.e. shrinking an expanded image, we can only recover
-    the original image with a low tolerance:
+    Even in the opposite order, i.e., shrinking an expanded image, the inversion is
+    not perfect. In this example with pixel values between 0 and 1,
+    there are differences on the order of 1e-3:
 
     .. plot::
       :context: close-figs
@@ -922,8 +923,8 @@ def shrink(x: Tensor, factor: int) -> Tensor:
       >>> torch.allclose(img, shrink_after_expand, atol=1e-3)
       False
       >>> po.imshow(
-      ...     [img, shrink_after_expand],
-      ...     title=["original", "shrink after expand"],
+      ...     [img, shrink_after_expand, img - shrink_after_expand],
+      ...     title=["original", "shrink after expand", "difference"],
       ... )
       <PyrFigure...>
     """


src/plenoptic/tools/signal.py
```diff
@@ -167,6 +188,15 @@ def polar_to_rectangular(amplitude: Tensor, phase: Tensor) -> Tensor:
...
         The analogous function for real-valued signals.
+
+    Examples
+    --------
+    >>> import plenoptic as po
+    >>> import torch
+    >>> amplitude = torch.tensor([1.4142, 1.4142])
+    >>> phase = torch.tensor([0.7854, -0.7854])
+    >>> po.tools.polar_to_rectangular(amplitude, phase)
+    tensor([1.+1.j, 1.-1.j])
```
billbrod: can you add an example using one of the steerable pyramid complex coefficients?

something like:
```python
img = po.simul.einstein()
spyr = po.simul.SteerablePyramidFreq(img.shape[-2:], is_complex=True)
coeff = spyr(img)[0]
rect_coeff = polar_to_rectangular(coeff)
po.imshow(coeff)
po.imshow(rect_coeff, title=["vertical", "horizontal"])
```

src/plenoptic/tools/signal.py
```diff
@@ -301,7 +331,17 @@ def make_disk(
...
+        Tensor mask with shape ``torch.Size(img_size)``.
+
+    Examples
+    --------
+    .. plot::
+
+      >>> import plenoptic as po
+      >>> import torch
+      >>> disk = po.tools.make_disk((128, 128), outer_radius=50, inner_radius=25)
+      >>> po.imshow(disk[None, None], cmap="gray")
```
billbrod: I worry `[None, None]` is a bit obscure for people. add a comment say this is equivalent to `.unsqueeze(0)` twice?

also, add something that shows using it as a mask:

```python
disk = disk[None, None]
img = po.data.einstein()
po.imshow(disk * img)
```

something like that

src/plenoptic/tools/signal.py
```diff
@@ -437,6 +527,23 @@ def autocorrelation(x: Tensor) -> Tensor:
...
+      >>> import plenoptic as po
+      >>> import torch
+      >>> img = po.data.einstein()
+      >>> img.shape
+      torch.Size([1, 1, 256, 256])
+      >>> ac = po.tools.autocorrelation(img)
+      >>> ac.shape
+      torch.Size([1, 1, 256, 256])
+      >>> po.imshow([img, ac], title=["Input", "Autocorrelation"])
+      <PyrFigure ...>
```
billbrod: in addition to this, I'm wondering if there's a signal / image whose autocorrelation would be easier to parse. some sort of sinusoid?

src/plenoptic/tools/signal.py
```diff
@@ -153,6 +153,30 @@ def rectangular_to_polar(x: Tensor) -> tuple[Tensor, Tensor]:
...
+    for working with steerable pyramid coefficients:
+    >>> import plenoptic as po
+    >>> import torch
+    >>> img = po.data.einstein()  # starting from an image
+    >>> img.shape
+    torch.Size([1, 1, 256, 256])
+    >>> # taking the complex steerable pyramid coefficients
+    >>> spyr = po.simul.SteerablePyramidFreq(img.shape[-2:], is_complex=True)
+    >>> coeff = spyr(img)[0][:, :, 0]  # let's only look at 1 scale and 1 orientation
+    >>> # we can use this function to get the polar coordinates
```
billbrod: given that this example is about `rectangular_to_polar`, I think this comment is unnecessary

src/plenoptic/tools/signal.py
```diff
@@ -477,11 +531,9 @@ def modulate_phase(x: Tensor, phase_factor: float = 2.0) -> Tensor:
...
       >>> img = po.data.einstein()
       >>> spyr = po.simul.SteerablePyramidFreq(img.shape[-2:], is_complex=True)
-      >>> coeffs = spyr(img)
-      >>> mod_coeffs = po.tools.modulate_phase(coeffs[1], 2)
-      >>> mod_coeffs.shape
-      torch.Size([1, 1, 4, 128, 128])
-      >>> po.imshow(mod_coeffs[0])
+      >>> coeff = spyr(img)[3][:, :, 0]  # let's only look at 1 scale and 1 orientation
+      >>> mod_coeff = po.tools.modulate_phase(coeff, 2)
+      >>> po.imshow([coeff, mod_coeff], title=["original", "modulated"])
```
billbrod: Set `zoom=8` (which should get it back to 256x256) here, so we can see what's going on.

and add a brief sentence description (not as a comment, but after the block) pointing out how the phase has visibly changed.

src/plenoptic/tools/signal.py
```diff
@@ -544,6 +595,25 @@ def autocorrelation(x: Tensor) -> Tensor:
       torch.Size([1, 1, 256, 256])
       >>> po.imshow([img, ac], title=["Input", "Autocorrelation"])
       <PyrFigure ...>
+
+    With a simpler periodic example, the autocorrelation is more interpretable:
```
billbrod: I'm not actually sure this is more interpretable, my first impression is that it looks like the same image, just shifted. so maybe scratch this and show that if you compute the autocorrelations on a patch of random noise (`torch.rand`), there's none of the structure that you see in the picture of einstein above, and say something like "random noise has none of the correlation structure that is found in natural images".

And then, I'd like to know if you think this is useful: you'd most likely not use the full autocorrelation here, but the central N shifts, so is it worth showing that you can take the autocorrelation image and pass it to `center_crop` to get just the values that correspond to the smallest shifts? (This is how it gets used by the texture model)

src/plenoptic/tools/signal.py
```diff
@@ -660,6 +730,11 @@ def expand(x: Tensor, factor: float) -> Tensor:
     >>> expanded = po.tools.expand(img, factor=2)
     >>> expanded.shape
     torch.Size([1, 1, 512, 512])
+    >>> po.imshow(
+    ...     [img, expanded],
+    ...     title=["original", "expanded (note the different dimensions!)"],
+    ... )
+    <PyrFigure...>
```
billbrod: this needs to be wrapped in a `.. plot::` block to render correctly

billbrod (changes requested):
Some small changes and then I think this is done!

I'm tempted to add something to expand/shrink to point out that, while they invert each other, it's not exact. Do you think these points would be useful? and if so, where do they belong? They seem to both belong to `shrink`.
- `shrink` throws away info that cannot be recovered so that `expand(shrink(img))` will look different from `img`
- `shrink(expand(img))`, while not as different as `expand(shrink(img))`, is still not too similar: `torch.allclose(img, shrink(expand(img)), atol=1e-2)` will pass, but no smaller values (and playing with `rtol` doesn't help).

billbrod:
I pushed a small change to `CONTRIBUTING.md` to clarify some issues around which blocks get run / how one can reuse objects between blocks in a doctest that Wolf and I discussed as part of this PR

src/plenoptic/tools/signal.py
```diff
@@ -246,7 +247,7 @@ def polar_to_rectangular(amplitude: Tensor, phase: Tensor) -> Tensor:
     torch.Size([1, 1, 256, 256])
     >>> phase.shape
     torch.Size([1, 1, 256, 256])
-    >>> # and from those, we can use this function to get the rectangular coordinates
+    >>> # and from those, we can use this function to recoverthe rectangular coordinates
```
billbrod: "recover the", missing a space

src/plenoptic/tools/signal.py
```diff
@@ -868,7 +878,54 @@ def shrink(x: Tensor, factor: int) -> Tensor:
...
+      >>> torch.allclose(img, expand_after_shrink, atol=1e-2)
+      False
+      >>> po.imshow(
+      ...     [img, expand_after_shrink],
+      ...     title=["original", "expand after shrink"],
+      ... )
+      <PyrFigure...>
+
+    Even in the opposite order, i.e. shrinking an expanded image, we can only recover
+    the original image with a low tolerance:
```
billbrod: I worry that "recover the original image with a low tolerance" is confusing for folks who aren't used to floating point comparisons. maybe the following:

"even in the opposite order, i.e., shrinking an expanded image, the inversion is not perfect. In this example with pixel values between 0 and 1, there are differences on the order of 1e-3."

and maybe show the difference (as in, add `img-shrink_after_expand` to imshow)? since looking at these two images, they look pretty similar

PR #425, opened 2026-04-07
Store penalties over synthesis

billbrod:

Following #411, both @ershook and I spent some time trying to synthesize metamers with different penalties. This process required fine-tuning the value of `penalty_lambda` (see [plenoptic-experiments](https://github.com/plenoptic-org/plenoptic-experiments/tree/main/penalty_visual_diversity)), and as part of that process, I was always manually computing the penalty value on `saved_metamer`.

Realized this meant that I should just be tracking penalty over time, similar to losses. So this PR does that.

This was slightly more complex than expected. I was updating `losses`, `gradient_norm`, etc in `_optimizer_step`, which calls `_closure`, which calls `objective_function`. The penalty was already being computed in `objective_function`, so I initially added the update to `penalties` there. However, this means that `penalties` can get updated outside of synthesis (if the user just calls `objective_function`), which wasn't desired. I could instead re-compute the penalty in `_optimizer_step`, but this is wasteful.

Instead, we:
- add a new private method `_objective_function`, which computes the components used to compute the objective function value (for metamer, this is the metamer loss and penalty; for MAD Competition, the penalty, and each metric)
- This method is called by both `_closure` and `objective_function`, which combine them in the same way. (new tests are added to ensure this)
- `_closure` additionally caches the penalty value in a hidden temporary attribute (`_penalty_tmp`). This is then appended to `_penalties` list in `_optimizer_step`
    - The appending doesn't happen in `_closure` because it can be called multiple times per synthesis iteration (e.g., by LBFGS).
    - Any attributes ending in `_tmp` are not saved to disk
- Because of the two above points, `_closure` is thus moved from the abstract `OptimizedSynthesis` superclass into the concrete subclasses, and all future subclasses will have to implement it.
- We add a hidden attribute `_current_penalty`, analogous to `_current_loss` (see #370)

(During this process, I realized `MADCompetition` was re-computing its metrics during `_optimizer_step` to cache their values, so this behavior was updated to match how we handle `penalties`)

#411 added `penalty_function`

- Update location: I have to update the cached penalty (and metric values, for MADCompetition) in `_closure`, rather than `_optimizer_step`, where loss, gradient norm, and pixel change are all updated, to avoid duplicating calls because `_closure` can only return a single value (otherwise, LBFGS fails, though Adam is fine; see [pytorch forum](https://discuss.pytorch.org/t/multiple-return-variables-in-lbfgs-closure/190381), and [this line](https://github.com/pytorch/pytorch/blob/a2552d55215a0f5bc245ffcdb85024b92ff821ac/torch/optim/lbfgs.py#L315) in the sourcecode). I don't think there's a way around this (unlike with jax, which has a [`has_aux`](https://docs.jax.dev/en/latest/_autosummary/jax.grad.html) option), but I'm annoyed by the inconsistency.
- Loading behavior: when loading in an object that has been saved before this functionality was added, we output a warning. We are unable to recover the penalties over time, in general. Should I add something to the docs explaining how to compute the penalties for `saved_metamer`? I don't want to automatically compute it, because existence of that attribute depends on the argument `store_progress` passed to `synthesize`:
    -  if `store_progress=True`, then we'll have cached the metamer on each iteration of synthesis can update this losslessly
    - if `store_progress=False`, then `saved_metamer` will be empty and we can't update this at all
    - if `store_progress=n`, `n>1`, then we'll have cached the metamer every `n` iterations and thus updating `penalties` in this way would lead to different-from-expected behavior (as `len(penalties)<len(losses)`)
    - UPDATE: I have added a bit to the Compatibility docs showing how to compute the penalties for `saved_metamer`

**Describe changes**

- As described above, adds `_objective_function`, called by both `objective_function` and `_closure`
- As described above, moves `_closure` from `OptimizedSynthesis` to each concrete subclass
- The loading behavior for pre-penalty objects now handles `_penalties` and `_current_penalty` as well.
- Adds `penalty` to progress bar postfix 
- Adds `penalty_function` to API docs
- Updates synthesis design document to describe the change in `_closure`
- Updates User Guide / Metamer to show the penalties attribute
- Removes some `$` from docstrings (they were supposed to convert to LaTeX math, but that's not the right syntax; decided they weren't necessary)
- Adds tests to ensure `_closure` matches `objective_function`
- Adds tests to ensure that `losses` and `penalties` behave the same after initialization, after setup, after synthesize, and after objective_function

**Checklist**


ershook:
_re: Should I add something to the docs explaining how to compute the penalties for saved_metamer?_ I think this would be sufficient. Overall tracking the penalty is going to be really helpful -- thanks for adding this!

src/plenoptic/synthesize/mad_competition.py
```diff
@@ -335,9 +340,51 @@ def synthesize(
...
+        penalty
+            1-element tensor containing the penalty on this step.
+        """
+        if image is None:
+            image = self.image
+        if mad_image is None:
+            mad_image = self.mad_image
+            # if this is empty, then self.mad_image hasn't been initialized
+            if mad_image.numel() == 0:
+                return torch.empty(0)
```
BalzaniEdoardo: this path doesn't match the return (tuple[tensor,tensor,tensor]) of the path of the initialized image

src/plenoptic/synthesize/mad_competition.py
```diff
@@ -335,9 +340,51 @@ def synthesize(
...
             self._current_loss = self.objective_function().item()
+            self._current_penalty = self.penalty_function(self.mad_image).item()
 
         pbar.close()
 
+    def _objective_function(
+        self,
+        mad_image: Tensor | None = None,
+        image: Tensor | None = None,
+    ) -> Tensor:
```
BalzaniEdoardo:
```suggestion
    ) -> Tuple[Tensor, Tensor, Tensor]:
```
I think

src/plenoptic/synthesize/mad_competition.py
```diff
@@ -472,6 +509,39 @@ def get_progress(
             store_progress_attributes=["saved_mad_image"],
         )
 
+    def _closure(self) -> float:
+        r"""
+        Calculate the gradient, before the optimization step.
+
+        This enables optimization algorithms that perform several evaluations
+        of the gradient before taking a step (ie. second order methods like
```
BalzaniEdoardo: or methods with line searches, that includes standard GD and many others

billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/425

src/plenoptic/synthesize/mad_competition.py
BalzaniEdoardo: I leave this one just to show that I did some work, even if you already fixed this XD

src/plenoptic/synthesize/mad_competition.py
BalzaniEdoardo: I am leaving the comments that you already solved

src/plenoptic/synthesize/synthesis.py
```diff
@@ -309,13 +320,17 @@ def penalize_range(img: torch.Tensor) -> torch.Tensor:
...
+                tmp_dict["_penalties"] = [torch.nan] * len(tmp_dict["_losses"])
                 warnings.warn(
                     "The saved object was saved before penalty_function and "
                     "penalty_lambda existed and will not be compatible with future "
                     "releases. Save this object with the current version of plenoptic "
                     "or see the 'Reproducibility and Compatibility' page of the "
                     "documentation for how to make the saved object futureproof and "
-                    "avoid this warning.",
+                    "avoid this warning. However, we cannot recover the history of the "
+                    "penalty_function output (the penalties attribute); filling with"
```
BalzaniEdoardo:
```suggestion
                    "penalty_function output (the penalties attribute); filling with "
```

tests/test_metamers.py
```diff
@@ -1232,3 +1240,58 @@ def test_penalty_effect(self, einstein_img, penalty_function, model):
...
+        shape = (1, 1, 100, 100)
+        img = torch.rand(shape, device=DEVICE)
+        for _ in range(5):
+            met = po.synth.Metamer(img, model)
+            met.setup(torch.rand(shape, device=DEVICE), optimizer=optim)
+            loss = met.objective_function()
+            assert loss == met._closure()
+
+    @pytest.mark.parametrize("seed", range(5))
+    @pytest.mark.parametrize("model", ["PortillaSimoncelli"], indirect=True)
```
BalzaniEdoardo, line 1266:
```suggestion
    @pytest.mark.parametrize("model", ["CTF"], indirect=True)
```

I don't know if this is the exact string,but you see the point

BalzaniEdoardo (commented):
Overall the docstrings are clear, and most of the code fixes I found were already fixed. 

I agree with the code duplication being ok, and the loading warning is clear.

Claude added this comment:
> Design note (not a bug): The penalties abstract method in OptimizedSynthesis has a concrete body that subclasses invoke via super().penalties(synth_attr) inside a @property. This works fine in Python but is an unusual
  pattern — the abstract method has a signature (self, synth_attr) while the overrides are properties. Worth being aware of if someone adds a new subclass.   


Which is more a code pattern and can def be ignored, plus it thinks that LBFGS may be concatenating multiple penalties,

>   Bug: _penalties length mismatch with LBFGS                                                                                                                                                                                   
                                                                                                                                                                                                                            In both Metamer._closure and MADCompetition._closure, self._penalties.append(...) is called inside the closure. With LBFGS, _closure can be called multiple times per optimizer step (that's the whole point of the closure  
  pattern), but self._losses.append(loss) happens once in _optimizer_step. After synthesis, len(_penalties) != len(_losses), which breaks the penalties property (it assumes parity with losses).                                                                                                                                                                                                                                                           The test_closure test parametrizes over LBFGS but only checks objective_function() == _closure() before synthesis, not length consistency after running. This bug would surface during actual synthesis with LBFGS.                                 
  Fix: either accumulate penalties per step (like _losses), or average/take the last penalty value over the multiple closure calls per step, similar to how the optimizer returns a single loss.

tests/test_metamers.py
billbrod: This "parametrize" is actually a way for me to grab the already-initialized `PortillaSimoncelli` model from conftest, and so it needs to have this exact name

PR #431, opened 2026-04-13
Change `PortillaSimoncelli` default `spatial_corr_width` from 9 to 7

### LINKED ISSUE #387: Portilla-Simoncelli `spatial_corr_width` default?
Plenoptic's Portilla-Simoncelli texture model (`PortillaSimoncelli`) defaults to `spatial_corr_width=9`. However, in the [associated paper](https://www.cns.nyu.edu/pub/eero/portilla99-reprint.pdf), they use a value of 7. The [matlab code](https://github.com/LabForComputationalVision/textureSynth) released along with the paper does not have a default value.  I believe our value of 9 came from [`example1.m`](https://github.com/LabForComputationalVision/textureSynth/blob/master/example1.m) in that repo, but do not remember whether there was a specific reason we went with 9 over 7.

We should probably change the default to 7. Then ensure that all the test synthesis still look good.

billbrod:

This  PR changes the default `spatial_corr_width` argument for the `PortillaSimoncelli` model to 7 (from our current value of 9), which matches the value used in the original paper to generate the figures. See #387 for discussion.

Closes #387

**Describe changes**

- Changes default value
- Adds `versionchanged` to docstring informing users of this change
- Updates docs and `test_uploaded_files.py` to use 7 instead of 9
- The only place where I'm specifically still using 9 is the existing `test_models::TestPortillaSimoncelli::test_ps_synthesis` test, which is a regression test of PS metamer synthesis using `MetamerCTF` and `Adam`. The more up-to-date regression tests, in `test_uploaded_files.py` are the ones we use in the docs and have been updated.

**Checklist**

### DIFF
--- a/docs/user_guide/models_and_metrics/portilla_simoncelli/advanced_usage/ps_extensions.md
+++ b/docs/user_guide/models_and_metrics/portilla_simoncelli/advanced_usage/ps_extensions.md
@@ -123,11 +123,11 @@ class PortillaSimoncelliMask(po.simul.PortillaSimoncelli):
         im_shape,
         n_scales=4,
         n_orientations=4,
-        spatial_corr_width=9,
+        spatial_corr_width=7,
         mask=None,
         target=None,
     ):
-        super().__init__(im_shape, n_scales=4, n_orientations=4, spatial_corr_width=9)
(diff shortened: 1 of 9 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/431

PR #433, opened 2026-04-22
Fix apidocs linting

billbrod:

During #418, realized that I had forgotten to add `validate_penalty` to the public api docs, which should be caught by my `linting/check_apidocs.py`. realized that it needs to be set to run every time, without passing it any file names. This required changing how the code works, so it doesn't need to run sphinx to generate the autodoc stub files (which would require plenoptic to be installed). 

instead we do some really basic parsing of the autosummary files in `docs/api` to grab the function names.

Additionally, corrects the formatting for readme warning.

**Checklist**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -81,9 +81,8 @@ repos:
     - id: check-api-docs
       language: system
       name: Ensure all public functions / classes are included in API docs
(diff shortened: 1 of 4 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/433

PR #436, opened 2026-05-17
Jenkins migration to new system

dylex:
Updates needed to build on our [new jenkins system](https://jenkins-new.flatironinstitute.org/job/CCN/job/neurorse/job/plenoptic/) (currently jenkins-new, but will eventually become jenkins).

I've tried to preserve all the configuration, but the build syntax has changed somewhat, mainly due to more flexibility on the new system for resource requests. I've kept the default 4 cores, and the v100 GPUs, but both of these are now configurable. There are A100 (7x MIG) GPUs avaliable, as well as more hardware overall.

Once you merge this in, builds will begin working on the new system, and stop on the old. At that point we can remove the old project to avoid the duplicate reports.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -1,132 +1,94 @@
-pipeline {
-  agent none
-  options {
-    disableConcurrentBuilds(abortPrevious: true)
-    buildDiscarder(logRotator(numToKeepStr: '8', daysToKeepStr: '20'))
-    timeout(time: 3, unit: 'HOURS')
-  }
-  triggers {
-    cron '@weekly'
-  }
-  stages {
-    stage('parallel-jobs') {
-      parallel {
+properties([
+  disableConcurrentBuilds(abortPrevious: true),
+  buildDiscarder(logRotator(numToKeepStr: '8', daysToKeepStr: '20')),
+  pipelineTriggers([cron('@weekly')])
+])
+
+try {
+  timeout(time: 3, unit: 'HOURS') {
+    buildImage(context: 'jenkins')
+    if (env.BRANCH_NAME == 'main' && currentBuild.getBuildCauses('hudson.triggers.TimerTrigger$TimerTriggerCause')) {
+      runPod(gpus: 2) {
         stage('weekly') {
-          when {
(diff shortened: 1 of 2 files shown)


billbrod (approved):
this looks good to me, thanks Dylan! Is it possible to switch to the a100s now? if so, I'll go ahead and do that

jenkins/Jenkinsfile
```diff
@@ -1,132 +1,94 @@
...
+  disableConcurrentBuilds(abortPrevious: true),
+  buildDiscarder(logRotator(numToKeepStr: '8', daysToKeepStr: '20')),
+  pipelineTriggers([cron('@weekly')])
+])
+
+try {
+  timeout(time: 3, unit: 'HOURS') {
+    buildImage(context: 'jenkins')
+    if (env.BRANCH_NAME == 'main' && currentBuild.getBuildCauses('hudson.triggers.TimerTrigger$TimerTriggerCause')) {
+      runPod(gpus: 2, gpuType: 'v100') {
```
billbrod, line 11:
```suggestion
      runPod(gpus: 2, gpuType: 'a100') {
```

jenkins/Jenkinsfile
```diff
@@ -1,132 +1,94 @@
...
-          }
-          post {
-            always {
-              archiveArtifacts artifacts: "uploaded_files/*pt", fingerprint: true
-              dir("$HOME") {
-                deleteDir()
+      }
+    } else {
+      parallel commit: {
+        runPod(gpus: 2, gpuType: 'v100') {
```
billbrod, line 39:
```suggestion
        runPod(gpus: 2, gpuType: 'a100') {
```

dylex:
There was a run earlier that accidentially ran on the a100 and it failed: https://jenkins-new.flatironinstitute.org/job/CCN/job/neurorse/job/plenoptic/job/jenkins/7/
So I'm guessing something else is needed to work there. (Those errors may mean out of memory -- the A100 instances only have 5GB.)

billbrod:
Ah okay, if they only have 5GB then let's stick with V100s, I need more memory than that.

PR #438, opened 2026-05-20
Plot refactor

billbrod:

Following up from #418, this restructures a lot of the synthesis-specific plotting code. In that PR, I realized that there was a lot of redundancy (e.g., `po.plot.mad_pixel_values` and `po.plot.metamer_pixel_values`) were essentially the same function, and that I could make the plotting code either to use by consolidating it. To that end, I have combined many of the synthesis-specific plotting function into single `synthesis_{plot}` functions (e.g., `synthesis_loss`) which accept multiple synthesis objects. If the plot cannot accept a given synthesis object (e.g., `synthesis_loss` cannot accept an `Eigendistortion` object, because eigendistortions don't have optimization loss), then they raise an informative error message to that effect (and the docstrings/annotations hopefully clarify this). For plotting functions that only operate on a single synthesis object (e.g., `metamer_representation_error`), they have been kept separate and retain the naming scheme from #418.

Consolidated functions:

- `synthesis_loss`: `metamer_loss`, `mad_loss`
- `synthesis_histogram`: `metamer_pixel_values`, `mad_pixel_values`. Adds capability for `Eigendistortion` and renamed because a histogram is potentially valuable even if the values aren't pixels.
- `synthesis_imshow`: `metamer_imshow`, `mad_imshow`, `eigendistortion_imshow`.
- `synthesis_status`: `metamer_synthesis_status`, `mad_synthesis_status`. Adds capability for `Eigendistortion`.
- ~`synthesis_animshow`~ `synthesis_animate`: `metamer_animshow`, `mad_animshow`.

Adds:

- `histogram`: tensor-level histogram plot used by `synthesis_histogram` (analogous to how we have an `imshow` that operates directly on image tensors used by `synthesis_imshow`)

Remaining functions:

- `metamer_representation_error`
- `eigendistortion_imshow_all`
- `mad_imshow_all`
- `mad_loss_all`

For giving feedback on how users will interact with these functions, I think the most informative way to review will be to look at the [plotting section of the API guide](https://docs.plenoptic.org/docs/pulls/438/api/plot.html), and look through the docstrings of each of the functions there.

Follow up from #418, closes #417.

- We generally use `batch_idx` to refer to the first index in the tensors, as image tensors in pytorch typically are 4d: `batch, channel, height, width`. For eigendistortions, this has a special meaning: the eigenindex (e.g., eigenindex=0 means the distortion corresponds to the first eigenvector). In the shared functions, I've used `batch_idx` as the argument name, describing the relationship in the docstring, whereas I've kept it as `eigenindex` for `eigendistortion_imshow_all` (also explaining in the docstring). Is this better than being consistent?
- `synthesis_status` and `synthesis_animshow` create single figures by combining the outputs of the axis-level functions (`synthesis_loss`, `synthesis_imshow`, etc). I have made it so that users can pass arguments to these functions directly using kwargs (hopefully docstrings are clear on this), so they don't clutter up the function signature. Is that clear right now?
    - The one weird thing here is `synthesis_animshow`'s `ylim` arg, which controls how we rescale the y-limits of the `metamer_representation_error` plot. It is only related to that function, but it only makes sense when animating its output. Therefore, I've kept it as an argument for the animshow function. Is this okay? is there a better way to handle it?
    - EDIT: changed how I handle this arg, renaming it `rescale_ylim` after discussing with Sarah, see [comment](https://github.com/plenoptic-org/plenoptic/pull/438#issuecomment-4558593182)
- What is now `synthesis_animshow` was `animate` before #418, to try and make it more consistent with the tensor-level functions. However, now I'm not sure of that name -- `synthesis_animshow` is the animated analogue of `synthesis_status`, animating that figure over synthesis iteration. But this naming makes it sound like it's the animated analogue of `synthesis_imshow` (as `po.plot.animshow` is the animated analogue of `po.plot.imshow`). So that makes me inclined to change the name -- `synthesis_status_animate` or somethign seems too long to me. I could go back to `synthesis_animate`. Any other ideas?
    - EDIT: changed back to `synthesis_animate`
- I have put `versionadded` tags in the docstring of each of the consolidated functions above, but the way they are interacted with has changed somewhat. Should I try to explain that more explicitly? Or just rely on people reading the docstrings? (The section in the "Migration Guide" also basically just says "check their docstrings")

**Describe changes**

- Consolidated synthesis plotting functions into a small number which work on multiple synthesis objects.
- Added new tensor-level `po.plot.histogram` function, used by the `po.plot.synthesis_histogram` function.
- Added "New function" table to migration guide, which just consists of `po.plot.histogram`. Updates section on plotting functions.
- Added new example synthesis objects to `test_uploaded_files` / fetch capability: `example_mad.pt`, `example_mad-cuda.pt`, `example_eigendistortion_color.pt`
- `po.plot.imshow` and `po.plot.animshow` raise some more informative error messages around dimensionality/shape.
- Moved eigendistortion plotting tests from `test_eigendistortion` into `test_display`.
- Added variety of tests for new functionality.
- Renamed `Synthesis`, `OptimizedSynthesis` to `_Synthesis`, `_OptimizedSynthesis` since they're private.
- Updated references to consolidated functions throughout.
- Restructured the plotting section of the API docs (formally, "Display", now "Plotting functions").
- in CI, set doctests to continue on failure for more informative errors.

**Checklist**


billbrod:
Documentation built by Jenkins: https://docs.plenoptic.org/docs/pulls/438/index.html

src/plenoptic/plot/synthesis.py
```diff
@@ -0,0 +1,2104 @@
...
+    See Also
+    --------
+    :func:`~plenoptic.plot.imshow`
+        Function used by this one to visualize the metamer image.
+    synthesis_status
+        Create a figure combining this with other axis-level plots to summarize
+        synthesis status at a given iteration.
+    synthesis_animshow
+        Create a video animating this and other axis-level plots changing over
+        the course of synthesis.
```
sjvenditto, line 783: Is it worth adding links to the two `imshow_all` functions as well?

src/plenoptic/plot/mad_competition.py
```diff
@@ -886,6 +73,19 @@ def mad_imshow_all(
...
+    --------
+    :func:`~plenoptic.plot.synthesis_imshow`
+        Display the image from a single :class:`~plenoptic.MADCompetition` instance.
+    :func:`~plenoptic.plot.synthesis_status`
+        Create a composite plot showing synthesis status of a single
+        :class:`~plenoptic.MADCompetition` instance.
+
+    Examples
+    --------
+    See the `MAD Competition <mad-nb>`_ `tutorial notebooks <mad-concept>`_ in the User
```
sjvenditto: This link isn't working

src/plenoptic/plot/synthesis.py
```diff
@@ -0,0 +1,2104 @@
...
+    Make a plot showing synthesis status.
+
+    .. versionadded:: 2.0
+       Combines previously separate loss plotting functions for
+       :class:`~plenoptic.Metamer`, :class:`~plenoptic.MADCompetition`, and adds support
+       for :class:`~plenoptic.Eigendistortion`.
+
+    We create several subplots to analyze this. The plots to include are
+    specified by including their name in the ``included_plots`` list. All plots
+    can be created separately using the method with the same name.
```
sjvenditto:
```suggestion
    can be created separately using this method with the individual plot name.
```

I got a little hung up on the phrasing of this

sjvenditto (commented):
Overall I think it's looking good! One thing I noticed is that there are a number of raised errors that aren't being tested. The ones I think are more important to cover are:
- Tests that when combined functions are called on an unsupported synthesis object (e.g. eigendistorion with `synthesis_animshow`), an error is raised
- The default included plots aren't being tested for eigendistorion. I believe the test `test_allowed_plots_exception` is what is testing it for metamers and MAD competition?

I've also added a few minor comments at specific lines

For your other questions:
> We generally use batch_idx to refer to the first index in the tensors, as image tensors in pytorch typically are 4d: batch, channel, height, width. For eigendistortions, this has a special meaning: the eigenindex (e.g., eigenindex=0 means the distortion corresponds to the first eigenvector). In the shared functions, I've used batch_idx as the argument name, describing the relationship in the docstring, whereas I've kept it as eigenindex for eigendistortion_imshow_all (also explaining in the docstring). Is this better than being consistent?

I think this is fine, the first positional variable name changes as well (which makes sense since it's specific to eigendistortions)

> synthesis_status and synthesis_animshow create single figures by combining the outputs of the axis-level functions (synthesis_loss, synthesis_imshow, etc). I have made it so that users can pass arguments to these functions directly using kwargs (hopefully docstrings are clear on this), so they don't clutter up the function signature. Is that clear right now?

I think the explanation on how to name the kwargs is clear; however, one concern is that the type shouldn't really be `Any`, but instead should be clear that each individual `_kwargs` is a `dict`. 

> The one weird thing here is synthesis_animshow's ylim arg, which controls how we rescale the y-limits of the metamer_representation_error plot. It is only related to that function, but it only makes sense when animating its output. Therefore, I've kept it as an argument for the animshow function. Is this okay? is there a better way to handle it?

I think the way you handle this comes down to how likely the user is to change the value. It also seems like you're introducing some new rescaling logic of the ylim that the individual `Metamer_representation_error` doesn't have. It also looks like if the user sets `kwargs["metamer_representation_error_kwargs"]["ylim"]` directly, it will get overwritten. My thoughts are:
- If it is something that you think will be interacted with a reasonable amount over other plot kwargs, leave it outside of the plot kwargs like you have done already
- The parameter name should be changed to make it more obvious that it is only for the one Metamer plot, something like `metamer_ylim`. You might even move it to be the last keyword in the signature before `**kwargs` to further separate it from parameters that are general to all subplot types
- For the most consistency with how other plot kwargs are set while also introducing the rescale logic you could:
     - make the variable something like `rescale_metamer_ylim` to be `None` or a string. It will only overwrite `kwargs["metamer_representation_error_kwargs"]["ylim"]` if it is not `None`
     - if the user want's to pass `ylim` directly to the `metamer_representation_error`, they assign it themselves in the `metamer_representation_error_kwargs` dictionary. 

The issue with my suggested approach is when `"rescale"` is the default, it will always overwrite any `ylim` setting, so you might also need to check if `ylim` is set in `metamer_representation_error_kwargs` before overwriting

> What is now synthesis_animshow was animate before https://github.com/plenoptic-org/plenoptic/pull/418, to try and make it more consistent with the tensor-level functions. However, now I'm not sure of that name -- synthesis_animshow is the animated analogue of synthesis_status, animating that figure over synthesis iteration. But this naming makes it sound like it's the animated analogue of synthesis_imshow (as po.plot.animshow is the animated analogue of po.plot.imshow). So that makes me inclined to change the name -- synthesis_status_animate or somethign seems too long to me. I could go back to synthesis_animate. Any other ideas?

I would go back to `synthesis_animate`. I agree that it will help distinguish that it not just an animated `imshow`

> I have put versionadded tags in the docstring of each of the consolidated functions above, but the way they are interacted with has changed somewhat. Should I try to explain that more explicitly? Or just rely on people reading the docstrings? (The section in the "Migration Guide" also basically just says "check their docstrings")

I think you can just rely on them reading the docstring

src/plenoptic/plot/synthesis.py
billbrod: yep

billbrod:
> Overall I think it's looking good! One thing I noticed is that there are a number of raised errors that aren't being tested. The ones I think are more important to cover are:
> 
>     * Tests that when combined functions are called on an unsupported synthesis object (e.g. eigendistorion with `synthesis_animshow`), an error is raised
> 
>     * The default included plots aren't being tested for eigendistorion. I believe the test `test_allowed_plots_exception` is what is testing it for metamers and MAD competition?
 
Agreed, done! I ended up going through and I think I brought the code coverage up as high as it can go -- the remaining missing lines are `def __dir__` and also the checks in `_get_synthesis_image` that also happen in `_get_synthesis_title`, which is currently always called first (I'm leaving the checks there in case there's a function that added that doesn't need `_get_synthesis_title`).
 
> I've also added a few minor comments at specific lines

done.
 
> For your other questions:
> 
> > We generally use batch_idx to refer to the first index in the tensors, as image tensors in pytorch typically are 4d: batch, channel, height, width. For eigendistortions, this has a special meaning: the eigenindex (e.g., eigenindex=0 means the distortion corresponds to the first eigenvector). In the shared functions, I've used batch_idx as the argument name, describing the relationship in the docstring, whereas I've kept it as eigenindex for eigendistortion_imshow_all (also explaining in the docstring). Is this better than being consistent?
> 
> I think this is fine, the first positional variable name changes as well (which makes sense since it's specific to eigendistortions)
> 
> > synthesis_status and synthesis_animshow create single figures by combining the outputs of the axis-level functions (synthesis_loss, synthesis_imshow, etc). I have made it so that users can pass arguments to these functions directly using kwargs (hopefully docstrings are clear on this), so they don't clutter up the function signature. Is that clear right now?
> 
> I think the explanation on how to name the kwargs is clear; however, one concern is that the type shouldn't really be `Any`, but instead should be clear that each individual `_kwargs` is a `dict`.

I like that. I also added an informative error message if there's anything extra in kwargs, since nothing happens with it, so it's probably a typo.
 
> > The one weird thing here is synthesis_animshow's ylim arg, which controls how we rescale the y-limits of the metamer_representation_error plot. It is only related to that function, but it only makes sense when animating its output. Therefore, I've kept it as an argument for the animshow function. Is this okay? is there a better way to handle it?
> 
> I think the way you handle this comes down to how likely the user is to change the value. It also seems like you're introducing some new rescaling logic of the ylim that the individual `Metamer_representation_error` doesn't have. It also looks like if the user sets `kwargs["metamer_representation_error_kwargs"]["ylim"]` directly, it will get overwritten. My thoughts are:
> 
>     * If it is something that you think will be interacted with a reasonable amount over other plot kwargs, leave it outside of the plot kwargs like you have done already
> 
>     * The parameter name should be changed to make it more obvious that it is only for the one Metamer plot, something like `metamer_ylim`. You might even move it to be the last keyword in the signature before `**kwargs` to further separate it from parameters that are general to all subplot types
> 
>     * For the most consistency with how other plot kwargs are set while also introducing the rescale logic you could:
>       
>       * make the variable something like `rescale_metamer_ylim` to be `None` or a string. It will only overwrite `kwargs["metamer_representation_error_kwargs"]["ylim"]` if it is not `None`
>       * if the user want's to pass `ylim` directly to the `metamer_representation_error`, they assign it themselves in the `metamer_representation_error_kwargs` dictionary.
>
> The issue with my suggested approach is when `"rescale"` is the default, it will always overwrite any `ylim` setting, so you might also need to check if `ylim` is set in `metamer_representation_error_kwargs` before overwriting

as we talked about, I:
- renamed this to `rescale_ylim` and have it only govern the rescaling behavior. Right now it only affects the `metamer_representation_error` plot (noted in the docstring), but could affect others in the future.
- It can accept "rescale", "rescaleN", or `False` (don't rescale)
- the other functionality (setting ylim at initialization) should thus be handled by passing ylim to kwargs. I don't show a specific example of this -- should I?
- moved it to the end, before kwargs. I'm unsure about that now -- what do you think? the question is whether it belongs before or after the figure-related args
  
> > What is now synthesis_animshow was animate before #418, to try and make it more consistent with the tensor-level functions. However, now I'm not sure of that name -- synthesis_animshow is the animated analogue of synthesis_status, animating that figure over synthesis iteration. But this naming makes it sound like it's the animated analogue of synthesis_imshow (as po.plot.animshow is the animated analogue of po.plot.imshow). So that makes me inclined to change the name -- synthesis_status_animate or somethign seems too long to me. I could go back to synthesis_animate. Any other ideas?
> 
> I would go back to `synthesis_animate`. I agree that it will help distinguish that it not just an animated `imshow`

agreed, done.
 
> > I have put versionadded tags in the docstring of each of the consolidated functions above, but the way they are interacted with has changed somewhat. Should I try to explain that more explicitly? Or just rely on people reading the docstrings? (The section in the "Migration Guide" also basically just says "check their docstrings")
> 
> I think you can just rely on them reading the docstring

sjvenditto (approved):
I like the new way you're handling the rescale! For your additional questions:
> the other functionality (setting ylim at initialization) should thus be handled by passing ylim to kwargs. I don't show a specific example of this -- should I?

I don't think you need an extra example, it falls within the umbrella of setting plot specific kwargs.

> moved it to the end, before kwargs. I'm unsure about that now -- what do you think? the question is whether it belongs before or after the figure-related args

I'm indifferent on this, I think it's fine to move it back up now that it's clearer (at least to me) that it is something specific to `synthesis_animate`.

billbrod:
> I like the new way you're handling the rescale! For your additional questions:
> 
> > the other functionality (setting ylim at initialization) should thus be handled by passing ylim to kwargs. I don't show a specific example of this -- should I?
> 
> I don't think you need an extra example, it falls within the umbrella of setting plot specific kwargs.
> 
> > moved it to the end, before kwargs. I'm unsure about that now -- what do you think? the question is whether it belongs before or after the figure-related args
> 
> I'm indifferent on this, I think it's fine to move it back up now that it's clearer (at least to me) that it is something specific to `synthesis_animate`.

Thanks! I'll wait to see whether @ershook has any opinions on this, and then this is good to go!

ershook:
#

> We generally use batch_idx to refer to the first index in the tensors, as image tensors in pytorch typically are 4d: batch, channel, height, width. For eigendistortions, this has a special meaning: the eigenindex (e.g., eigenindex=0 means the distortion corresponds to the first eigenvector). In the shared functions, I've used batch_idx as the argument name, describing the relationship in the docstring, whereas I've kept it as eigenindex for eigendistortion_imshow_all (also explaining in the docstring). Is this better than being consistent?

I would keep it Eigenindex and not worry about being consistent.

> synthesis_status and synthesis_animshow create single figures by combining the outputs of the axis-level functions (synthesis_loss, synthesis_imshow, etc). I have made it so that users can pass arguments to these functions directly using kwargs (hopefully docstrings are clear on this), so they don't clutter up the function signature. Is that clear right now?

This all makes sense to me. 

> Any other ideas?
EDIT: changed back to synthesis_animate

This also makes sense to me.

> I have put versionadded tags in the docstring of each of the consolidated functions above, but the way they are interacted with has changed somewhat. Should I try to explain that more explicitly? Or just rely on people reading the docstrings? (The section in the "Migration Guide" also basically just says "check their docstrings")

I agree with @sjvenditto you can just rely on people reading the doc strings.

Overall, think this refactor makes the plotting functions much more user friendly and less clunky.

billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/438

PR #444, opened 2026-06-01
Adds `raise_on_checks` to synthesis load methods

### LINKED ISSUE #384: Add `raise_on_checks` to Synthesis load methods
Add `skip_checks` argument to the synthesis object load methods, which skips all the checks that we use to ensure that the saved and loading objects are initialized in the same way. Intended use case is to make it easier to load an object that was saved with an older version of plenoptic when that's the *only* change. My worry is that it would be easy to get into a mixed up state (with subtle differences in e.g., the loss function). I guess instead of skipping the checks, I could raise a warning if they fail rather than an error? So then, instead of `skip_checks`, make it `raise_on_checks`, defaulting to True.

Not sure the best way to test this -- should I just take all the load tests that currently fail, and instead check that they warn appropriately instead of fail? should probably also check that the state after loading is as expected (all attributes set at initialization come from loading object, all others come from saved one).

Assuming not too much time has passed, the following file (which shows the output of `git diff`) may help (from #381). It can be applied with `patch -p1 < skip_checks.txt` (using [patch](https://en.wikipedia.org/wiki/Patch_%28Unix%29))

[skip_checks.txt](https://github.com/user-attachments/files/24088110/skip_checks.txt)

So it's stated somewhere: my goal with the behavior of save/load (in #317 )  is to ensure that users are actually loading the object they think they are. That is, to avoid initializing `Metamer` with one image/model/loss_function/optimizer and loading in an object run with a different one, overriding the initialization (while also avoiding saving live code objects, which is both difficult and unsafe).

billbrod:

This PR adds `raise_on_checks` to the synthesis objects load methods. During load, we perform a variety of checks to ensure that the saved object was initialized in the same way as the loading one (since we don't save live code objects). Normally, these result in errors if they fail. setting `raise_on_checks=False` turns them into `plenoptic.io.LoadWarning` warnings instead (a new custom warning class). The intended use case here is if the user e.g., changes the name of an object or I do something annoying across versions. 

This should be used sparingly and all warnings should be manually expected.

Closes #384

- What do you think of this arg? It's a little of "you're on your own" if you use it.
- I made a custom warning largely so it's easier to just grab it -- what do you think? (I've never done that before.)

**Describe changes**

- Adds `raise_on_checks`, `LoadWarning` as above
- Warning message should clarify whether we use the saved or initialized object (it's the saved one for all non-callables, initialized for callables)
- There are exceptions: if the dtype or device is different, that will always result in an error, because there are better ways to change that (calling `met.to(dtype)` or `met.load("file.pt", map_location=device)`)
- Adds section to reproducibility docs page describing this arg and the rationale behind why load behaves the way it does
- Adds tests
- Small change to `nlpd` allowing users to specify the `epsilon` used to stabilize the computation near 0. this is to make sure `nlpd` can be used as a metric for `MADCompetition` (the default epsilon was too high).
- Was getting a strange `ImportMismatchError` when running the doctests for the `src/plenoptic/conftest.py` file, unsure why. Running them with `--import-mode=importlib` fixes the problem, and is the [recommended import-mode](https://docs.pytest.org/en/stable/explanation/goodpractices.html#which-import-mode) for new projects from pytest. However, it doesn't work with our regular tests, since [it prevents test modules from importing each other](https://docs.pytest.org/en/stable/explanation/pythonpath.html#import-modes). This would require a decently sized change to our tests with no real gain, so we're just adding that flag to the doctest run.

**Checklist**


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/444

sjvenditto (commented):
I like the idea of having this feature, especially since it helps with backward compatibility of loading in old models with old package paths. I also like the custom warning since it makes it more descriptive of what is going on vs a standard warning.

One issue I see is that it allows you to quietly overwrite critical functions. E.g. modifying the test `test_raise_on_checks_new_attr` in `test_eigendistortion.py`, the following load raises no errors at all:
```
class Eigendistortion(po.Eigendistortion):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.synthesize = True

eig = Eigendistortion(einstein_img[..., :SMALL_DIM, :SMALL_DIM], model)
eig.load(gaussian_einstein_img_eig_saved, raise_on_checks=False)
```
(the default load will raise an error, of course, because of the name change). I'm not fully sure on how to handle this, since you can't really test to check that the old and new definitions of `synthesis` match, because you won't have access to the old version. Maybe the best approach, since this is already a "use at your own risk" flag, is to add an additional warning in the documentation that there's no way to check for method mismatch if you use any custom modifications to synthesis classes?

billbrod:
That's a good catch. I don't think there's anything I can do about this, and presumably the most likely cases where this would happen is if I change the object (in which case I'll try to preserve backwards compatibility) or the user does (in which case they know what they did). So hopefully it's not too much of a problem. But I'll add a note to the docs, docstrings, and warning.

PR #445, opened 2026-06-02
First basic example of batch of images using imshow

emeyer121:

> This PR adds a basic doctest example in imshow to display a batch of grayscale images.

> - Test case for adding examples to docstrings, will continue adding more.

**Describe changes**

In a list, describe the changes you made. This should be more granular than the description at the beginning of the PR. These descriptions can be short, but this would also be the place to describe why something seemingly unrelated was included. e.g.,

> - added plot of 5 images with imshow
> - duplicate of example from load_images for testing
> - will add more examples within imshow and other display functions

**Checklist**

### DIFF
--- a/src/plenoptic/plot/display.py
+++ b/src/plenoptic/plot/display.py
@@ -234,6 +234,19 @@ def imshow(
     generally hold in notebooks as well, but will fail if, e.g., you plot an image
     that's 2000 pixels wide on an monitor 1000 pixels wide; the browser handles the
     rescaling in a way we can't control.
+
+    Examples
+    --------
+
+    .. plot::
+      :context: reset
+
+      >>> import plenoptic as po
+      >>> img_dir = po.data.fetch_data("test_images.tar.gz") / "256"
+      >>> titles = ["color_wheel", "curie", "einstein", "metal", "nuts"]
+      >>> imgs = po.load_images(img_dir)
+      >>> po.plot.imshow(imgs, title=titles)
+      <PyrFigure size ... with 5 Axes>
     """
     if not isinstance(image, list):
         image = [image]
--- a/src/plenoptic/plot/display.py
+++ b/src/plenoptic/plot/display.py
@@ -249,14 +249,9 @@ def imshow(
       >>> po.plot.imshow(einstein)
       <PyrFigure size ... with 1 Axes>
 
-    For a batch of images, imshow will plot every image:
-
-    .. plot::
-      :context: close-figs
-
-      >>> import plenoptic as po
-      >>> po.plot.imshow([po.data.einstein(), po.data.curie()])
-      <PyrFigure size ... with 2 Axes>
+    For an image tensor with multiple elements along the batch dimension and a
+    single channel element, this function will plot each batch independently as
+    grayscale images:
 
     .. plot::
       :context: close-figs
@@ -269,7 +264,19 @@ def imshow(
       >>> po.plot.imshow(imgs)
       <PyrFigure size ... with 2 Axes>
 
-    Specifying batch_idx will limit the output to a single image:
+    A list of 4d tensors will be concatenated along the batch dimension before plotting.
+    Thus, the following example is the same as above:
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> po.plot.imshow([po.data.einstein(), po.data.curie()])
+      <PyrFigure size ... with 2 Axes>
+
+    You may use the ``title`` argument for any number of images, either as a string
+    applied to all images or as a list the length of images. Additionally, ``col_wrap``
+    specifies the number of images per row.
 
     .. plot::
       :context: close-figs
@@ -279,21 +286,26 @@ def imshow(
       >>> imgs = torch.cat([po.data.einstein(), po.data.curie()])
       >>> print(imgs.shape)
       torch.Size([2, 1, 256, 256])
-      >>> batch_idx = 1
-      >>> po.plot.imshow(imgs, batch_idx=batch_idx)
-      <PyrFigure size ... with 1 Axes>
+      >>> po.plot.imshow(imgs, title=["einstein", "curie"], col_wrap=1)
+      <PyrFigure size ... with 2 Axes>
 
-    The vrange flag will normalize image values independently or across all images.
-    Note the image ranges for each method.
+    Specifying ``batch_idx`` will plot the corresponding element in
+    the batch dimension (i.e., ``imgs[batch_idx]``)
 
     .. plot::
       :context: close-figs
 
       >>> import plenoptic as po
       >>> import torch
-      >>> imgs = torch.cat([po.data.einstein(), po.data.einstein() * 2])
-      >>> po.plot.imshow(imgs, vrange="indep1")
-      <PyrFigure size ... with 2 Axes>
+      >>> imgs = torch.cat([po.data.einstein(), po.data.curie()])
+      >>> print(imgs.shape)
+      torch.Size([2, 1, 256, 256])
+      >>> po.plot.imshow(imgs, batch_idx=1)
+      <PyrFigure size ... with 1 Axes>
+
+    The vrange argument allows control over the min and max values of the color range.
+    In addition to a 2-tuple of labels, plenoptic includes several special strings:
+    ``"auto1"`` sets all images to have the same range.
 
     .. plot::
       :context: close-figs
@@ -304,14 +316,20 @@ def imshow(
       >>> po.plot.imshow(imgs, vrange="auto1")
       <PyrFigure size ... with 2 Axes>
 
-    Using zoom will up- or downscale the number of display pixels:
+    Meanwhile, ``"indep1"`` sets each image's range independently. Note the different
+    ranges in the titles!
 
     .. plot::
       :context: close-figs
 
       >>> import plenoptic as po
-      >>> po.plot.imshow(po.data.einstein())
-      <PyrFigure size ... with 1 Axes>
+      >>> import torch
+      >>> imgs = torch.cat([po.data.einstein(), po.data.einstein() * 2])
+      >>> po.plot.imshow(imgs, vrange="indep1")
+      <PyrFigure size ... with 2 Axes>
+
+    The ``zoom`` argument allows users to set the ratio of display to image pixels,
+    increasing or decreasing the size of the resulting plot:
 
     .. plot::
       :context: close-figs
@@ -320,7 +338,18 @@ def imshow(
       >>> po.plot.imshow(po.data.einstein(), zoom=0.5)
       <PyrFigure size ... with 1 Axes>
 
-    You can use the plot_complex flag for complex values:
+    If ``zoom<1`` and the value is not a divisor of the largest image size, it will
+    result in an error:
+
+    >>> import plenoptic as po
+    >>> print(po.data.einstein().shape)
+    torch.Size([1, 1, 256, 256])
+    >>> po.plot.imshow(po.data.einstein(), zoom=0.7)
+    Traceback (most recent call last):
+    Exception: zoom * signal.shape must result in integers!
+
+    You can use the ``plot_complex`` argument to control how complex tensors are
+    plotted:
 
     .. plot::
       :context: close-figs
@@ -329,28 +358,46 @@ def imshow(
       >>> import torch
       >>> img = po.data.einstein()
       >>> fft_img = torch.fft.fft2(img)
-      >>> po.plot.imshow([img, fft_img], plot_complex="polar")
+      >>> po.plot.imshow([img, fft_img], plot_complex="logpolar")
       <PyrFigure size ... with 3 Axes>
 
-    Value errors can occur if the as_rgb flag and image dimensions are misaligned:
+    To plot an RGB image, you must set ``as_rgb=True`` to plot a color image:
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> color_wheel = po.data.color_wheel()
+      >>> print(color_wheel.shape)
+      torch.Size([1, 3, 600, 600])
+      >>> po.plot.imshow(color_wheel, as_rgb=True)
+      <PyrFigure size ... with 1 Axes>
+
+    Otherwise, images with multiple channels will have each channel plotted as a
+    separate grayscale image:
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> color_wheel = po.data.color_wheel()
+      >>> print(color_wheel.shape)
+      torch.Size([1, 3, 600, 600])
+      >>> po.plot.imshow(color_wheel)
+      <PyrFigure size ... with 3 Axes>
+
+    Value errors can occur if ``as_rgb=True`` and the input image doesn't have the
+    required number of channels:
 
     >>> import plenoptic as po
     >>> einstein = po.data.einstein()
-    >>> einstein.shape
+    >>> print(einstein.shape)
     torch.Size([1, 1, 256, 256])
     >>> po.plot.imshow(einstein, as_rgb=True)
     Traceback (most recent call last):
     ValueError: If as_rgb is True, then channel must have 3 or 4 elements!
 
-    >>> import plenoptic as po
-    >>> img = torch.cat([po.data.color_wheel(), po.data.color_wheel()])
-    >>> print(img.shape)
-    torch.Size([2, 3, 600, 600])
-    >>> po.plot.imshow(img)
-    Traceback (most recent call last):
-    ValueError: Don't know how to plot non-rgb images with more than one channel and ...
-
-    Images will automatically rescaled to be displayed at the same height and widths
+    Images will be automatically rescaled to be displayed at the same heights and widths
     if sizes are scalar multiples of each other:
 
     .. plot::
--- a/src/plenoptic/plot/display.py
+++ b/src/plenoptic/plot/display.py
@@ -259,7 +259,7 @@ def imshow(
       >>> titles = ["color_wheel", "curie", "einstein", "metal", "nuts"]
       >>> imgs = po.load_images(img_dir)
       >>> print(imgs.shape)
-      torch.Size([5, 3, 256, 256])
+      torch.Size([5, 1, 256, 256])
       >>> po.plot.imshow(imgs, title=titles)
       <PyrFigure size ... with 5 Axes>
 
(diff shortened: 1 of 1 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/445

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,19 @@ def imshow(
...
     rescaling in a way we can't control.
+
+    Examples
+    --------
+
+    .. plot::
+      :context: reset
+
+      >>> import plenoptic as po
+      >>> img_dir = po.data.fetch_data("test_images.tar.gz") / "256"
```
billbrod: I worry that this example is pretty opaque if people don't already understand what `fetch_data` and `load_data` do. For example, they won't know what type of object (or shape) `imgs` is, which is pretty important.

instead, can the first example be super basic: just plot a single image (one of the ones that exists under `po.data`) and, before plotting, print out the shape of it. then have some text before the example that says something like "plot a single grayscale image", so it's clear what the example is showing.

billbrod (changes requested):
Thanks for this Emily! I left some comments inline

emeyer121:
Changed the first example to make it very basic. Added a few additional examples that show plotting multiple images, using batch_idx to plot single image from a batch, normalizing images with vrange, and a ValueError when trying to plot a grayscale image with as_rgb=True. I am not sure if the vrange example is helpful since they aren't very obviously different. I also considered using a value error example where there is a batch of RGB images and not specifying a batch or channel, but the error message is too long for ruff and wasn't sure how to deal with that yet. Happy to add/remove/edit any of these example options

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+      :context: reset
+
+      >>> import plenoptic as po
+      >>> einstein = po.data.einstein()
+      >>> einstein.shape
+      torch.Size([1, 1, 256, 256])
+      >>> po.plot.imshow(einstein)
+      <PyrFigure size ... with 1 Axes>
+
+    For a batch of images, imshow will plot every image:
```
billbrod:
```suggestion
    For an image tensor with multiple elements along the batch dimension and a single channel element, this function will plot each batch independently as grayscale images:
```

and then follow this with the example using `torch.cat`.

I typically use "this function" when referring to itself, as a way of protecting myself against possibly changing the function name in the future (this doesn't happen much anymore, but better safe than sorry)

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
```
billbrod: add some text before this example saying something like: "A list of 4d tensors will be concatenated along the batch dimension before plotting. Thus, the following example is the same as above"

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+      >>> import torch
+      >>> imgs = torch.cat([po.data.einstein(), po.data.curie()])
+      >>> print(imgs.shape)
+      torch.Size([2, 1, 256, 256])
+      >>> po.plot.imshow(imgs)
+      <PyrFigure size ... with 2 Axes>
+
+    Specifying batch_idx will limit the output to a single image:
```
billbrod:
```suggestion
    Specifying ``batch_idx`` will plot the corresponding element in the batch dimension (i.e., `imgs[batch_idx]`)
```

use double backticks when referring to argument names (this will render as `batch_idx` in the docs, it's double backticks because our docstrings are [reStructuredText / rST](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html), sphinx's preferred format, rather than markdown, which is used by github (and many other places, including the contents of our `docs` folder))

the change I made to the end of the sentence is because, if your input has multiple batches and multipel channels, you will not get a single image, e.g.,:

```python
imgs = torch.cat([po.data.einstein(False), po.data.curie(False)])
po.plot.imshow(imgs, batch_idx=0)
```

will plot each of the channels of the einstein image separately

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> import torch
+      >>> imgs = torch.cat([po.data.einstein(), po.data.curie()])
+      >>> print(imgs.shape)
+      torch.Size([2, 1, 256, 256])
+      >>> batch_idx = 1
+      >>> po.plot.imshow(imgs, batch_idx=batch_idx)
```
billbrod:
```suggestion
      >>> po.plot.imshow(imgs, batch_idx=1)
```

no reason to be verbose

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+      <PyrFigure size ... with 1 Axes>
+
+    The vrange flag will normalize image values independently or across all images.
+    Note the image ranges for each method.
```
billbrod:
```suggestion
    The vrange argument allows control over the min and max values of the color range.
    In addition to a 2-tuple of labels, plenoptic includes several special strings: ``"indep1"`` 
    sets each range independently.
```

In python, functions inputs are "arguments". (flags typically refer inputs like `--a-flag` that are used on the command line)

normalizing vrange independently/across images are the most common, but not the only behavior: one could pass a single tuple to specify the range across all images, in which case there's no normalizing, and we're probably going to soon add the ability to pass a list of tuples to set each independently.

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> import torch
+      >>> imgs = torch.cat([po.data.einstein(), po.data.einstein() * 2])
+      >>> po.plot.imshow(imgs, vrange="indep1")
+      <PyrFigure size ... with 2 Axes>
+
+    .. plot::
```
billbrod, line 312: add text description saying "While ``"auto1"`` sets all images to have the same range. Note the different range in the title!"

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+      >>> import plenoptic as po
+      >>> import torch
+      >>> imgs = torch.cat([po.data.einstein(), po.data.einstein() * 2])
+      >>> po.plot.imshow(imgs, vrange="auto1")
+      <PyrFigure size ... with 2 Axes>
+
+    Using zoom will up- or downscale the number of display pixels:
+
+    .. plot::
+      :context: close-figs
```
billbrod, line 322: don't need to show this example, they've already seen the einstein image by itself above

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+
+      >>> import plenoptic as po
+      >>> import torch
+      >>> imgs = torch.cat([po.data.einstein(), po.data.einstein() * 2])
+      >>> po.plot.imshow(imgs, vrange="auto1")
+      <PyrFigure size ... with 2 Axes>
+
+    Using zoom will up- or downscale the number of display pixels:
```
billbrod:
```suggestion
    The ``zoom`` argument allows users to set the ratio of display to image pixels, increasing or
    decreasing the size of the resulting plot:
```

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+      >>> import plenoptic as po
+      >>> po.plot.imshow(po.data.einstein())
+      <PyrFigure size ... with 1 Axes>
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> po.plot.imshow(po.data.einstein(), zoom=0.5)
+      <PyrFigure size ... with 1 Axes>
```
billbrod, line 355: include an example of setting an illegal zoom value, pointing out this because that zoom is meant to avoid aliasing, and so must result in an integer number of display pixels per image pixel.

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+
+    You can use the plot_complex flag for complex values:
```
billbrod:
```suggestion
    You can use the ``plot_complex`` argument to control how complex tensors are plotted:
```

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> import torch
+      >>> img = po.data.einstein()
+      >>> fft_img = torch.fft.fft2(img)
+      >>> po.plot.imshow([img, fft_img], plot_complex="polar")
```
billbrod:
```suggestion
      >>> po.plot.imshow([img, fft_img], plot_complex="logpolar")
```

maybe? "the amplitude plot doesn't look like anything for polar, since so much power is concentrated at the origin

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+      <PyrFigure size ... with 3 Axes>
+
+    Value errors can occur if the as_rgb flag and image dimensions are misaligned:
```
billbrod:
```suggestion
    Value errors can occur if ``as_rgb=True`` and the input image doesn't have the required number of channels
```

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> import torch
+      >>> img = po.data.einstein()
+      >>> fft_img = torch.fft.fft2(img)
+      >>> po.plot.imshow([img, fft_img], plot_complex="polar")
+      <PyrFigure size ... with 3 Axes>
+
```
billbrod, line 365: should show an example of successful color image plotting first

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+    >>> import plenoptic as po
+    >>> img = torch.cat([po.data.color_wheel(), po.data.color_wheel()])
+    >>> print(img.shape)
+    torch.Size([2, 3, 600, 600])
+    >>> po.plot.imshow(img)
+    Traceback (most recent call last):
+    ValueError: Don't know how to plot non-rgb images with more than one channel and ...
+
+    Images will automatically rescaled to be displayed at the same height and widths
```
billbrod:
```suggestion
    Images will be automatically rescaled to be displayed at the same height and widths
```

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,133 @@ def imshow(
...
+    Traceback (most recent call last):
+    ValueError: If as_rgb is True, then channel must have 3 or 4 elements!
+
+    >>> import plenoptic as po
+    >>> img = torch.cat([po.data.color_wheel(), po.data.color_wheel()])
+    >>> print(img.shape)
+    torch.Size([2, 3, 600, 600])
+    >>> po.plot.imshow(img)
+    Traceback (most recent call last):
+    ValueError: Don't know how to plot non-rgb images with more than one channel and ...
```
billbrod: I now think this example is a little more complex, and think it can probably be removed; it needs more explanation than I think is warranted. The error message is confusing here, because these are RGB images, we just haven't told that to imshow at this point.

billbrod (changes requested):
Thanks Emily! Left more suggestions in line.

I also think we need an example setting `title` and `col_wrap`. They can potentially be in the same example (they're not complicated), right after the example passing the list of images

src/plenoptic/plot/display.py
emeyer121: Thinking it maybe makes more sense to have auto1 first with the following descriptions?

The vrange argument allows control over the min and max values of the color range. In addition to a 2-tuple of labels, plenoptic includes several special strings: ``"auto1"`` sets all images to have the same range
[auto1 figure w/ one image darker]
Meanwhile, ``"indep1"`` sets each range independently. Note the different ranges in the titles!
[indep1 figure]

billbrod: yep, that sounds good to me!

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,180 @@ def imshow(
...
+      >>> import plenoptic as po
+      >>> import torch
+      >>> imgs = torch.cat([po.data.einstein(), po.data.curie()])
+      >>> print(imgs.shape)
+      torch.Size([2, 1, 256, 256])
+      >>> po.plot.imshow(imgs, batch_idx=1)
+      <PyrFigure size ... with 1 Axes>
+
+    The vrange argument allows control over the min and max values of the color range.
+    In addition to a 2-tuple of labels, plenoptic includes several special strings:
```
billbrod:
```suggestion
    In addition to a 2-tuple of floats, this function accepts several special strings:
```

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,180 @@ def imshow(
...
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> po.plot.imshow(po.data.einstein(), zoom=0.5)
+      <PyrFigure size ... with 1 Axes>
+
+    If ``zoom<1`` and the value is not a divisor of the largest image size, it will
+    result in an error:
```
billbrod:
```suggestion
    If ``zoom<1`` and the value is not a divisor of the largest image size, this
    function will raise an error:
```

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,180 @@ def imshow(
...
+    To plot an RGB image, you must set ``as_rgb=True`` to plot a color image:
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> color_wheel = po.data.color_wheel()
+      >>> print(color_wheel.shape)
+      torch.Size([1, 3, 600, 600])
+      >>> po.plot.imshow(color_wheel, as_rgb=True)
```
billbrod:
```suggestion
      >>> po.plot.imshow(color_wheel, as_rgb=True, zoom=0.5)
```

just because this is such a big image

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,180 @@ def imshow(
...
+    separate grayscale image:
+
+    .. plot::
+      :context: close-figs
+
+      >>> import plenoptic as po
+      >>> color_wheel = po.data.color_wheel()
+      >>> print(color_wheel.shape)
+      torch.Size([1, 3, 600, 600])
+      >>> po.plot.imshow(color_wheel)
```
billbrod:
```suggestion
      >>> po.plot.imshow(color_wheel, zoom=0.5, title=["R", "G", "B"])
```

src/plenoptic/plot/display.py
```diff
@@ -234,6 +234,180 @@ def imshow(
...
+      <PyrFigure size ... with 3 Axes>
+
+    Value errors can occur if ``as_rgb=True`` and the input image doesn't have the
+    required number of channels:
```
billbrod:
```suggestion
    This function will raise a `ValueError` if ``as_rgb=True`` and the input image 
    doesn't have the required number of channels:
```

billbrod (changes requested):
We're almost there! One thing I'm just noticing -- currently, you've wrote each example so as to be completely independent (e.g., you're re-importing `plenoptic` in each of them). I decided against doing it that way -- you can reuse objects (including imports) from previous blocks. Can you go through and remove the duplicate imports? And reuse image tensors where possible (e.g., use `einstein` rather than `po.data.einstein()` after the first block, reuse `imgs`). It makes it clearer what the focus of the example is.

emeyer121:
Cleaned up the redundancies in the code and made variables names more consistent. The only thing I left that is a bit redundant is printing out images sizes in a few cases where it's relevant (``batch_idx``, ``zoom`` error, and ``as_rgb`` error), but happy to remove if you think it's unnecessary.

src/plenoptic/plot/display.py
```diff
@@ -295,25 +289,20 @@ def imshow(
...
-      >>> import torch
-      >>> imgs = torch.cat([po.data.einstein(), po.data.curie()])
       >>> print(imgs.shape)
       torch.Size([2, 1, 256, 256])
       >>> po.plot.imshow(imgs, batch_idx=1)
       <PyrFigure size ... with 1 Axes>
 
     The vrange argument allows control over the min and max values of the color range.
-    In addition to a 2-tuple of labels, plenoptic includes several special strings:
+    In addition to a 2-tuple of floats, this functions accepts several special strings:
```
billbrod:
```suggestion
    In addition to a 2-tuple of floats, this functions accepts several special strings (see docstring for details). For example,
```

billbrod: don't want it to sound like "auto1" and "indep1" are the only such strings

src/plenoptic/plot/display.py
```diff
@@ -334,17 +320,15 @@ def imshow(
...
       :context: close-figs
 
-      >>> import plenoptic as po
-      >>> po.plot.imshow(po.data.einstein(), zoom=0.5)
+      >>> po.plot.imshow(einstein, zoom=0.5)
       <PyrFigure size ... with 1 Axes>
 
-    If ``zoom<1`` and the value is not a divisor of the largest image size, it will
-    result in an error:
+    If ``zoom<1`` and the value is not a divisor of the largest image size, this
```
billbrod:
```suggestion
    Note that if ``zoom<1`` and the value is not a divisor of the largest image size, this
```

src/plenoptic/plot/display.py
```diff
@@ -354,23 +338,19 @@ def imshow(
...
-      >>> import plenoptic as po
-      >>> import torch
-      >>> img = po.data.einstein()
-      >>> fft_img = torch.fft.fft2(img)
-      >>> po.plot.imshow([img, fft_img], plot_complex="logpolar")
+      >>> einstein_fft = torch.fft.fft2(einstein)
+      >>> po.plot.imshow([einstein, einstein_fft], plot_complex="logpolar")
       <PyrFigure size ... with 3 Axes>
 
     To plot an RGB image, you must set ``as_rgb=True`` to plot a color image:
```
billbrod:
```suggestion
    To plot a RGB(A) image in color, you must set ``as_rgb=True``:
```

billbrod (changes requested):
I think the places you left those printing out of sizes are helpful, I just have a few small tweaks and then I think this is ready to go!

billbrod (approved):
This is great, thanks Emily!

PR #446, opened 2026-06-02
Fix doctest issue, adds basic `__repr__` for synthesis objects

### LINKED ISSUE #439: Not all doctests running
With #418, some of our doctests have stopped running: all in `src/plenoptic/_synthesize` and `src/plenoptic/models`. I have not been able to figure out a simple reason why this is the case, but when running `pytest --doctest-modules -n 0 src/plenoptic --collect-only -vvvvv`, I can see that pytest doesn't even see the modules there.

I thought this might have something to do with lazy loading, because of [this caveat](https://scientific-python.org/specs/spec-0001/#caveats), but that situation doesn't sound the same as ours. In particular, `src/plenoptic/models/naive.py` contains multiple objects with doctests, none called `Naive` and none get run.

The issue only appears to happen with classes and appears to happen with all our classes: everything in `models` and `_synthesize` are classes, and it appears that `src/plenoptic/process/laplacian_pyramid.py` and `src/plenoptic/process/steerable_pyramid_freq.py` also don't gather the doctests present.

You can see 110 tests running [in this build](https://github.com/plenoptic-org/plenoptic/actions/runs/24618728925/job/71985470822) before merging #418, and only [57 running](https://github.com/plenoptic-org/plenoptic/actions/runs/24790074889/job/72544821135) during that PR. I haven't done any further splits to see when during that PR the problem started.

billbrod:

This fixes #439, where I noticed that not all the doctests where running. After investigating, the culprit turned out to be manually over-riding `__module__` in my objects in #418. The reason I had done that was to improve the string representation of the objects.

So here, I remove that manual over-riding, make a small change to the loading code to remain functional, and add a basic `__repr__` for each of the synthesis objects (the models, my only other objects, don't need one, since they inherit it from `torch.nn.Module`). This basic `__repr__` just summarizes how the object was initialized.

An extension of this could provide some information on synthesis status, but I think this is sufficient for now.

Closes #439

**Describe changes**

- Remove `__module__` from all objects
- Updates loading code to map synthesis object names to `plenoptic.OBJ_NAME` for the purposes of checking object identity.
- Adds basic `__repr__` to synthesis objects, making use of helper method in `_Synthesis`
- Adds tests for `__repr__`

**Checklist**

### DIFF
--- a/src/plenoptic/_synthesize/eigendistortion.py
+++ b/src/plenoptic/_synthesize/eigendistortion.py
@@ -74,8 +74,6 @@ class Eigendistortion(_Synthesis):
            https://www.cns.nyu.edu/~lcv/eigendistortions/
     """
 
-    __module__ = "plenoptic"
-
     def __init__(self, image: Tensor, model: torch.nn.Module):
         super().__init__()
         validate_input(image, no_batch=True)
--- a/src/plenoptic/_synthesize/eigendistortion.py
+++ b/src/plenoptic/_synthesize/eigendistortion.py
@@ -654,6 +654,10 @@ class Eigendistortion(_Synthesis):
         # it here
         self._init_representation(self.image)
 
+    def __repr__(self) -> str:
+        # numpydoc ignore=GL08
+        return super()._repr_format(["image", "model"])
+
     @property
     def model(self) -> torch.nn.Module:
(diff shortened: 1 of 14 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/446

PR #452, opened 2026-06-09
Hotfix: correct loss animation

billbrod:

In #438, I changed the curve plotted by `synthesis_loss` for the metamer object to be the "metamer loss", rather than the output of the objective function. This means its `losses - penalty_lambda * penalties`, but the update in `synthesis_animate` (for the red dot on the loss plot) is just using the `losses` attribute.

You can see the result in the [quickstart](https://docs.plenoptic.org/docs/tags/2.0.0/getting_started/quickstart.html) notebook, where the red dot is slightly above the loss curve (for syntheses of a reasonable length, this effect will be subtle, but it's pretty obvious if you run it for 10 iterations).

**Checklist**

### DIFF
--- a/src/plenoptic/plot/synthesis.py
+++ b/src/plenoptic/plot/synthesis.py
@@ -1998,7 +1998,13 @@ def synthesis_animate(
 
     if isinstance(synthesis_object, Metamer):
         saved_synth = synthesis_object.saved_metamer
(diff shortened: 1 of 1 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/452

PR #453, opened 2026-06-09
adding a doc test for to_numpy()

raffles-xingqi-zhu:

> This PR adds a code example for the function to_numpy(). 

> This contributes to adding user facing examples in tensor.py in Issue #237. 

> - Please check everything looks correct since it's my first time! 
> - Do we ever run into tensors with rank>3 in any part of Plenoptic for any reason? (My guess is probably not)

**Describe changes**

In a list, describe the changes you made. This should be more granular than the description at the beginning of the PR. These descriptions can be short, but this would also be the place to describe why something seemingly unrelated was included. e.g.,

> - Tensor shape is chosen to match a 2D greyscale image. 
> - The effect of "squeeze" parameter is demonstrated. 

**Checklist**

### DIFF
--- a/src/plenoptic/tensors.py
+++ b/src/plenoptic/tensors.py
@@ -82,7 +82,10 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
     >>> print(f"Type: {type(array2)}, Shape: {array2.shape}")
     Type: <class 'numpy.ndarray'>, Shape: (256, 256)
 
-    This is handy for saving the synthesized metamer as an image
+    This is handy for saving the synthesized metamer as an image to disk (here, we use
+    :func:`plenoptic.convert_float_to_int` to convert the image to an 8-bit integer
+    for saving).
+
     >>> import plenoptic as po
     >>> import torch
     >>> import imageio
@@ -94,10 +97,8 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
     >>> met.to(torch.float64)
     >>> met.load(po.data.fetch_data("example_metamer_gaussian.pt"))
     >>> met_squeezed = po.to_numpy(met.metamer, squeeze=True)
-    >>> met_squeezed = np.clip(met_squeezed * 255, 0, 255).astype("uint8")
+    >>> met_squeezed = po.convert_float_to_int(np.clip(met_squeezed, 0, 1))
     >>> imageio.v3.imwrite("example_metamer_gaussian.png", met_squeezed)
-
-    .. image:: example_metamer_gaussian.png
     """
     with contextlib.suppress(AttributeError):
         # if this fails, it's already a numpy array
--- a/src/plenoptic/tensors.py
+++ b/src/plenoptic/tensors.py
@@ -88,6 +88,7 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
 
     >>> import imageio
     >>> import numpy as np
+    >>> # Load in an example metamer synthesis
     >>> img = po.data.einstein()
     >>> model = po.models.Gaussian(30).eval()
     >>> po.remove_grad(model)
--- a/src/plenoptic/tensors.py
+++ b/src/plenoptic/tensors.py
@@ -74,7 +74,7 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
     --------
     >>> import plenoptic as po
     >>> import torch
-    >>> tensor = torch.rand((256, 256, 1))
+    >>> tensor = torch.rand((1, 1, 256, 256))
     >>> array1 = po.to_numpy(tensor)
     >>> print(f"Type: {type(array1)}, Shape: {array1.shape}")
     Type: <class 'numpy.ndarray'>, Shape: (256, 256, 1)
--- a/src/plenoptic/tensors.py
+++ b/src/plenoptic/tensors.py
@@ -84,10 +84,12 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
 
     This is handy for saving the synthesized metamer as an image to disk (here, we use
     :func:`plenoptic.convert_float_to_int` to convert the image to an 8-bit integer
-    for saving). Note that we clip the image to $[0, 1]$ before converting and saving;
-    it contains some values just outside that range. In general, it's up to the user to
-    ensure this is reasonable (you could also use :func:`plenoptic.process.rescale` to
-    linearly rescale the image to that same range).
+    for saving).
+
+    Note that we clip the image to [0, 1] before converting and saving; it contains some
+    values just outside that range. In general, it's up to the user to ensure this is
+    reasonable (you could also use :func:`plenoptic.process.rescale` to linearly rescale
+    the image to that same range).
 
     >>> import imageio
     >>> import numpy as np
--- a/src/plenoptic/tensors.py
+++ b/src/plenoptic/tensors.py
@@ -72,6 +72,8 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
 
     Examples
     --------
+    >>> import plenoptic as po
+    >>> import torch
     >>> tensor = torch.rand((256, 256, 1))
     >>> array1 = po.to_numpy(tensor)
     >>> print(f"Type: {type(array1)}, Shape: {array1.shape}")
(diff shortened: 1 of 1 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/453

src/plenoptic/tensors.py
```diff
@@ -69,6 +69,18 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
     -------
     array
         The former tensor, now an array.
+
+    Examples
+    --------
+    >>> import plenoptic as po
+    >>> import torch
+    >>> tensor = torch.rand((256, 256, 1))
```
billbrod:
```suggestion
    >>> tensor = torch.rand((1, 1, 256, 256))
```

images are generally represented with shape `(batch, channel, height, width)` in pytorch, so let's go with that in this example.

billbrod (changes requested):
Thanks for this Raffles, this example looks good! However, I'd say the most common use of this function is to take the outputs of a model or metamer and convert them to numpy for saving as an image (and I'm realizing I should've told you this beforehand)

So can we add an additional examples showing that? it should load in an example metamer (as you can see in the first example in `metamer_representation_error` in `plot/metamer.py`) and convert the `met.metamer` attribute to a numpy array (squeezing it), save it to a png with [imageio.imwrite](https://imageio.readthedocs.io/en/stable/_autosummary/imageio.v3.imwrite.html#imageio.v3.imwrite) and then use `    .. image:: to_numpy-example.png` to show it (you can see an example of using `.. image::` in the examples for `synthesis_animate`  in `plot/synthesis.py`, and we can chat about it tomorrow)

src/plenoptic/tensors.py
```diff
@@ -69,6 +69,35 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
...
+    >>> import torch
+    >>> tensor = torch.rand((1, 1, 256, 256))
+    >>> array1 = po.to_numpy(tensor)
+    >>> print(f"Type: {type(array1)}, Shape: {array1.shape}")
+    Type: <class 'numpy.ndarray'>, Shape: (1, 1, 256, 256)
+    >>> array2 = po.to_numpy(tensor, squeeze=True)
+    >>> print(f"Type: {type(array2)}, Shape: {array2.shape}")
+    Type: <class 'numpy.ndarray'>, Shape: (256, 256)
+
+    This is handy for saving the synthesized metamer as an image
```
billbrod: I think you need a blank new-line here for the docs to render correctly. you can build the docs locally to check how things look before pushing (see contributing guide for details)

billbrod:
```suggestion
    This is handy for saving the synthesized metamer as an image to disk (here, we use 
    :func:`plenoptic.convert_float_to_int` to convert the image to an 8-bit integer, as needed 
    for saving).
```

or some other description of that conversion process

(note that my suggestions might be too line, so it might be better to make these changes yourself locally so you can ensure they look good)

src/plenoptic/tensors.py
```diff
@@ -69,6 +69,35 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
...
+    >>> import imageio
+    >>> import numpy as np
+    >>> img = po.data.einstein()
+    >>> model = po.models.Gaussian(30).eval()
+    >>> po.remove_grad(model)
+    >>> met = po.Metamer(img, model)
+    >>> met.to(torch.float64)
+    >>> met.load(po.data.fetch_data("example_metamer_gaussian.pt"))
+    >>> met_squeezed = po.to_numpy(met.metamer, squeeze=True)
+    >>> met_squeezed = np.clip(met_squeezed * 255, 0, 255).astype("uint8")
```
billbrod: use https://docs.plenoptic.org/docs/pulls/453/api/generated/plenoptic.convert_float_to_int.html for this

and add a bit to the description above about this, as I noted.

src/plenoptic/tensors.py
```diff
@@ -69,6 +69,35 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
...
+    >>> imageio.v3.imwrite("example_metamer_gaussian.png", met_squeezed)
+
+    .. image:: example_metamer_gaussian.png
```
billbrod: as I mentioned in my message, maybe this isn't the best way to end this example. but instead load in https://imageio.readthedocs.io/en/stable/_autosummary/imageio.v3.imread.html after saving and print out its representation? or maybe just show that it can be saved.

billbrod: what do you think? we can talk about it in person too

billbrod (changes requested):
Thanks Raffles! Some more changes here. I'm not sure why exactly that sphinx job is failing -- but if we end up switching how we finish this test it won't matter. (to figure it out, I'd build the docs locally and see if you can find the saved `example_metamer_gaussian.png` file in the `docs/_build/html` folder and, if so, if you can open it.)

src/plenoptic/tensors.py
raffles-xingqi-zhu: One thing I noticed is that the metamer image contains pixel values outside 0 and 1. I'll need to check how this is handled in the convert_float_to_int() function. My guess is as long as the increment/decrement above/below the cap values are small they will effectively get rounded to the closest integer, 0 and 255. Thoughts on this?

src/plenoptic/tensors.py
raffles-xingqi-zhu: I think just showing we can save it is enough. Displaying the PNG doesn't add much.

src/plenoptic/tensors.py
raffles-xingqi-zhu: That's super handy to know. I wasn't aware of that!

src/plenoptic/tensors.py
billbrod: `convert_float_to_int` doesn't do anything to handle this, it's expected that folks will clip it before passing an image in, as you do here.

billbrod: In general, it's up to the user to ensure that they're not throwing away too much info.

src/plenoptic/tensors.py
```diff
@@ -69,6 +69,36 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
...
+    >>> array2 = po.to_numpy(tensor, squeeze=True)
+    >>> print(f"Type: {type(array2)}, Shape: {array2.shape}")
+    Type: <class 'numpy.ndarray'>, Shape: (256, 256)
+
+    This is handy for saving the synthesized metamer as an image to disk (here, we use
+    :func:`plenoptic.convert_float_to_int` to convert the image to an 8-bit integer
+    for saving).
+
+    >>> import plenoptic as po
+    >>> import torch
```
billbrod:
```suggestion
```

we don't reimport everything in each block, to avoid cluttering it up

src/plenoptic/tensors.py
```diff
@@ -69,6 +69,36 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
...
+    >>> import imageio
+    >>> import numpy as np
+    >>> img = po.data.einstein()
```
billbrod, line 97:
```suggestion
    >>> # Load in an example metamer synthesis
    >>> img = po.data.einstein()
```

just to clarify what we're doing here

src/plenoptic/tensors.py
```diff
@@ -69,6 +69,36 @@ def to_numpy(x: Tensor | np.ndarray, squeeze: bool = False) -> np.ndarray:
...
+    >>> array1 = po.to_numpy(tensor)
+    >>> print(f"Type: {type(array1)}, Shape: {array1.shape}")
+    Type: <class 'numpy.ndarray'>, Shape: (1, 1, 256, 256)
+    >>> array2 = po.to_numpy(tensor, squeeze=True)
+    >>> print(f"Type: {type(array2)}, Shape: {array2.shape}")
+    Type: <class 'numpy.ndarray'>, Shape: (256, 256)
+
+    This is handy for saving the synthesized metamer as an image to disk (here, we use
+    :func:`plenoptic.convert_float_to_int` to convert the image to an 8-bit integer
+    for saving).
```
billbrod, line 87:
```suggestion
    for saving). Note that we clip the image to $[0, 1]$ before converting and saving;
    it contains some values just outside that range. In general, it's up to the user to
    ensure this is reasonable (you could also use :func:`plenoptic.process.rescale` to
    linearly rescale the image to that same range).
```

billbrod (changes requested):
A couple of really minor changes and then this is good to go!

billbrod (approved):
Thanks, this looks good! I pushed a small change (removing the $, since that doesn't actually render correctly).

The tests are failing because of an update to pytest, so I'm going to fix that in a separate PR and then we'll merge this!

PR #457, opened 2026-06-16
Fix for class scoped fixtures

billbrod:

Pytest 9.1 deprecates [class-scoped fixtures as instance methods](https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method), which we were using in several of our tests. Therefore, that release caused many of our tests to start raising the deprecation warning and thus fail. Here, we update to the recommended way of using them: turning them into class methods. We don't actually *use* the behavior this enables, so it doesn't require any other changes.

**Checklist**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,8 @@ docs = [
 ]
 
 test = [
-    "pytest>=5.1.2",
+    # 9.1.0 unsupported because of https://github.com/pytest-dev/pytest/issues/14591
+    "pytest>=5.1.2,!=9.1.0",
     'pytest-cov',
     'pytest-xdist',
     'pytest-order',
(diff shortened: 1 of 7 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/457

billbrod:
Current failing tests are due to: https://github.com/pytest-dev/pytest/issues/14591 and will be fixed in `pytest=9.1.1` release.

> using `@pytest.mark.parametrize(..., indirect=True)` does not seem to override parameters of a fixture that has parameters explicitly set (i.e., `@pytest.fixture(params=[...]))` - attempting to do so now results in `duplicate parametrization of {fixture_name}` error.

Will push an update dropping support for `pytest=9.1.0` specifically

PR #460, opened 2026-06-22
Adds `DeepNetFeatures`

### LINKED ISSUE #396: Add examples for working with other packages' models
Can add small pages to the documentation that show how to use models from [timm](https://huggingface.co/timm) and [torchvision](https://docs.pytorch.org/vision/main/models.html) with plenoptic. This might require writing a small adapter to the codebase as well.

See [brainhack-2025](https://github.com/plenoptic-org/brainhack-2025) for how to do this.

billbrod:

This PR adds a new model, `DeepNetFeatures`. This is essentially a nice wrapper around torchvision's feature extraction capabilities, and simplifies the use of deepnets from [torchvision](https://docs.pytorch.org/vision/main/models.html) and [timm](https://huggingface.co/docs/timm/reference/models) with plenoptic (other models should also work, but those are the two I've tested with).

I wrote `DeepNetFeatures` to work with any number of layers. This is not used here, but will be useful for #179.

The main contribution here is the new `src/plenoptic/models/feature_extractor.py` file and the corresponding pages in the user guide: [one demonstrating basic usage](https://docs.plenoptic.org/docs/pulls/460/user_guide/models_and_metrics/deep_nets.html) and [one reproducing some examples from Feather et al., 2023](https://docs.plenoptic.org/docs/pulls/460/user_guide/reproduce/feather2023.html). ~That page got complicated because I wanted to show how use it with models from either timm or torchvision, and metamers from any of the three layers I looked at. That involved using a lot of tabs and you can't run code in tabs so... the resulting page is a myst notebook and has some executable cells, but it also has a single hidden cell and a lot of (non-executed) code blocks. The actual image showing the metamer is generated by the matplotlib sphinx extension from paired helper scripts in `docs/scripts/` (which are the source for the non-executed code blocks). I think this all works (I test the new script separately in a new CI job), but want to make sure I'm not missing a simpler way of handling this.~ (description of old docs page, now split in two and simplified.)

Closes #396, facilitates #179

Exploration of optimization hyperparameters [here](https://github.com/plenoptic-org/plenoptic-experiments/tree/main/deepnet_metamer). I figured out how to use [altair](https://altair-viz.github.io/) to create a scatter plot with embedded images -- download the `results-macaque.html` file and open it up in your browser to see.

Probably most useful to look at rendered docs page (linked above) and [API docs](https://docs.plenoptic.org/docs/pulls/460/api/generated/plenoptic.models.DeepNetFeatures.html)

- ~How's the name `FeatureExtractorModel`? I was originally calling this `TorchVisionAdapter`, which is both too narrow and not actually accurate. I want to highlight that this is making use of the feature extractor and so anything that works with that will work with this, but I worry it's a little too abstract (hence why I called the docs page "using deep nets with plenoptic", but the feature extractor will work with non-deep net models as well, so I didn't want to call it `DeepNetAdapter` or something)~ Decided on `DeepNetFeatures` since that is the primary use-case
- ~The new notebook contains several code-blocks where I want to show the output (e.g., what the torchvision transform looks like). I think the proper way to do this is with [sphinx.ext.doctest](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html) and/or [sybil](https://sybil.readthedocs.io/en/latest/patterns.html). I'm going to look into that for #459, but there's not something else I'm missing is there?~ I still need #459 for some of the blocks, but not as much.
- ~I used [`literalinclude`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) to show lines from the accompanying helper script for those non-executed code blocks. This way I know that the lines run (that there's not some dumb typo in the version displayed in the notebook). I was originally `:lines:` to control what was being shown, but using it was easy for them to get off by some number while I was editing the helper script. Now I'm using `:start-at:`/`:end-at:`, which should be more resilient, but still, ideally I'd have some way of checking that they are the lines I mean, or of pairing the two. Any idea?~ no longer used
- ~In order to get the "summary stats" for the different layers' metamers at the end of the notebook, I'm actually generating a png with matplotlib to visualize, the same way I'm getting the `synthesis_status` figure. I don't think there's a clean alternative to that?~ no longer doing this
- ~Because this notebook is such a tangled mess, I don't provide links to let people download/run the notebook at the top. instead, I provide a link to download the accompanying script. Does that make senes? should I improve the docs in that script?~ simplified so works like the rest of the notebooks
- I don't change how I handle dependencies here: torchvision was already an optional dependency, and I mark it in the notebooks where it's required, plus with an adomnition in FeatureExtractorModel docstring. anything else I should do?
    - timm, which isn't required for anything but is an alternative model zoo, is now added to my docs/tests bundle (where torchvision already was). it's mentioned in the parts of the docstring/notebook where required, but that's it. 
- My notebook shows how to generate a metamer similar to the ones in [this paper](https://www.nature.com/articles/s41593-023-01442-0), but the method is different. Is the description of the relationship between what we did vs. what they did clear enough?

**Describe changes**

- Adds `DeepNetFeatures` with associated tests, doctests, new docs page.
- Removes warning in validate_model if model output is 2d. That's what FeatureExtractorModel gives, and that works without any problems (remove associated test as well)
- Adds new built-in example image, macaque, which is one of the [monkey selfies](https://en.wikipedia.org/wiki/Monkey_selfie). In addition to being a fun (public domain) image, it's also more ImageNet like than any of my current examples: it's in color and features a single object in the center of frame whose category is reasonably captured by one of the [ImageNet1K  classes](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/) ("guenon", which is a different type of old-world monkey).
- ~Adds new CI job, "run additional doc scripts" to run the new `docs/scripts/feature_extractor.py`, since not all variants are run during the docs build, and I want users to be able to use it.~ removed
- Made some changes to how `plot_representation` and `update_plot` work:
    - `plot_representation` and `update_plot` now accept either dict or tensor
    - rescaling of ylimits is now handled within `update_plot`, and so can be handled directly by the model
    - `_rescale_ylim` doesn't try to update ylim on image artists, because that's always wrong (ylim controls the extent of the plot)
    - moves `update_plot` dimensionality check after trying to call `model.update_plot` because if model knows how to handle we don't care about its dimensionality
    - `plot_representation` a little more flexible in how it lays out plots
    - Documented required signature for `model.plot_representation` if a model is going to implement it to be compatible with the synthesis plotting machinery
- Updates Eigendistortion examples using ResNet / VGG to use the new model.
    - As part of this, realized that the input image used for the VGG eigendistortion, reproducing the original paper, was 254x254 (instead of standard 224x224). looking back at the paper, I think that is what they used, so left it as it
    - (Changed the example not reproducing a result from the literature to use 224x224 input)
    - Changed the ImageNet normalization to be the actual one instead of approximately correct (previously, was just z-scoring using the single images mean/std, rather than the whole dataset's)
- Sets rcParams for the matplotlib sphinx directive acros the docs to set the DPI and `savefig.bbox="tight"`. This has slightly changed the appearance of some figures in the API guide

**Checklist**


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/460

emeyer121:
This is really exciting! I will add more nitpicky comments directly in the code but some general thoughts here:
- I'm not loving the name ``FeatureExtractorModel`` especially because it's called as ``models.FeatureExtractorModel`` which is a bit repetitive and it's more pulling out features from models rather than being it's own model. I feel like even calling it ``FeatureExtractor`` or ``NetworkFeatureExtractor`` could work, or even ``ModelFeatureExtractor`` to make it more clear that it's extracting features from a model.
- It's a little odd to me that the output of extracting from multiple layers is just a vector concatenated across all layers. Obviously using the ``convert_to_dict`` is great and necessary, but if all I care about is extracting all of the features from a bunch of layers, it seems like a bit of an annoying extra step.
- Regarding the user guide, I think it's generally pretty clear how each of the steps work. I'm a little confused why there's not an ``.ipynb`` script matching it like there are with the other models. I feel like one can be made where you just use one layer with all of the steps and encourage people to look at results using other layers as their own test (while keeping the user guide the same!)
- If you want to keep just the ``.py`` file, I think it needs to be made more clear that it's not an exact replica of what the user guide is showing and also note that both ``torchvision`` and ``timm`` are required to run it as is.
- I think the description of what was different in the tutorial vs what is done in the paper is sufficient! Generally makes sense and I think there's enough info if people really want to dig into it.

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+    Parameters
+    ----------
+    model
+        The pytorch module to use.
+    return_nodes
+        The names of the nodes to return. See Examples and
+        :external+torchvision:doc:`torchvision documentation <feature_extraction>`.
+    transform
+        Pre-processing transform to apply to image before passing to model. If
+        ``None``, will not apply any transform.
```
emeyer121: Maybe it's just me but I was unsure at first glance whether ``transform`` was something you were applying or something external. Figured it was external because it's input is a ``Module`` and not ``True/False`` but could make it more clear with something like "Torchvision or timm preprocessing transform Module"

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+       >>> # size.
+       >>> img = po.process.center_crop(
+       ...     po.data.einstein(False), tv_transform.crop_size[0]
+       ... )
+       >>> img.shape
+       torch.Size([1, 3, 224, 224])
+       >>> model(img).shape
+       torch.Size([1, 401408])
+       >>> po.remove_grad(model)
+       >>> po.validate.validate_model(model, image_shape=img.shape)
```
emeyer121: Probably useful to add some comment here about what ``remove_grad`` and ``validate_model`` do here. Especially because my association with validating some network model is more training/testing/validation, which is not what this ``validate_model`` is doing here.

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+       )
+       >>> norm = torchvision.transforms.Normalize(tv_transform.mean, tv_transform.std)
+       >>> model = po.models.FeatureExtractorModel(tv_model, "layer2", norm)
+       >>> # this model requires a 3d input, and expects it to have a certain input
+       >>> # size.
+       >>> img = po.process.center_crop(
+       ...     po.data.einstein(False), tv_transform.crop_size[0]
+       ... )
+       >>> img.shape
+       torch.Size([1, 3, 224, 224])
```
emeyer121, line 113: It's confusing to me why the comment is saying "this model requires a 3d input" then ``img.shape`` is 4d but you don't have to squeeze it or anything. This is the case in future examples too.

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+           std=[0.229, 0.224, 0.225]
+           interpolation=InterpolationMode.BILINEAR
+       )
+       >>> norm = torchvision.transforms.Normalize(tv_transform.mean, tv_transform.std)
+       >>> model = po.models.FeatureExtractorModel(tv_model, "layer2", norm)
+       >>> # this model requires a 3d input, and expects it to have a certain input
+       >>> # size.
+       >>> img = po.process.center_crop(
+       ...     po.data.einstein(False), tv_transform.crop_size[0]
+       ... )
```
emeyer121, line 111: Very minor but might be helpful to write out ``po.data.einstein(as_gray=False)`` so people know what the ``False`` is referring to without going to it's own documentation. Same with future uses too.

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+        --------
+        convert_to_tensor
+            Convert dictionary representation to a 2d tensor.
+
+        Examples
+        --------
+        >>> import plenoptic as po
+        >>> import torchvision
+        >>> weights = torchvision.models.ResNet50_Weights.IMAGENET1K_V1
+        >>> tv_model = torchvision.models.resnet50(weights=weights)
```
emeyer121:
```suggestion
        >>> tv_model = torchvision.models.resnet50(weights=weights).eval()
```

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+        convert_to_dict
+            Convert tensor representation to a dictionary, whose keys are the feature
+            names, with their original shapes.
+
+        Examples
+        --------
+        >>> import plenoptic as po
+        >>> import torchvision
+        >>> weights = torchvision.models.ResNet50_Weights.IMAGENET1K_V1
+        >>> tv_model = torchvision.models.resnet50(weights=weights)
```
emeyer121:
```suggestion
        >>> tv_model = torchvision.models.resnet50(weights=weights).eval()
```

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
```
emeyer121:
```suggestion
        >>> tv_model = torchvision.models.resnet50(weights=weights).eval()
```

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+        self._packed_shapes = None
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """
+        Compute feature activity of an input.
+
+        We flatten across all dimensions except the batch / first dimension. This allows
+        us to support returning features of different shapes and dimensionality (as is
+        common across layers in deep nets), while still returning only a single tensor,
+        as is necessary for our synthesis methods.
```
emeyer121, line 257: Probably useful to mention somewhere that ``model.forward(img)`` is equivalent to ``model(img)`` (assuming I understand correctly that this is the case).

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+
+        Examples
+        --------
+        .. plot::
+           :context: reset
+
+           >>> import plenoptic as po
+           >>> import torchvision
+           >>> weights = torchvision.models.ResNet50_Weights.IMAGENET1K_V1
+           >>> tv_model = torchvision.models.resnet50(weights=weights)
```
emeyer121:
```suggestion
           >>> tv_model = torchvision.models.resnet50(weights=weights).eval()
```

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
```
emeyer121:
```suggestion
           >>> tv_model = torchvision.models.resnet50(weights=weights).eval()
```

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+        --------
+        >>> import plenoptic as po
+        >>> import torchvision
+        >>> weights = torchvision.models.ResNet50_Weights.IMAGENET1K_V1
+        >>> tv_model = torchvision.models.resnet50(weights=weights)
+        >>> # This model's transform consists of resizing, cropping, and normalizing.
+        >>> # We recommend only including the normalizing in the transform.
+        >>> tv_transform = weights.transforms()
+        >>> norm = torchvision.transforms.Normalize(tv_transform.mean, tv_transform.std)
+        >>> model = po.models.FeatureExtractorModel(tv_model, "layer2", norm).eval()
```
emeyer121: Oh I guess here you use ``eval()`` after extracting the features? If this also works that's fine but probably good to be consistent or at least make notes about it. In the user guide it is explicitly mentioned that ``.eval()`` is needed but not in the docs, so probably good to specify.

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+    feature_extraction = None
+
+from ..plot import display
+
+
+class FeatureExtractorModel(torch.nn.Module):
+    """
+    Return features from model.
+
+    This adapter combines a torch model with a feature extractor and optional transform,
```
raffles-xingqi-zhu: We can link to the notebook showing how the function can be used here

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+       >>> # We recommend only including the normalizing in the transform.
+       >>> tv_transform = weights.transforms()
+       >>> tv_transform
+       ImageClassification(
+           crop_size=[224]
+           resize_size=[256]
+           mean=[0.485, 0.456, 0.406]
+           std=[0.229, 0.224, 0.225]
+           interpolation=InterpolationMode.BILINEAR
+       )
```
raffles-xingqi-zhu, line 104: It would be nice to start a new code block after the first output so it’s easier to read

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+
+    .. plot::
+       :context: reset
+
+       >>> import plenoptic as po
+       >>> import torchvision
+       >>> weights = torchvision.models.ResNet50_Weights.IMAGENET1K_V1
+       >>> tv_model = torchvision.models.resnet50(weights=weights).eval()
+       >>> # This model's transform consists of resizing, cropping, and normalizing.
+       >>> # We recommend only including the normalizing in the transform.
```
raffles-xingqi-zhu, line 95: It might be worth explaining why we don’t use the crop and resize in the original model transform but instead crop the image manually. Not all users reading the API will want to synthesize metamers. They could just want to visualize the activations in a specific layer for whatever reason.

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+    >>> model(img).shape
+    torch.Size([1, 501760])
+
+    The order of elements in ``return_nodes`` does not matter: the outputs are always
+    returned based on their order in ``model``.
+
+    >>> rep = model(img)
+    >>> model = po.models.FeatureExtractorModel(tv_model, ["layer4", "layer2"], norm)
+    >>> rep[0, 0] == model(img)[0, 0]
+    tensor(True)
```
raffles-xingqi-zhu, line 203: It is unclear in the documentation when you pass more than one layer what happens to the output size. For two layers, my understanding is that the output is of shape (batch, representation) with the representation size being features_1*x_1*y_1 + features_2*x_2*y_2. Is this correct?

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+       >>> model = po.models.FeatureExtractorModel(tv_model, "layer2", norm)
+       >>> # this model requires a 3d input, and expects it to have a certain input
+       >>> # size.
+       >>> img = po.process.center_crop(
+       ...     po.data.einstein(False), tv_transform.crop_size[0]
+       ... )
+       >>> img.shape
+       torch.Size([1, 3, 224, 224])
+       >>> model(img).shape
+       torch.Size([1, 401408])
```
raffles-xingqi-zhu, line 115: The default output of plenoptic.models.FeatureExtractorModel is a 2D tensor but personally it’s more helpful for me to see the the dictionary as the default (that shows me which layer of activations I’m looking at and the shape of activations). But I understand if there're conventions in the deep learning community. In that case we just need to make it clear in the case of multiple layers (see my previous comment).

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+
+    >>> [(k, v.shape) for k, v in model.convert_to_dict(model(img)).items()]
+    [('layer2', torch.Size([1, 512, 28, 28])), ('layer4', torch.Size([1, 2048, 7, 7]))]
+
+    Visualize model representation with :meth:`plot_representation`:
+
+    .. plot::
+       :context: close-figs
+
+       >>> fig, axes = model.plot_representation(model(img))
```
raffles-xingqi-zhu, line 217: What happened to layer4 representation in the visualization?

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+           :context: reset
+
+           >>> import plenoptic as po
+           >>> import torchvision
+           >>> weights = torchvision.models.ResNet50_Weights.IMAGENET1K_V1
+           >>> tv_model = torchvision.models.resnet50(weights=weights)
+           >>> # This model's transform consists of resizing, cropping, and normalizing.
+           >>> # We recommend only including the normalizing in the transform.
+           >>> tv_transform = weights.transforms()
+           >>> norm = torchvision.transforms.Normalize(
```
raffles-xingqi-zhu, line 482: Across different code blocks the initialization and transform are shown multiple times. Is there a way to define once at the beginning for brevity?

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+           >>> # this model requires a 3d input, and expects it to have a certain input
+           >>> # size.
+           >>> img = po.process.center_crop(
+           ...     po.data.einstein(False), tv_transform.crop_size[0]
+           ... )
+           >>> fig, axes = model.plot_representation(model(img))
+           >>> img = po.process.center_crop(
+           ...     po.data.curie(False), tv_transform.crop_size[0]
+           ... )
+           >>> model.update_plot(axes, model(img))
```
raffles-xingqi-zhu, line 495: Would be nice to add an example showing how update_plot() can be used for animation as the current example is not clear.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+:::{admonition} Run this notebook yourself!
+:class: important
+
+Download the script: [`feature_extractor.py`](../../scripts/feature_extractor.py)!
+
+:::
+
+(feature_extractor)=
+# Using Deep Neural Networks with plenoptic
```
raffles-xingqi-zhu: “Using deep neural networks with plenoptic” is quite general for a title. I feel like we can almost call it “Reproducing Feather et al., 2023”. (again, depends on whether we want to treat this as a reproduction example, see my general comment in the PR thread)

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+:::{warning}
+This notebook requires the optional dependency `torchvision`, which can be installed with `pip`.
+:::
+
+plenoptic is compatible with any model written in pytorch, including deep neural networks from the model zoos {external+torchvision:ref}`TorchVision <models>` and {external+timm:doc}`timm <models>`. In this notebook, we'll show how to adapt a deep net from these two packages for use with plenoptic, recreating some ResNet50 metamers shown in {cite:alp}`Feather2023-model-metam`, figure 2e.
```
raffles-xingqi-zhu:
```suggestion
Plenoptic is compatible with any model written in pytorch, including deep neural networks from the model zoos {external+torchvision:ref}`TorchVision <models>` and {external+timm:doc}`timm <models>`. In this notebook, we'll show how to adapt a deep net from these two packages for use with plenoptic, recreating some ResNet50 metamers shown in {cite:alp}`Feather2023-model-metam`, figure 2e.
```

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+:class: warning dropdown
+
+The example metamer shown in this notebook takes about 15 minutes to synthesize on a GPU. Thus, instead of performing synthesis in this notebook, we have cached the result of it online and only download them for investigation.
+
+:::
+
+
+## Initializing the model
+
+When synthesizing images for deep nets, as in {cite:alp}`Feather2023-model-metam`, it is common to pick a specific intermediate layer whose representation we wish to use. `torchvision` contains a "feature extractor" to grab activity from intermediate layers, and plenoptic's {class}`plenoptic.models.FeatureExtractorModel` is a small wrapper to simplify this process.
```
raffles-xingqi-zhu: Would be nice to link “feature extractor” to the torch documentation

docs/scripts/feature_extractor.py
```diff
@@ -0,0 +1,133 @@
...
+        deepnet, norm, crop, imagenet_categories = torchvision_setup()
+    elif model_zoo == "timm":
+        deepnet, norm, crop, imagenet_categories = timm_setup()
+    model = po.models.FeatureExtractorModel(deepnet, target_layer, norm)
+    po.remove_grad(model)
+    img = prepare_image(crop)
+
+    def get_category(image):
+        image_cat = po.to_numpy(
+            torch.nn.functional.softmax(deepnet(norm(image)), dim=1).squeeze()
```
raffles-xingqi-zhu: Should we put .cpu() after .squeeze() ? I’m actually not sure what is the principled way to use/omit it.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+fig, _ = model.plot_representation(rep)
+```
+
+## Synthesizing the metamer
+
+:::{warning}
+We do not perform synthesis in the exact same way as {cite:alp}`Feather2023-model-metam`. However, the resulting metamer is qualitatively similar. We note the differences below.
+:::
+
+Let us initialize our metamer object using the above image and model. Unlike in {cite:alp}`Feather2023-model-metam`, we are using the mean-squared error (the default for {class}`~plenoptic.Metamer`) as our loss function. We also initialize with a sample of uniformly-distributed noise whose values range from 0 to 1, whereas the paper initialized with "a sample from a normal distribution with a standard deviation of 0.05 and a mean of 0.5". Like that paper, we find better synthesis results if we use a learning-rate scheduler to halve the optimizer's learning rate regularly, using {class}`~torch.optim.lr_scheduler.StepLR` (see the following dropdown for more details):
```
raffles-xingqi-zhu: So what loss function did Feather et al., 2023 use? We should specify it.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+```{code-cell} ipython3
+met = po.Metamer(img, model)
+met.to(DEVICE)
+met.load(
```
raffles-xingqi-zhu: I’m getting an error after this code block: /home/jenkins/agent/workspace/CCN_neurorse_plenoptic_PR-460/lib/python3.12/site-packages/plenoptic/_synthesize/synthesis.py:562: UserWarning: You will need to call setup() to instantiate scheduler
warnings.warn(

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+:::{tab-item} layer4
+:sync: layer4
+
+```{eval-rst}
+.. plot:: scripts/feature_extractor_mpl.py layer4
+  :include-source: false
+```
+:::
+::::
+
```
raffles-xingqi-zhu: For the representation error over space (and model.plot_representation over space before), it could be beneficial to add a small colorbar showing the range and corresponding values. The value range in the title is pretty hard to read.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+In the above plots, we can see the metamer in the leftmost subplot, the loss over synthesis iterations in the middle, and the representation error on the right:
+- Our metamers match the results discussed earlier in this notebook:  the layer 2 metamer looks almost identical to the target image, the layer 3 metamer starts to add RGB noise, and the layer 4 is almost completely unidentifiable, looking almost completely like random RGB noise.
+- We can see that the optimization performed reasonably well: the loss decreased gradually over synthesis. If you were using these stimuli in an experiment (especially for `"layer4"`), it may be worth continuing a bit more to get the loss even lower, but these demonstrate the point.
```
raffles-xingqi-zhu: It is also interesting the converged loss increases as you go to deeper layers. What’s the intuition behind this? Did the paper show this as well?

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+- The representation error plot has the same structure as the {func}`~plenoptic.models.FeatureExtractorModel.plot_representation` plot above. We see that the error is fairly uniform across both space and channels.
```
raffles-xingqi-zhu: I also wouldn’t say that error is uniform.. there is clearly hotspots in space.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+The authors of {cite:alp}`Feather2023-model-metam` used two additional checks to verify that metamer synthesis had succeeded (quotes from "Results > Metamer optimization" section, pdf page 5):
+- "the metamer had to result in the same classification decision by the model as the reference stimulus" (here, `guenon`):
+- "measures of the match between the activations for the natural reference stimulus and its model metamer at the matched stage had to be much higher than would be expected by chance, as quantified with a null distribution". The authors used three measures here: Pearson and Spearman correlations and signal-to-noise ratio. Here, we show the Pearson correlation:
```
raffles-xingqi-zhu: Like you mentioned, the correlation values on their own can’t verify success unless compared against the distribution of correlations from the null distribution. Would that be an easy addition?

raffles-xingqi-zhu (commented):
Response to outstanding questions / particular feedback (in no particular order): 

1. `ModelFeatureExtractor()`  or `FeatureExtractor()` would be a better name. `FeatureExtractorModel` makes me think it is a model that does feature extraction, which doesn’t make sense. 
2. For the example doc, it is not crystal clear to me that the accompanied script is not the notebook in the documentation. The warning block even says “Run this notebook yourself!”
3. I wonder if we could provide a notebook that has all executable cells using only torchvision and one layer (e.g. layer 2). I always find interactive jupyter notebooks informative than .py script. On the documentation page, we can still show tabs of torchvision vs timm, layer 2 vs 3 vs 4, so the reader knows what the code and outputs would look like and could readily copy and paste to the notebook to test it out themselves. On the tabs we could write “layer 2 (used in notebook)” and “torchvision (used in notebook)”. This way we both showcase the different options (like good reproduction should do!) but give the user a notebook that they can run and modify. 
4. I feel the example doc belongs more to “Reproducing example from the literature” than “Models and metrics”. 
5. I don’t know why timm is not called an optional dependency but torchvision is on both the example and API docs.

docs/user_guide/models_and_metrics/feature_extractor.md
dherrera1911: My two cents: I like the current title. The tutorial shows how to use DNNs with plenoptic, which is a very important use case, so this should be easily findable by the title.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+# set seed for reproducibility
+po.set_seed(0)
+```
+
+:::{admonition} This notebook retrieves cached synthesis results
+:class: warning dropdown
+
+The example metamer shown in this notebook takes about 15 minutes to synthesize on a GPU. Thus, instead of performing synthesis in this notebook, we have cached the result of it online and only download them for investigation.
+
+:::
```
dherrera1911:
```suggestion
:::{admonition} Executing this notebook takes a long time
:class: important

The example metamer shown in this notebook takes about 15 minutes to synthesize on a GPU.
:::
```

dherrera1911: I think that the cache detail is too much detail for the user, although it is useful to mention that the notebook takes a while to run.

dherrera1911:
> That page got complicated because I wanted to show how use it with models from either timm or torchvision, and metamers from any of the three layers I looked at. That involved using a lot of tabs and you can't run code in tabs so... the resulting page is a myst notebook and has some executable cells, but it also has a single hidden cell and a lot of (non-executed) code blocks. The actual image showing the metamer is generated by the matplotlib sphinx extension from paired helper scripts in docs/scripts/ (which are the source for the non-executed code blocks). I think this all works (I test the new script separately in a new CI job), but want to make sure I'm not missing a simpler way of handling this.

The tutorial looks very nice with the usage of tabs. But if it is too messy, I would consider having the tutorial not use tabs, by just focusing on `torchvision` and having `timm` as an example in an admonition. The tutorial could just plot all three layers without using tabs. I think it would be nice if you could get rid of the extra `feature_extractor.py` script and have everything in one place.

dherrera1911:
> How's the name FeatureExtractorModel? I was originally calling this TorchVisionAdapter, which is both too narrow and not actually accurate. I want to highlight that this is making use of the feature extractor and so anything that works with that will work with this, but I worry it's a little too abstract (hence why I called the docs page "using deep nets with plenoptic", but the feature extractor will work with non-deep net models as well, so I didn't want to call it DeepNetAdapter or something)

I feel strongly that you should have `DeepNet` or `DNN` in the model name, like `DeepNetFeatures`. I imagine that that will be the overwhelming use of this new model, even if it could "work with non-deep net models as well". And DNNs being such a crucial use case, I think it is important to have a model called `DeepNet` in `plenoptic`.

docs/user_guide/models_and_metrics/feature_extractor.md
dherrera1911:
```suggestion
To synthesize images for deep nets, we have to pick a specific layer whose representation we wish to use. {class}`plenoptic.models.FeatureExtractorModel` allows to easily grab activity from a deep net intermediate layers.
```

dherrera1911: More focused emphasis on what the model subject of this tutorial does, rather than the secondary info of it being a wrapper around another's library similar functionality.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+```
+:::
+:::{tab-item} layer4
+:sync: layer4
+
+```python
+target_layer = "layer4"
+```
+:::
+::::
```
dherrera1911: These tabs that select the layer seem out of place here. Move to after the paragraph starting with "Next, we need to specify the layer to target." (current line 146 https://github.com/plenoptic-org/plenoptic/pull/460/changes#diff-290f4a0fc1cbf8ae472da0fdeefd5a319069b23365c324d20d3544aa76da96dcR146)

docs/user_guide/models_and_metrics/feature_extractor.md
dherrera1911:
```suggestion
plenoptic is compatible with any model written in pytorch, including deep neural networks. In this notebook we show how to use `plenoptic` with models from the deep network zoos {external+torchvision:ref}`TorchVision <models>` and {external+timm:doc}`timm <models>` to recreate some ResNet50 metamers shown in {cite:alp}`Feather2023-model-metam`, figure 2e.
```

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+In the rest of this section, we show how to initialize a plenoptic-compatible model using the weights from either the {external+torchvision:ref}`TorchVision <models>` or {external+timm:doc}`timm <models>` model zoos; their behavior after this section is the same.
```
dherrera1911:
```suggestion
In the rest of this section, we show how to initialize a plenoptic-compatible model using the weights from either the {external+torchvision:ref}`TorchVision <models>` or {external+timm:doc}`timm <models>` model zoos. For that we need to specify three things: 1) The deep net model, 2) the layer(s) to extract, and 3) the image pre-processing to use.
```

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+Next, we ensure that our model is in evaluation mode. Many models, including ResNet50, behave differently when in training and evaluation mode. In plenoptic, models are fixed and so we want the evaluation behavior:
+
+```{literalinclude} ../../scripts/feature_extractor.py
+:dedent:
+:start-at: deepnet.eval()
+:end-at: deepnet.eval()
+```
+
+Next, we need to specify the layer to target. If we look at the ResNet50 metamers in Figure 2e from {cite:alp}`Feather2023-model-metam`, we can see an interesting progression in layers 2 through 4: the layer 2 metamer looks almost identical to the target image, the layer 3 metamer starts to add RGB noise, and the layer 4 is almost completely unidentifiable, looking almost completely like random RGB noise.
```
dherrera1911:
```suggestion
Next, we specify the layer to target. Figure 2e from {cite:alp}`Feather2023-model-metam` shows an interesting progression for ResNet50 metamers: the layer 2 metamer looks almost like the target image, the layer 3 metamer shows some RGB noise, and the layer 4 metamer is almost completely unidentifiable.
```

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+Let's start with `"layer3"`, but note the metamer synthesis procedure in this notebook works with any of `"layer2"`, `"layer3"`, or `"layer4"` (and possibly others, they just haven't been tested).
```
dherrera1911:
```suggestion
Let's start with `"layer3"`, but note that the metamer synthesis procedure in this notebook works with any of `"layer2"`, `"layer3"`, or `"layer4"`, and that you can also specify multiple layers simultaneously.
```

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+You can view possible layer names with {external+torchvision:func}`torchvision.models.feature_extraction.get_graph_node_names`. (For more details on the node naming conventions, please see the {external+torchvision:ref}`About Node Names <about-node-names>` heading in the {external+torchvision:doc}`torchvision documentation <feature_extraction>`.)
+
+```python
+from torchvision.models import feature_extraction
+# this function returns two lists, the first for training mode, the second for eval mode
+feature_extraction.get_graph_node_names(deepnet)[1]
+```
+
+And note that you can specify multiple layers!
```
dherrera1911:
```suggestion
```

dherrera1911: I suggested adding this important detail to the main text.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+:::{admonition} This notebook retrieves cached synthesis results
+:class: warning dropdown
+
+The example metamer shown in this notebook takes about 15 minutes to synthesize on a GPU. Thus, instead of performing synthesis in this notebook, we have cached the result of it online and only download them for investigation.
+
+:::
+
+
+## Initializing the model
```
dherrera1911: Add subsection headers for "Initialize model weights", "Select layer" and "Specify preprocessing" and "Prepare image".

I think this would make the tutorial A LOT more readable.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+```python
+from torchvision.models import feature_extraction
+# this function returns two lists, the first for training mode, the second for eval mode
+feature_extraction.get_graph_node_names(deepnet)[1]
+```
+
+And note that you can specify multiple layers!
+:::
+
+Next, we grab the preprocessing transform from the model. As the [torchvision docs](https://docs.pytorch.org/vision/stable/models.html#using-the-pre-trained-models) explain it (quoting version `0.27`):
```
dherrera1911:
```suggestion
Finally, it is important to specify the preprocessing transform of the model. During neural network training images are transformed before being passed to the network, and the same transformation needs to be applied when using the trained network. As the [torchvision docs](https://docs.pytorch.org/vision/stable/models.html#using-the-pre-trained-models) explain it (quoting version `0.27`):
```

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+> Before using the pre-trained models, one must preprocess the image (resize with right resolution/interpolation, apply inference transforms, rescale the values etc). There is no standard way to do this as it depends on how a given model was trained. It can vary across model families, variants or even weight versions. Using the correct preprocessing method is critical and failing to do so may lead to decreased accuracy or incorrect outputs.
+
+For models trained on ImageNet, this preprocessing consists of two steps: resizing to a height and width of 224 pixels and normalizing the color channels (subtracting means and dividing by standard deviations). Following {cite:alp}`Feather2023-model-metam`, we recommend including the normalization step in the model for metamer synthesis, but handling the image resizing externally. We demonstrate how to do so below.
```
dherrera1911:
```suggestion
For models trained on ImageNet, this preprocessing consists of two steps: resizing to a height and width of 224 pixels and normalizing the color channels (subtracting means and dividing by standard deviations). Following {cite:alp}`Feather2023-model-metam`, we recommend including the normalization step in the model for metamer synthesis, but handling the image resizing externally.

Both torchvision and timm provide the additional model-specific objects implementing the preprocessing transforms. We show below how to use them to create a normalizing transform to pass to {class}`~plenoptic.models.FeatureExtractorModel`.
```

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+Let's grab the normalizing transform and then initialize our plenoptic model:
```
dherrera1911:
```suggestion
```

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+::::{tab-set}
+:::{tab-item} torchvision
+:sync: torchvision
+
+In torchvision, the transform is a single torch Module which we cannot easily subdivide, so we create a separate normalization transform, which we pass to {class}`~plenoptic.models.FeatureExtractorModel`:
```
dherrera1911:
```suggestion
In torchvision, the transform is a single torch Module which we cannot subdivide, that includes resizing, cropping and normalization. We extract the normalization parameters from this torch Module to create a separate standalone normalization transform:
```

dherrera1911: I think it was not clear what things the torch Module includes, that we cannot subdivide.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+
+```python
+ImageClassification(
+    crop_size=[224]
+    resize_size=[256]
+    mean=[0.485, 0.456, 0.406]
+    std=[0.229, 0.224, 0.225]
+    interpolation=InterpolationMode.BILINEAR
+)
+```
```
dherrera1911: Is there a way to make this render more like a printed output? I was very confused because it looked like a class definition.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+print(transform)
+```
+
+```python
+Compose(
+    Resize(size=256, interpolation=bilinear, max_size=None, antialias=True)
+    CenterCrop(size=(224, 224))
+    MaybeToTensor()
+    Normalize(mean=tensor([0.4850, ...]), std=tensor([0.2290, ...]))
+)
```
dherrera1911: Is there a way to make this render more like a printed output? I was very confused because it looked like a class definition.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+```
+
+:::
+::::
+
+Finally, we'll pass our neural network, target layer, and preprocessing transform to plenoptic's {class}`~plenoptic.models.FeatureExtractorModel`:
```
dherrera1911:
```suggestion
Now we can finally create our model by passing the neural network, target layer, and preprocessing transform to plenoptic's {class}`~plenoptic.models.FeatureExtractorModel`:
```

dherrera1911: Just changed because I used "Finally" for the step of specifying the normalization transform above.

docs/user_guide/models_and_metrics/feature_extractor.md
```diff
@@ -0,0 +1,494 @@
...
+:dedent:
+:start-at: def get_category
+:end-at: get_category(img
+```
+
+```python
+guenon
+```
+
+The category, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it.
```
dherrera1911: I suggest some reorganization to impose fewer sidetracks on the reader.

I would move this setup of the classification part to the end, right before where you actually show the classification results.

Additionally, I would move the section "Synthesizing the metamer" before "Understanding the model", so the reader gets the actual main point of the tutorial earlier. I see that the tutorial would show the layer visualization outputs in `po.plot.synthesis_status()` before they are actually explained, but I still think it's best.

In the the section "Understanding the model" I would add two subsections, "Visualizing the features" and something like "Metamer classification".

docs/user_guide/models_and_metrics/feature_extractor.md
dherrera1911: Agree.

src/plenoptic/models/feature_extractor.py
dherrera1911: At least for crop and the case of image synthesis, there is an easy explanation that the part of the image cropped by the model would not be optimized, and the resulting synthesized image would have an optimized center and surrounding noise.

This is a comment in response to https://github.com/plenoptic-org/plenoptic/pull/460#discussion_r3483286016

dherrera1911:
Sharing my opinion about the shape of the default output, whether to use the 2D vector or the dictionary, since it's something that both @emeyer121 and @raffles-xingqi-zhu mentioned.

The main goal of plenoptic is image synthesis. The 2D vector output makes it the easy to directly plug `FeatureExtractorModel()` into the synthesis methods. If the default output was the dictionary, synthesis would require the extra step of transforming the output. So, I think we should keep the 2D output as default.

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+    from torchvision.models import feature_extraction
+except ImportError:
+    feature_extraction = None
+
+from ..plot import display
+
+
+class FeatureExtractorModel(torch.nn.Module):
+    """
+    Return features from model.
```
dherrera1911:
```suggestion
    Return features from deep neural network layers.
```

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+
+    This adapter combines a torch model with a feature extractor and optional transform,
+    allowing us to target the output of one or more particular layers in a deep neural
+    network for use with synthesis objects.
```
dherrera1911: Maybe this could be made clearer by leaning more directly onto the models being DNNs from the model zoos, instead of the more general "torch model". Also, it says that it combines a "feature extractor", but I don't think it will be clear what it means. Do you mean the `torchvision` utility? If so, that's all internal, and can be confusing to the user.

I would go with something like "This adapter takes a DNN model from `torchvision` or `timm`, and returns the activations of the specified layers. It also takes an optional image preprocessing transform to include in the model."

src/plenoptic/models/feature_extractor.py
dherrera1911: There is a type annotation for `transform` to be an `nn,Module`. Should we require that it is an `nn.Module`? Should we add a check?

Also, I was wondering if we want to add some other checks like that `transform` keeps the gradient, the precision, etc. But if it doesn't, those should be caught downstream by `Metamer()` and `MADCompetition()` when trying to use the instantiated object, I think, so it wouldn't be necessary.

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+            return_nodes = [return_nodes]
+        if feature_extraction is None:
+            raise ImportError(
+                "Missing optional dependency torchvision, which is needed for "
+                "FeatureExtractorModel, please install it!"
+            )
+        self.extractor = feature_extraction.create_feature_extractor(
+            model, return_nodes
+        )
+        self.model = model
```
dherrera1911:
```suggestion
```
`self.model` is not used anywhere. While I appreciate it can be convenient to keep the model handy, I think that `model` occupies a lot of extra space in memory. In the code below, I get that the instantiated object takes `97 MB` in memory currently, and only `33 MB` when I comment out this line.

```python
import plenoptic as po
import torchvision

weights = torchvision.models.ResNet50_Weights.IMAGENET1K_V1
deepnet = torchvision.models.resnet50(weights=weights)
deepnet.eval()

transform = weights.transforms()
norm = torchvision.transforms.Normalize(transform.mean, transform.std)

model = po.models.FeatureExtractorModel(deepnet, "layer3", norm)
po.remove_grad(model)

# Check the size in memory of the model. There is no po.utils for getting size
def module_memory_size_bytes(model):
    params = sum(p.numel() * p.element_size() for p in model.parameters())
    buffers = sum(b.numel() * b.element_size() for b in model.buffers())
    return params + buffers

size = module_memory_size_bytes(model)
print(f"{size / 1024**2:.2f} MiB")
```

dherrera1911 (commented):
I think this is a very valuable addition to plenoptic!

I like the scope of the new class, and even though it is a very simple wrapper, I think it is key to make it as easy as possible to use DNNs with plenoptic, and to have dedicated tutorials.

The change I feel most strongly about is adding `DeepNet` or `DNN` to the name of this new class.

Then I suggested many edits to the tutorial to improve organization and readability.

I left a comment saying I support the default output of the class to be the 2D vector, which is most natural for synthesis, the main purpose of `plenoptic`. But I also agree that it's a bit annoying to have to call `convert_to_dict` on the output. I wonder if you could add a function like `model.activation_map(img)` that returns the structured output directly, for convenience. I don't feel to strongly about it though.

Then I point out that you're saving the full DNN in the feature extractor, which seems to take up a bunch of space, but not be used for anything. I think you should not be storing the full DNN.

src/plenoptic/models/feature_extractor.py
dherrera1911: Although maybe it would be good to add a test verifying that there is an informative error message when we instantiate an invalid `FeatureExtractorModel()` (e.g. normalization that removes the gradient) and try to synthesize a metamer.

What do you think?

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+           (<Figure ...>, [<Axes...>, <Axes...>, <Axes...>, <Axes...>])
+
+        Plot on an existing axes object:
+
+        .. plot::
+           :context: close-figs
+
+           >>> fig, axes = plt.subplots(1, 2)
+           >>> model.plot_representation(model.convert_to_dict(model(img)), ax=axes[1])
+           (<Figure ...>, [<Axes...>, <Axes...>, <Axes...>, <Axes...>])
```
sjvenditto, line 643: this plot looks a little funky in the docs

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+            raise ImportError(
+                "Missing optional dependency torchvision, which is needed for "
+                "FeatureExtractorModel, please install it!"
+            )
+        self.extractor = feature_extraction.create_feature_extractor(
+            model, return_nodes
+        )
+        self.model = model
+        if hasattr(model, "training") and model.training:
+            warnings.warn(
```
sjvenditto, line 239: I would add a test to make sure this warning is raised

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+        >>> img = po.process.center_crop(
+        ...     po.data.einstein(False), tv_transform.crop_size[0]
+        ... )
+        >>> representation_dict = model.convert_to_dict(model(img))
+        >>> [(k, v.shape) for k, v in representation_dict.items()]
+        [('layer2', torch.Size([1, 512, 28, 28])),
+         ('layer4', torch.Size([1, 2048, 7, 7]))]
+        """
+        if self._packed_shapes is None or self._out_keys is None:
+            raise ValueError(
```
sjvenditto, line 418: I would add a test to make sure this error is raised

src/plenoptic/models/feature_extractor.py
```diff
@@ -0,0 +1,637 @@
...
+
+            # Average representation across channels
+            avg_channel_rep = v.mean(dim=1, keepdim=True)
+            # Average representation across additional dimensions (probably space)
+            per_channel_rep = v.mean(dim=tuple(np.arange(2, v.ndim)))
+            while per_channel_rep.ndim < 3:
+                per_channel_rep = per_channel_rep.unsqueeze(0)
+
+            if avg_channel_rep.ndim == 3:
+                height_ratios = [1, 1]
```
sjvenditto, line 677: This case is not being tested

sjvenditto (commented):
> How's the name FeatureExtractorModel? I was originally calling this TorchVisionAdapter, which is both too narrow and not actually accurate. I want to highlight that this is making use of the feature extractor and so anything that works with that will work with this, but I worry it's a little too abstract (hence why I called the docs page "using deep nets with plenoptic", but the feature extractor will work with non-deep net models as well, so I didn't want to call it DeepNetAdapter or something)

Echoing the other's comments, I would go with just `FeatureExtractor` or `DeepNetFeatureExtractor` if you want to be more specific. Even though you say it works on non-deep net models, unless you can think of a specific example of an alternative model, I would just lean into the DNN use case.

> The new notebook contains several code-blocks where I want to show the output (e.g., what the torchvision transform looks like).

Another reorganizing possibility: Split the tutorial into two halves, "Initializing the model" and "Understanding the model & Synthesizing the metamer" (you could also do three separate sections, but the last two aren't very long by themselves). In the first half, which is the only part that differs between torchvision and timm, have two separate versions like "Initializing a torchvision model" and "Initializing a timm model". So in all, there would be 3 total sections under a "Using Deep Neural Networks with plenoptic" landing page. The downside is that the use would have to download two separate notebooks to run the entire tutorial.

Alternatively (echoing what was already said), you can have the "main" tutorial focus only on torchvision, and then have a second, shorter tutorial for timm that is only the model setup. 

Also, instead of just choosing a single layer for the tutorial, why not just use all three together? It would also highlight better the function of `convert_to_dict`, and `plot_representation` would also show all three together. If you did end up doing this, I think you can leave the layer tabs in the metamer section as it is.

> In order to get the "summary stats" for the different layers' metamers at the end of the notebook, I'm actually generating a png with matplotlib to visualize, the same way I'm getting the synthesis_status figure. I don't think there's a clean alternative to that?

I think this looks fine

> Because this notebook is such a tangled mess, I don't provide links to let people download/run the notebook at the top. instead, I provide a link to download the accompanying script. Does that make senes? should I improve the docs in that script?

I think I agree with Emily that it'll probably make more sense to clean it up to also be a `.ipynb` download (with my suggestions above). Can you not have any tabbed code blocks at all in the notebook? I do like it for the layer plots/output in the last section.

---

For other comments, I noticed that one of the example plots for `plot_representation` (specifically plotting on existing axes) has a weird size ratio in the resulting plot in the documentation (I put a comment in the code for this)

Also for untested errors/warnings/cases, I added inline comments for places I think are important to test

billbrod:
Thanks for all your feedback everyone! I think I'm going to switch the model name to `DeepNetFeatures` (deep nets being the most likely use case for this model) and I'm going to split the docs page in two: one minimal notebook using layer3 and torchvision that folks can download focusing on basic usability (maybe with a separate admonition to show how to initialize the timm model, under "Models and Metrics"), and a separate one that shows the results for the other layers and discusses them as reproducing the results in Feather et al. 2023 (under "Reproducing results"). I'll ping you all again once I've made the changes

src/plenoptic/models/feature_extractor.py
billbrod: I want this to be a single block so that users can copy and paste it entirely from the docs. I've added some notes to refer people to the notebook for more details.

src/plenoptic/models/feature_extractor.py
billbrod: Daniel's point is exactly right. 

Also, intended use case of this model is to use for synthesis. If all you want is to visualize activations in a specific layer, they can use torchvision's feature extraction machinery directly.

billbrod: I've also added something to the deep net notebook about this

src/plenoptic/models/feature_extractor.py
billbrod: yes, the features are just concatenated together. I've updated the text to explain this a bit more

src/plenoptic/models/feature_extractor.py
billbrod: Good catch. this is because  matplotlib only runs the sections of this here that are within the `.. plot::` directive. I'll fix this.

src/plenoptic/models/feature_extractor.py
billbrod: by code blocks, you mean in different functions? no, each function has to be independent. both for how doctests work and also because when using it, a user may read one of these docstrings independent of the others (e.g., by pulling up the docstring for `plot_representation` in jupyter).

they are not necessarily looking at the page on the docs site

src/plenoptic/models/feature_extractor.py
billbrod: Users are not expected to use this function to make animations themselves. They should see `synthesis_animate` to do so. I've updated docstring to try to make this clear

docs/scripts/feature_extractor.py
billbrod: `po.to_numpy` calls `.cpu` for the user, so it's not necessary

docs/user_guide/models_and_metrics/feature_extractor.md
billbrod: this is a warning, not an error. did it show up as an error? how were you running this?

I'll add a bit of explanation about the warning

billbrod: actually, I'll suppress the warning

docs/user_guide/models_and_metrics/feature_extractor.md
billbrod: none of the other similar plots have error bars, so I'm not going to here. I'll add an example in the doctest showing how users can do it after.

docs/user_guide/models_and_metrics/feature_extractor.md
billbrod: I do not have an intuition here. The paper doesn't discuss the absolute loss value, or if any of their success metrics differ across layers.

The paper uses a relative loss (dividing the Euclidean distance by the L2-norm of the representation from the layer), and it's possible that doing that would normalize this value.

docs/user_guide/models_and_metrics/feature_extractor.md
billbrod: Good point, I've updated this wording

docs/user_guide/models_and_metrics/feature_extractor.md
billbrod: unfortunately, no. That would require getting all the images they used. I've tweaked the wording a bit

docs/user_guide/models_and_metrics/feature_extractor.md
billbrod: I disagree -- I refer to the cache this and other notebooks (Portilla-Simoncelli, Berardino replication) to be clear why we're not calling synthesize.

I've now split up the notebook into two sections, one for "how to use deep nets" and one for "replicating Feather et al, 2023", and the cache warning is only on the Feather notebook (since the "how to use deep nets" one does call synthesize). I think this works, let me know what you think

src/plenoptic/models/feature_extractor.py
billbrod: good catch on this! yeah there's no particular reason I was carrying this around

src/plenoptic/models/feature_extractor.py
billbrod: huh, I dunno. I don't want to check the type (actually, it just needs to be a callable that accepts and returns a tensor, so I've updated the type annotation to include that)

all the other things you mentioned would be caught by the synthesis objects themselves, so I don't think it's necessary here? for the vast majority of use-cases, people will be doing like what's in the examples, which will behave fine

billbrod: and also, those checks are only necessary if you want to use the model for synthesis, so I think that's when they should be caught.

so I'm deciding against doing any checks here. if people end up having issues with this (passing illegal transforms and then not being able to debug it), we'll revisit this.

src/plenoptic/models/feature_extractor.py
billbrod: I don't want to make it sound like it only works with deep nets, so I tried something intermediate between what you said and what I had, let me know what you think.

(I agree about feature extractor being too in the weeds)

docs/user_guide/models_and_metrics/feature_extractor.md
billbrod: I agree on everything except for the relative location of the "Synthesizing" and "Understanding" sections. hopefully some of the other changes I've made have helped streamline this, let me know what you think

billbrod:
Okay, thanks everyone again for all your reviews! I believe I have addressed everything. I've gone through and responded to many of your specific comments. If I just marked it resolved without responding, that means "I agree and have made those changes"

Some big picture:
- I renamed the model to `DeepNetFeatures`. I agree that, since deep nets are the intended use case, that should be in the name. I tried to thread the needle in the docstring by not explicitly saying it only worked for deep nets, unsure if that was successful
- As Daniel mentioned, the output is a 2d vector because the goal of this is to simplify the use of deep nets *with plenoptic's synthesis methods.* Therefore, they should work with them directly -- if `forward` returned a dictionary then you would need to write an additional class wrapping around this to make it compatible with the synthesis methods. if a user does not want to use a deep net with the synthesis methods but just look at intermediate activations, they should use torchvision's feature extractor directly.
    - the name `convert_to_dict` (and the fact that it requires the output of `model.forward`, instead of an image directly) is because that's how it's handled by `PortillaSimoncelli`, so I want to be consistent. also, it makes the function more helpful to me as an internal helper function (e.g., when plotting)
- I broke up the one tutorial page into two separate notebooks, one of which just demonstrates how to use `DeepNetFeatures` (and synthesizes a single metamer), the other of which loads in the cached results of the layer2, 3, and 4 metamers and displays them. This definitely simplifies the flow (removes the issue about the separate script entirely), and makes them easier to understand. Let me know what you think.

I would like your input on how I describe the differences between our approach and that of the paper: the synthesis approach they use is fundamentally different from ours, based on modifying the [robustness](https://github.com/MadryLab/robustness) package's machinery for generating adversarial examples. Therefore, it's a bit complicated to compare directly. In particular, in the paper they say "Each step of gradient descent was constrained to have a maximum L2 norm of η, where η was initialized at 1 and was dropped by a factor of 0.5 after every 3,000 iterations." I'm not sure how to map that onto Adam's learning rate. They also don't discuss using any specific optimization algorithm. I don't want to go too much into the weeds in the notebook, but I do think it should be clear that our algorithm is different -- any suggestions?

@raffles-xingqi-zhu you asked about why is timm not described as an "optional dependency", but torchvision is: it's because torchvision must be installed in order for `DeepNetFeatures` to run, regardless of where the input deep net comes from. timm is only necessary if want to get your deep net from there. I recognize that's a subtle distinction.

I tried to address people's comments in different commits, unsure if that is helpful or not, but:
- @emeyer121  90dadb0b
- @raffles-xingqi-zhu : d04cfecc
- @dherrera1911: 7c36bda0
- @sjvenditto: aabd3835

billbrod:
reposting the docs link because it got kind of buried: https://docs.plenoptic.org/docs/pulls/460/

billbrod:
@emeyer121 I tried to respond to your comment about this but it got stuck at "pending", so I'm posting here:

> Oh I guess here you use eval() after extracting the features? If this also works that's fine but probably good to be consistent or at least make notes about it. In the user guide it is explicitly mentioned that .eval() is needed but not in the docs, so probably good to specify.

We can call eval before or after initializing DeepNetFeatures, but doing it before will avoid a warning. I've changed these all to be consistent and added more explanation.

billbrod:
Everything is currently failing because of an issue with OSF. I will retrigger the builds once that gets resolved.

EDIT: looks like it's fixed now, re-triggering everything

billbrod:
realized I forgot to respond to some of @sjvenditto points:

> Alternatively (echoing what was already said), you can have the "main" tutorial focus only on torchvision, and then have a second, shorter tutorial for timm that is only the model setup.

I now just have an admonition at the bottom of the "prepare model" section showing how to do this in timm, which I think works -- what do you think?

> Also, instead of just choosing a single layer for the tutorial, why not just use all three together? It would also highlight better the function of convert_to_dict, and plot_representation would also show all three together. If you did end up doing this, I think you can leave the layer tabs in the metamer section as it is.

Matching all three layer representations at the same time is not the same as matching each of them separately -- it should be the same as matching the first of the layers, since if the layer2 representation matches, then the layer3 must match and thus the layer4 must match. Instead, when doing a layer4 metamer, we want the layer2 and layer3 representations to differ, and only the last one to match. I think the solution I found for this (with splitting the notebooks up) works, let me know what you think.

sjvenditto (approved):
Looks good! I think the way you divided the notebooks works well, and moving the timm code to the dropdown admonitions helps with the flow. I think with the table in the new notebook you're making it clear that there are differences in the fitting procedures between plenoptic and Feather et al 2023, I don't think you need to go into more detail than this, but I don't have strong opinions.

The only potential issue I see is the synthesis animation is only showing the first and last iteration? This line of code in the first notebook: `po.plot.synthesis_animate(met, figsize=(15, 4.5))`. Is that intended? Otherwise I'm approving this because I think everything else looks good!

billbrod:
> The only potential issue I see is the synthesis animation is only showing the first and last iteration? This line of code in the first notebook: `po.plot.synthesis_animate(met, figsize=(15, 4.5))`. Is that intended? Otherwise I'm approving this because I think everything else looks good!

Gah good catch, I did not mean to do that.

docs/user_guide/models_and_metrics/feature_extractor.md
raffles-xingqi-zhu: Thanks for adding them. Now you've added them I feel like it's a bit extra. Up to you whether you want to keep it. I don't have strong opinions on this, especially if it's better to be consistent with everything else in the documentation.

docs/user_guide/models_and_metrics/deep_nets.md
```diff
@@ -0,0 +1,356 @@
...
+:::
+
+Let's call {func}`~plenoptic.plot.synthesis_status` to visualize the synthesis status:
+
+```{code-cell} ipython3
+po.plot.synthesis_status(met, figsize=(15, 4.5));
+```
+
+In the above plots, we can see the metamer in the leftmost subplot, the loss over synthesis iterations in the middle, and the representation error on the right:
+- Our metamers match the results discussed earlier in this notebook:  the layer 2 metamer looks almost identical to the target image, the layer 3 metamer starts to add RGB noise, and the layer 4 starts to lose much of the image structure in random RGB noise.
```
raffles-xingqi-zhu: There's no layer 2 and 4 now. Should remove.

docs/user_guide/models_and_metrics/deep_nets.md
```diff
@@ -0,0 +1,356 @@
...
+- We can see that the optimization performed reasonably well: the loss decreased gradually over synthesis. If you were using these stimuli in an experiment (especially for `"layer4"`), it may be worth continuing a bit more to get the loss even lower, but these demonstrate the point.
+- The representation error plot has the same structure as the {func}`~plenoptic.models.DeepNetFeatures.plot_representation` plot above. We can see that, while there's variation across both channels and space, there's not an obvious outlier whose error we have been unable to constrain.
+
+And we can animate the above figure over synthesis iterations as well, to see the metamer take shape:
+
+```{code-cell} ipython3
+po.plot.synthesis_animate(met, figsize=(15, 4.5))
```
raffles-xingqi-zhu, line 342: The animation jumps from beginning to end without playing on my machine.

raffles-xingqi-zhu (commented):
Thank you for making all the changes. Everything looks really good now! I just have two small comments and I think it's good to go!

docs/user_guide/models_and_metrics/deep_nets.md
```diff
@@ -0,0 +1,351 @@
...
+You may also be interested in [](feather2023), where we create model metamers for several ResNet50 intermediate layers, reproducing some of the results from {cite:alp}`Feather2023-model-metam`.
+
+```{code-cell} ipython3
+import matplotlib.pyplot as plt
+import numpy as np
+import torch
+
+import plenoptic as po
+
+# this notebook uses torchvision, which is an optional dependency. if this import fails,
```
emeyer121: this is super minor but I think if you shorten these comment lines you won't have a scroll bar at the bottom

docs/user_guide/reproduce/feather2023.md
```diff
@@ -0,0 +1,230 @@
...
+    scheduler=scheduler,
+    scheduler_kwargs=scheduler_kwargs
+)
+# by setting stop_iters_to_check=max_iter, we ensure it keeps going through
+# all 12k iterations
+met.synthesize(max_iter=12000, stop_iters_to_check=12000)
+```
+:::
+
+Note that the synthesis procedure we used differs somewhat with that of {cite:alp}`Feather2023-model-metam`. The following table summarizes these differences:
```
emeyer121: might appreciate one more sentence about why you chose to do things differently otherwise might bring up questions that distract people from the main point

emeyer121 (approved):
Added two additional minor comments but anything else I saw was already addressed. I think everything is reading very clearly now and should be easy for people to use!

docs/user_guide/models_and_metrics/feature_extractor.md
dherrera1911: I didn't realize we weren't calling synthesize. On my quick read, I thought you were adding the code for the synthesize call, but were not actually running it in the docs generation, and that that's what you were explaining to the user.

I think that, ideally, the notebook should do that: Have a sham synthesize call, but in the background actually load cached results, and not mention it. But even if you would consider this idea, I think it should be done in a future PR.

But for this PR, shouldn't this admonition be in the **Initialize and load Metamer objects** section, which is where the relevant code appears?

dherrera1911: I didn't realize we weren't calling synthesize, I though you were just doing the cache loading in the background. My bad.

I agree then with keeping the original admonition. Ideally, though, I think that the notebook should just have the code to do the synthesis. It could be a sham synthesis call, with the cache loading happening in the background, or have these docs synthesize the metamer but not be built on every push. But even if you would consider this idea, it should be done in a future PR.

But agreeing to keep the admonition for this PR, shouldn't it be in the **Initialize and load Metamer objects** section, which is where the relevant code appears?

docs/user_guide/models_and_metrics/deep_nets.md
```diff
@@ -0,0 +1,352 @@
...
+def get_category(image):
+    # Get probabilities of each image category
+    image_cat = torch.nn.functional.softmax(deepnet(norm(image)), dim=1)
+    # Convert to 1d numpy array, so we can use as index in
+    # imagenet_categories above.
+    image_cat = po.to_numpy(image_cat.squeeze())
+    return imagenet_categories[image_cat.argmax()]
+
+
+print(get_category(img))
```
dherrera1911:
```suggestion
print(f"ResNet50 predicted class: {get_category(img)}")
```

docs/user_guide/models_and_metrics/feature_extractor.md
dherrera1911: I think it is much better overall after the edits.

I personally would still prefer "Synthesis" first, to get the user to the use case earlier. But I don't feel strongly about this, and the new version is good as is too.

src/plenoptic/models/feature_extractor.py
dherrera1911: The new version looks good, no further comments.

src/plenoptic/models/feature_extractor.py
dherrera1911: I agree. Let's not complicate ourselves with unnecessary checks here.

dherrera1911:
Hi Billy. All looks good to me. I left two very small suggestions:
- Add a clearer print text of the class in one tutorial
- Suggest moving the admonition explaining cached results to where the relevant code is (I unresolved a conversation for this)

Then I shared two opinions, but none of them important for this PR:
- That ideally the code in the tutorial should be how to synthesize the metamer, and if cacheing is used, it should be in the backend
- That I still think "Synthesis" would be better before "Understanding" in the first tutorial, but I don't feel strongly about it.

But I think it's good to merge as is already.

Thanks for inviting me to review this PR :)

dherrera1911 (approved):

billbrod:
For some reason, Github is not letting me respond to threads (it just says "Pending"), so I'm going to do it in the main comments.

For the point about synthesize vs. cache:

I do not like the idea of doing a sham synthesis and then calling load in the background, for a couple reasons:
- I want users to see that they can load these outputs themselves, if they're interested.
- Coming up with a generic solution to this is hard: how do I know when someone wants to load vs. synthesize? I don't want users to call synthesize on their own machine and actually have it load something from the OSF project I use for docs/testing files
- A specific solution is also hard -- I'm not sure what a clean way would be to intercept that call (if it happens in the docs), and I don't like the idea that synthesize behaves differently depending on where it's called (if it happens in the library)

The actual metamer gets synthesized as part of the tests (`test_uploaded_files.py`, `test_resnet_macaque_metamer`), which synthesizes the metamer and checks it against the cached version. So I do know that it's up-to-date.

But I do agree about moving the admonition, I'll do that.

billbrod:
I disagree about synthesis being moved before understand (since I'd have to explain the plot structure anyway), so I'm keeping this as is.

Thanks for your review!

PR #462, opened 2026-07-08
When animating color videos/syntheses, only raise a single warning.

billbrod:

In workshop materials for [CSHL](https://workshops.plenoptic.org/workshops/CSHL-vision-course-2026/branch/main/exercises/4_torchvision.html), noticed that the matplotlib RGB clipping warning gets printed on every frame where it comes up during animations (probably because of how we update the image). In the link above, this meant that when animating a deepnet metamer synthesis, we got that warning printed 100 times.

Turns out, this isn't warned using `warnings.warn` but `logging.warn`, which is used because (per [python docs](https://docs.python.org/3/howto/logging.html)):

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
>
> A logger’s warning() method if there is nothing the client application can do about the situation, but the event should still be noted

Because of this, I can't catch the warning and do simple filtering using `warnings.filterwarnings`

Thus, instead we do the clipping ourselves in the animate functions, raising a single specific warning to say how many frames were clipped. This affects `animshow` and `synthesis_animate`, but in different ways.

Updated API pages for [synthesis_animate](https://docs.plenoptic.org/docs/pulls/462/api/generated/plenoptic.plot.synthesis_animate.html), [animshow](https://docs.plenoptic.org/docs/pulls/462/api/generated/plenoptic.plot.animshow.html)

**Describe changes**

- `animshow`: for each video, determine how many frames would be cliped and raise a separate warning for each (specifying video index and number of frames), then clip manually. no other functionality added, if users want to avoid clipping, they need to do so external to this function
- `synthesis_animate`:
    - warns similarly (though there's always only a single video). 
    - adds `process_image` argument, analogous to `synthesis_imshow`. set to none by default, in which case we clip color images (and don't do anything to grayscale). users can manually set this to avoid the clipping in other ways (e.g., with plenoptic's included linear-rescaling function)
- adds tests for that new functionality
- adds new examples to `synthesis_animate` docstring showing clipping and use of `process_image`
- adds simple helper function to determine whether a single image is RGB (for `synthesis_imshow` and `synthesis_animate`)
- adds helper function to determine how many frames would be clipped and format warning message.
- remove `warnings.filterwarnings` at top of Eigendistortion notebook, since it doesn't do anything.
- in `test_display.py`, figured out how to convert matplotlib's logged warnings into an error with the `no_logs_warnings` fixture. this is only applied to the tests where I'm interested in specifically looking into the clipping issue.

**Checklist**

### DIFF
--- a/docs/user_guide/synthesis/Eigendistortions.md
+++ b/docs/user_guide/synthesis/Eigendistortions.md
@@ -27,11 +27,6 @@ warnings.filterwarnings(
     message="plenoptic's methods have mostly been tested on",
     category=UserWarning,
 )
-
-warnings.filterwarnings(
-    "ignore",
-    message="Clipping input data to the valid range",
-)
 ```
 
 :::{admonition} Run this notebook yourself!
--- a/src/plenoptic/plot/display.py
+++ b/src/plenoptic/plot/display.py
@@ -509,7 +509,6 @@ def _clip_frames_warnings_msg(
     dims = " ".join(dims)
     n_frames_min = einops.reduce(video, f"{dims} -> t", "min") < 0
     n_frames_max = einops.reduce(video, f"{dims} -> t", "max") > 1
-    print(n_frames_min, n_frames_max)
     n_frames = einops.pack([n_frames_min, n_frames_max], "* t")[0].any(0).sum()
     if n_frames > 0:
         warning_msg = (
--- a/src/plenoptic/plot/display.py
+++ b/src/plenoptic/plot/display.py
@@ -463,6 +463,64 @@ def imshow(
     )
 
 
+def _clip_frames_warnings_msg(
+    video: torch.Tensor,
+    time_dim: int = 2,
+    video_name: str = "video",
+    to_avoid_txt: str = "",
+) -> str:
+    """
+    Get warning message for clipped frames in RGB(A) videos.
+
+    Computes the number of frames with values outside [0, 1] and returns formatted
+    string describing how many clipped.
+
+    If no pixels would be clipped, empty string is returned.
+
+    Parameters
+    ----------
+    video
+        5d Tensor of shape to check. Note that this doesn't check ``video`` is RGB(A),
+        that must be done externally.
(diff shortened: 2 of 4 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/462

raffles-xingqi-zhu:
The doc strings are clear to me. One thing I'm not sure about is whether animshow() also clips grayscale images.

billbrod:
> The doc strings are clear to me. One thing I'm not sure about is whether animshow() also clips grayscale images.

It does not. In the [docstring](https://docs.plenoptic.org/docs/pulls/462/api/generated/plenoptic.plot.animshow.html) section on `as_rgb` I say

> `as_rgb`: Whether to consider the channels as encoding RGB(A) values. If True, we attempt to plot the image in color, so your tensor must have 3 (or 4 if you want the alpha channel) elements in the channel dimension. If False, we plot each channel as a separate grayscale image. Note that if True, clipping may result, see Warns section below for details.

where that warns section says:

> UserWarning – If as_rgb and video contains values outside of [0, 1], we will clip video so all values lie within [0, 1]. (This is because of the requirements of [matplotlib.pyplot.imshow](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html#matplotlib.pyplot.imshow).) The warning will say how many frames of which video were clipped. To avoid clipping, you should remap video to [0, 1] before calling this function.

Do you have a suggestion for how to make that more salient? Maybe modify the last two lines to say: "If False, we plot each channel as a separate grayscale image. Note that if True, clipping may result, see Warns section below for details (no clipping is applied if `as_rgb=False`)."?

raffles-xingqi-zhu:
I think the way it's written is fine. I might have just missed it when I first read it. It would be great to eventually add a doc test that shows the behaviour (as_rgb True vs False).

billbrod:
Great idea, I added a note to #237

PR #463, opened 2026-07-09
Fix issue where MADCompetition attributes were not tab-completing in ipython/jupyter

billbrod:

I recently noticed that tab completion was not working for `po.MADCompetition` (either the class or instances) in ipython or jupyter. Turns out, it's an issue with [jedi](https://github.com/davidhalter/jedi/issues/1990), which underlies that tab completion. They should fix it, but until they do, the solution on our end is to make sure we have no `Literal` return annotations. This does that and then adds a new check to pre-commit to ensure we don't add any more of them.

From the user end, they could also run `%config IPCompleter.use_jedi = False` in ipython/jupyter, which fixes it. (found in https://github.com/ipython/ipython/issues/14412)

**Describe changes**

- Change return annotation of `MADCompetition.minmax` to `str` from `Literal['min','max']`.
- add new check script, `linting/check_return_literal.py`, which checks whether the substring `-> Literal` is present in `.py`.
- adds that script to pre-commit, only running on python files within `src/`

**Checklist**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -98,6 +98,12 @@ repos:
       files: ^docs/
       entry: linting/fix_tutorial_semicolons.py
       additional_dependencies: ["jupytext"]
+    - id: return-literal-annotation
+      language: python
+      name: Ensure we have no Literal return annotations
+      types_or: ["python"]
+      files: ^src/
+      entry: linting/check_return_literal.py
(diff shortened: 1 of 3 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/463

PR #464, opened 2026-07-09
Updates `__all__`, add `validate_penalty` to init

billbrod:

Realized that `validate_penalty` and `groupwise_relative_l2_norm_factory` weren't showing up when tab-completing, and turns out they weren't in `__all__`. This adds them, along with some checks to ensure this doesn't happen again.

Also adds `validate_penalty` to the initialization of Metamer and MADCompetition (@dherrera1911 had pointed out we forgot to put them there). I then changed the initialization of the test image to avoid using `torch.rand`, since that was essentially incrementing the random seed by an additional step, leading to breaking backwards compatibility.

(@dherrera1911 we had discussed using `torch.eye` but that wouldn't work if the user wanted to synthesize a tensor of shape `1,1,200`, for example. So instead I use `torch.ones` and then set some of the values to 0.)

**Describe changes**

- Adds `groupwise_relative_l2_norm_factory` and `validate_penalty` to their respective module's `__all__`
 - Adds `linting/check_api_change.py` to ensure that all public functions/classes are included in `_api_change.py` (follows same logic of earlier `linting/check_apidocs.py`)
 - Adds that to pre-commit
 - Adds test to `test_api.py` which checks that all functions/classes specified in `_api_change.py` are found in `__dir__` of a plenoptic module (the reverse was already tested)
 - Adds `validate_penalty` to `Metamer.__init__` and `MADCompetition.__init__`
     - removes conditions in `test_mad.test_load_penalty_change` which looked for loading behavior when shape or dtype were changed by penalty -- those are disallowed by `validate_penalty`
 - Adds RUF027 linting rule to check for unformatted f-strings
 - Adds `validate_penalty` to `_api_change.py`, along with the unchanged functions/attributes
 - Change `validate_penalty` to avoid using `torch.rand`
 - Fix f-string in `validate_penalty`

**Checklist**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -104,6 +104,12 @@ repos:
       types_or: ["python"]
       files: ^src/
       entry: linting/check_return_literal.py
+    - id: check-api-docs
+      language: system
+      name: Ensure all public functions / classes are included in _api_change.py
+      entry: python linting/check_apidocs.py
+      always_run: true
+      pass_filenames: false
 
 - repo: https://github.com/codespell-project/codespell
   # Configuration for codespell is in pyproject.toml
--- a/linting/check_api_change.py
+++ b/linting/check_api_change.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+
+# confirm that the _api_change.py file specifies all public functions
+
+import pathlib
+import re
+import sys
+
+# These are the files whose contents we don't want in the api docs. that's __init__.py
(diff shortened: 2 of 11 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/464

billbrod:
@dherrera1911 I'm pinging you so you this, but I'm going to go ahead and merge this -- we can do a follow up PR with additional tests.

PR #466, opened 2026-07-17
Adds "How to use Penalties" page to the documentation

billbrod:

This PR adds a new page to the docs, describing how to use penalties. It tries to be a gentle introduction to `penalty_function`, using a simple model (`Gaussian`) and penalties (range constraints). It ended up being a bit complex.

The goal of this page is to present an introduction to the software requirements for penalties (very simple), as well as the optimization-related ones (a bit tricky), so that users can hopefully write their own interesting penalty functions.

This is part of #428.

- The main feedback I'd like is on clarity -- is the page understandable? Do you feel like you could try playing around with your own penalties after reading it?
- Also, where should this page live? I created a new "Intermediate Synthesis" folder for this, but I don't think that's the right name there. I also don't think this belongs under Advanced, because I think it's more common than those examples (hence, intermediate).
    - I'm planning on adding two other penalty-related notebooks, so I could create a section just for penalties, but that feels a little overly narrow. 

**Describe changes**

- Adds "How to use penalties" notebook, in new section of User Guide
- Removes `Metamer regularization` section from metamer notebook and updates links to pint to new notebook

**Checklist**


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/466

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,287 @@
...
+Download the executed notebook: **{nb-download}`how_to_penalty.ipynb`**!
+
+Run it in your browser: **{binder}`how_to_penalty.ipynb`**!
+
+:::
+
+(how-to-penalty)=
+# How to use custom penalty functions
+
+Plenoptic's {class}`~plenoptic.Metamer` and {class}`~plenoptic.MADCompetition` objects allow users to specify a {attr}`~plenoptic.Metamer.penalty_function` at initialization. This penalty function is multiplied by {attr}`~plenoptic.Metamer.penalty_lambda` (also set at initialization) and added to the objective function of the object.
```
BalzaniEdoardo: What is `penalty_lambda`? Is that a scalar, another func?

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,287 @@
...
+    return -penalty
+
+
+print(f"All ones -- very negative penalty: {custom_penalty(torch.ones(10))}")
+print(f"All 0.5 -- zero penalty: {custom_penalty(0.5 * torch.ones(10))}")
+rand_penalty = custom_penalty(torch.rand(10))
+print(f"Random between 0 and 1 -- intermediate penalty: {rand_penalty}")
+```
+
+We are trying to minimize the value of `custom_penalty`. Thus, a large negative number if better than 0 (or any positive number). So of our three example inputs above, our penalty prefers the first. Let's see what this looks like with metamer synthesis:
```
BalzaniEdoardo:
```suggestion
We are trying to minimize the value of `custom_penalty`. Thus, a large negative number is better than 0 (or any positive number). So of our three example inputs above, our penalty prefers the first. Let's see what this looks like with metamer synthesis:
```

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,287 @@
...
+create_metamer_figure(met)
+```
+
+In the above figure, the top row shows: the target image, the metamer, and a histogram of pixel values in those two images. The bottom rows shows the model representation of the top two rows. We can see that the two representations on the bottom match and thus the two images on the top are metamers.
+
+By looking at the histogram (and the "range" section in the title of the images on the top row), we can see that the pixel values in our metamer vary between 0 and 1 and, since it was initialized with a patch of uniformly-distributed noise, those pixels are widely distributed between those values.
+
+## Construct the custom penalty
+
+All penalty functions in plenoptic must be callables that take the synthesized image as input and return some scalar penalty. We can write a custom penalty that makes use of the {func}`~plenoptic.regularize.penalize_range` function to penalize pixel values that fall outside some narrower range instead:
```
BalzaniEdoardo, line 158: Probably, before discussing setting a custom penalty, I would remind people what is the loss: loss = distance(f(original) - f(new)) + l * penalty(new)

In this way you can mention, the penalty acts like a bias to the loss. This has two implication:

- If at the minimum the penalty is non-zero, then the new image is not technically a metamer for the original model, which would have an unpenalized loss
- The minimum balances perceptual distance and the penalty.

I think this is worth point out early, as a short conceptual discussion about penalties.  so that people have the framework in mind clearly. 

Then you can introduce the custom penalty, and after that it is easier to talk about the fact that if your penalty is not bounded (and not minimal at zero), it may bias the metamer too much.

BalzaniEdoardo: Also, the plot of the two components of the loss will make more sense if the equation of the penalty is spelled out.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+The penalty value is a large negative value and still going lower, whereas our loss has actually increased. Since our optimization problem is minimizing the weighted sum of these two numbers, and loss is bounded below by 0, the penalty has become much more important to the optimization problem, and it can effectively ignore the metamer loss.
+
+In situations like this, the solution is to apply some function to the penalty so that its range is remapped from $(-\infty, 0]$ to $[0, \infty)$ (or $[0, x]$, for some finite $x$). Fortunately, the [exponential function](https://en.wikipedia.org/wiki/Exponential_function) will do this for us, remapping $(-\infty, 0]$ to $[0, 1]$:
+
+```{code-cell} ipython3
+def custom_penalty(image):
+    penalty = po.regularize.penalize_range(image, allowed_range=(0.4, 0.6))
+    return torch.exp(-penalty)
```
BalzaniEdoardo, line 274: This is true becasue penalty is positive, otherwise for a remapping as such, you'd need `exp(-p**2), maybe it is worth saying explicitly, because exp(-x) doesn't do the remapping on its own

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,287 @@
...
+```{code-cell} ipython3
+# Create custom_penalty function, that penalizes pixels outside of [0.1, 0.9] range
+def custom_penalty(image):
+    penalty = po.regularize.penalize_range(image, allowed_range=(0.1, 0.9))
+    return penalty
+```
+
+In the following, we can see that our `custom_penalty` accepts a single tensor and returns a scalar, quadratic penalty on any values it contains outside of 0.1 to 0.9:
+
+```{code-cell} ipython3
```
BalzaniEdoardo, line 176: plotting the penalty in 1D (using a tensor (n,1,1,1) ) may be a nice way to clarify what the range penalty does.

BalzaniEdoardo: And will definitely help to plot the penalties later when you construct a custom one

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,287 @@
...
+    imgs = [met.image, met.model(met.image), met.model(met.metamer)]
+    titles = ["Target image", "Representation of target", "Representation of metamer"]
+    for ax, im, t in zip(image_axes, imgs, titles):
+        po.plot.imshow(im, ax=ax, title=t, zoom=1, vrange=(0, 1))
+        ax.xaxis.set_visible(False)
+        ax.yaxis.set_visible(False)
+    axes[-1].set_visible(False)
+
+
+create_metamer_figure(met)
```
BalzaniEdoardo: adding suptitles would help when people run the code. i am generating a bunch of images but I lost track which is which

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,287 @@
...
+print(custom_penalty(torch.rand_like(img)).item())
+```
+
+One final modification: when synthesizing images, it is generally desirable for all pixel values to lie within $[0, 1]$, so they can be displayed (which is why this is plenoptic's default behavior). To ensure that, let's modify our penalty one last time, summing together the two penalties before returning them.
+
+```{code-cell} ipython3
+def custom_penalty(image):
+    outside_penalty = po.regularize.penalize_range(image, allowed_range=(0.4, 0.6))
+    inside_penalty = po.regularize.penalize_range(image, allowed_range=(0, 1))
+    return torch.exp(-outside_penalty / 1000) + inside_penalty
```
BalzaniEdoardo: As I was playing around with the penalty, I tried this Gaussian that works pretty well on restricting the domain to (0, 0.4) union (0.6, 1):


```python
def custom_gauss(image):
    outside_penalty = (0.05*torch.exp(-1*(image-.5)**2/0.001)).mean()
    inside_penalty = po.regularize.penalize_range(image, allowed_range=(0, 1))
    return  outside_penalty + inside_penalty
```

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,287 @@
...
+
+To see why this is, let's use {func}`plenoptic.plot.synthesis_loss` to plot our metamer loss alongside our penalty value:
+
+```{code-cell} ipython3
+ax = po.plot.synthesis_loss(met, plot_penalties=True)
+# convert this to symlog scale, so we can see negative values.
+ax["loss"].set_yscale("symlog")
+```
+
+The penalty value is a large negative value and still going lower, whereas our loss has actually increased. Since our optimization problem is minimizing the weighted sum of these two numbers, and loss is bounded below by 0, the penalty has become much more important to the optimization problem, and it can effectively ignore the metamer loss.
```
raffles-xingqi-zhu: You probably mean the metamer loss is bounded below by 0 here right?

docs/user_guide/intermediate_synthesis/how_to_penalty.md
raffles-xingqi-zhu: This issue is potentially minor but I was a bit confused when I first read `outside_penalty = po.regularize.penalize_range(image, allowed_range=(0.4, 0.6))` because at this stage it is still a penalty for "inside". It isn't until we apply the negative sign in front that it becomes "outside". 

I also find the convention for inside and outside not intuitive. I would've chosen the name "outside penalty" for the second line here. Because it is a penalty for being outside not inside.

BalzaniEdoardo (changes requested):
The description on how to set a penalty up is very clear, while the description of how to construct it in a meaningful way is a bit more convoluted than it should be.

Adding an equation showing the penalized loss would probably be useful: you can foreshadow the fact that the behavior of the penalized model depends on the balance of the terms. The balance can be tuned with the scaling term.

## First custom example

I would start with a good penalty directly: you can state what ideally we would like, and you can even show the step function which has zero penalty in the interval of interest and non-zero elsewhere. Then you should say, well, this doesn't work because it is not differentiable (actually all gradients of this function are zero everywhere and undefined at the step). You can use my exp function or anything that has that profile as a smooth approx of the step. 

Once you have that you can run a successful syntehsis. 

## Balancing the metamer loss and the penalty

Play with the scaling showing what happens when the loss is not balanced, and show the final plot of the loss functions on the 3 different cases as you are already doing, which is very instructive.

## Summarizing

I think that flow it's maybe easier to follow: the first part explains what is the ideal penalty we would want and why we can only approximate it. The second part is focused on the treadeoffs. 


In general, plotting the penalty as function of pixel value would make sense, to explain why we are constructing the penalty function in a specific way. I think creating a `ones_like` with the same shape as the image is the best way to go because the range penalty sums across pixel and therefore is pixel size dependent. 

Let me know what you think.

BalzaniEdoardo:
Extra comment, from the plot of the penalty, you can tell the user: look the slope of the penalty tells us where the gradient is moving the parameter: the (0,1) penalty pushes large  pixel values inwards (toward (0,1)) the other term pushes values from (0.4,0.6) outwards.

raffles-xingqi-zhu (commented):
Very important and interesting topic for plenoptic. Overall I find the notebook clear. The examples are well thought out and the text is easy to follow. Here are my thoughts to your questions: 

- One thing that left me desiring for more about is what are some other typical penalty functions used in optimization and more specifically metamer synthesis. The pixel range example here is great for pedagogical purposes (and constraining to [0,1] is left on by default) but I was unsure of its practical implications. I know there're other cool penalty functions because I read the CCN abstract but the general audience may not. The notebook could benefit from adding a short paragraph of the benefit of adding additional penalty functions in real life usage. 

- Deciding where to place the notebook(s) will ultimately depend on the other two related notebooks in the works. If we believe knowing how penalty works is a prerequisite to understanding metamer synthesis I think this notebook should belong to "Synthesis method introductions". Alternatively if it is leaning more on the hacky side, it should probably belong to "Advanced usage" or its own section. As it currently stands I'm leaning towards the latter.

raffles-xingqi-zhu:
Extra comment: adding the metamer synthesis objective with the penalty function and lambda scaling at the beginning of the notebook could be helpful.

billbrod:
From both @ershook and @dherrera1911 exploration of using penalties for the CCN poster, they found better results when doing synthesis in a two-stage process, synthesizing without a penalty first, and then re-initializing with the output of that process (and including the penalty). So probably add something about that to the docs.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
billbrod: I replaced this with a generic `penalty_one` and `penalty_two` because I couldn't come up with a better name.

billbrod: I'm going to keep using the one I have, but I mentioned this as an alternative

billbrod:
Thanks @raffles-xingqi-zhu and @BalzaniEdoardo for your comments! I think I've addressed all your comments. Biggest changes are that I added a section at the top laying out the objective function and giving a high-level overview, and I switched the print statements showing penalty values to a plot. I think both of those made the notebook much clearer!

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+Run it in your browser: **{binder}`how_to_penalty.ipynb`**!
+
+:::
+
+(how-to-penalty)=
+# How to use custom penalty functions
+
+Plenoptic's {class}`~plenoptic.Metamer` and {class}`~plenoptic.MADCompetition` objects allow users to specify a {attr}`~plenoptic.Metamer.penalty_function` at initialization. In this notebook, we'll walk through the process of creating a custom penalty function and explain best practices. We will do so for {class}`~plenoptic.Metamer`, but a similar procedure can be applied to {class}`~plenoptic.MADCompetition`. We will refer to the stimuli as "images", but the same procedure will work with other stimulus types as well.
+
+When synthesizing a metamer, we are searching for a new image whose model representation matches that of a target image. plenoptic performs this procedure as an iterative optimization function, minimizing the objective function:
```
raffles-xingqi-zhu:
```suggestion
When synthesizing a metamer, we are searching for a new image whose model representation matches that of a target image. Plenoptic performs this procedure as an iterative optimization function, minimizing the objective function:
```

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+
+$$ L(M(x), M(\hat{x})) + \lambda P(\hat{x}) $$
+
+where:
+
+- $M$ is {attr}`~plenoptic.Metamer.model`, the model whose representation we are trying to match.
+- $x$ is {attr}`~plenoptic.Metamer.image`, the image whose model representation we are trying to match.
+- $\hat{x}$ is {attr}`~plenoptic.Metamer.metamer`, the image we are synthesizing.
+- $L$ is {attr}`~plenoptic.Metamer.loss_function`, a function which accepts two tensors and returns a scalar, which defines how different two representation tensors are (by default, {func}`plenoptic.loss.mse`).
+- $\lambda$ is {attr}`~plenoptic.Metamer.penalty_lambda`, a scalar that controls the balance between the metamer loss and penalty terms (by default, `0.1`).
```
raffles-xingqi-zhu: I think it may be better to introduce the `penalty_function` before introducing `penalty_lambda`, so that you write in the `penalty_lambda` line: "controls the balance between `loss_function` and `penalty_function`.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+
+If you have a model operating in a reasonable manner on images, it is likely that we can find metamers that have minimal values for both the metamer loss and the penalty constraining pixel values to lie between 0 and 1. However, we cannot assume that is the case for more complicated penalty functions. As we will show in this notebook, it may be possible to reformulate your penalty term to make the problem more tractable and it is likely that trying different {attr}`~plenoptic.Metamer.penalty_lambda` will allow you to find images with low values for both components, but you may still find yourself in a situation where you have to choose which term to prioritize. If so, it is important to remember: for an image to be a model metamer, its metamer loss **must** be zero (see [](good-enough) and [](feather-synthesis-success) for discussion on what "zero" means in practice). If your scientific reasoning depends on your images being model metamers, this is non-negotiable. However, depending on your scientific question, it is likely that your penalty doesn't need to be met exactly, and you are thus encouraged to prioritize metamer loss.
```
raffles-xingqi-zhu: Reading through the notebook, this paragraph feels a bit too long and detailed to be helpful. By the time I got to the end I lost track of where I was and what this notebook is about. Maybe consider putting this to a later section?

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+
+In this notebook, we'll show how to change this penalty to use a different range of allowed values, exploring how the penalty can be used to encourage and discourage different properties in the resulting synthesized metamer, and demonstrating the importance of both properties of the function itself as well as the selection of the {attr}`~plenoptic.Metamer.penalty_lambda` value in successful metamer synthesis.
+
+:::{admonition} What other penalties can I use?
+
+This notebook demonstrates the technical requirements for {attr}`~plenoptic.Metamer.penalty_function`: it must be a differentiable function that accept a single tensor (the metamer-in-progress) and return a single scalar. **Which** function to use depends on your scientific question and what you're interested in. The other notebooks in this section demonstrate some additional uses of penalty functions beyond controlling the range of allowed pixel values.
```
raffles-xingqi-zhu: The MAD competition adversarial example notebook would be good to link here, even though it currently doesn't belong to the same section.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+    else:
+        ylim = ax.get_ylim()
+    ax.vlines(x_vlines, *ylim, "k", "dashed")
+    ax.set(ylim=ylim, xlabel="Pixel value", ylabel="Penalty value")
+    return fig
+```
+
+## Prepare model and image for synthesis
+
+In this notebook, we'll use a simple {class}`~plenoptic.models.Gaussian` model. This model just convolves a Gaussian kernel across the entire image. It is thus a low-pass model, only caring about low frequencies and disregarding high ones.
```
raffles-xingqi-zhu:
```suggestion
In this notebook, we'll use a simple {class}`~plenoptic.models.Gaussian` model. This model just convolves a Gaussian kernel across the entire image. It is thus a low-pass model, only caring about low frequencies and disregarding high frequencies.
```

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+img = po.data.einstein().to(DEVICE).to(torch.float64)
+met = po.Metamer(img, model)
+met.synthesize(MAX_ITER, stop_criterion=1e-16)
+
+# This plotting function was created in a hidden cell at the top of the notebook, expand
+# it if you're interested in the details.
+create_metamer_figure(met, "Default penalty");
+```
+
+In the above figure, the top row shows: the target image, the metamer, and a histogram of pixel values in those two images. The bottom rows shows the model representation of the top two rows. We can see that the two representations on the bottom match and thus the two images on the top are metamers.
```
raffles-xingqi-zhu:
```suggestion
In the above figure, the top row shows: the target image, the metamer, and a histogram of pixel values in those two images. The bottom row shows the model representation of the top two rows. We can see that the two representations on the bottom match and thus the two images on the top are metamers.
```

raffles-xingqi-zhu:
```suggestion
In the above figure, the top row shows: the target image, the metamer, and a histogram of pixel values in those two images. The bottom rows shows the model representation of the top two images. We can see that the two representations on the bottom match and thus the two images on the top are metamers.
```

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+
+Now that we understand its behavior, we can pass this `custom_penalty` to the {class}`~plenoptic.Metamer` class at initialization and call the {meth}`~plenoptic.Metamer.synthesize` method in the same manner as before:
+
+:::{admonition} Multi-stage synthesis
+:class: dropdown note
+
+In the examples in this notebook, we will always call {func}`~plenoptic.Metamer.synthesize` once, simultaneously optimizing for the metamer loss and our custom penalty. For more complex penalty functions, however, we have generally found better performance by first calling {func}`~plenoptic.Metamer.synthesize` without the custom penalty and then using the resulting un-penalized metamer to initialize a second synthesis procedure, like so:
```
raffles-xingqi-zhu: Consider breaking up the second sentence as it's currently a bit long.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+```
+
+The figure above has the same structure as earlier, and we can see that the new metamer's pixel values all fall within this narrower range of $[0.1, 0.9]$. By looking at the second row, we can see that this image is still a metamer: the representation still looks identical to that of the target image.
+
+:::{admonition} Why not clamp?
+:class: dropdown
+
+If you are familiar with the machine learning literature, you may be wondering, "why not just use {external+torch:func}`torch.clamp` to constrain the pixel values?" This would accomplish the same goal as this example, but in our experience the quadratic range penalty works more reliably across different models. (In particular, in {cite:alp}`Broderick2025-foveat-metam`, using {func}`~plenoptic.regularize.penalize_range` led to decreased metamer loss compared to clamping.)
+
+Additionally, and more importantly, {external+torch:func}`torch.clamp` is not nearly as flexible as using a penalty function! While it would allow you to keep the pixel values within ranges like $[0, 1]$ or $[0.1, 0.9]$, it would not easily support the behavior we will see in the rest of this notebook, let alone other, more complex penalty functions.
```
raffles-xingqi-zhu: Having read the notebook, it's still a bit unclear to me what "the behaviour" is that cannot be achieved through clamping.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+
+## Best practices
+
+The example shown above is relatively simple. However, for optimization to behave well, it is important that:
+1. penalties have some minimum value (i.e., they cannot go to negative infinity).
+2. this minimum value corresponds to the desired property (e.g., all pixels lie within the desired range).
+
+Furthermore, it is generally convenient to have this minimum value be zero.
+
+If the conditions above are violated, then the penalty can overwhelm the loss in the objective function, preventing a metamer from being found.
```
raffles-xingqi-zhu: This only happens if the first condition is violated?

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+
+Returning to our discussion at the [beginning of the notebook](how-to-penalty), by design, our penalty function is biasing the optimization procedure. Here, however, because the magnitude of the penalty is so much larger than that of the metamer loss, it is biasing the optimization procedure too much, preventing us from finding a solution whose metamer loss is small.
+
+Let's revisit the plot of the penalty as a function of pixel values:
+
+```{code-cell} ipython3
+plot_penalty(img, custom_penalty, [0.4, 0.6], (0, 1));
+```
+
+If we compare this plot to the [one for the penalty in the previous section](well-behaved-range-penalty), we can see an important difference: while both penalties go towards an absolute value of infinity outside of some range, the well-behaved penalty in the last section does that for the values it is **discouraging**, whereas this poorly-behaved penalty is doing that for the values it is **encouraging**.
```
raffles-xingqi-zhu: Maybe it's just me but I'm having trouble interpreting the meanings of "discouraging" vs "encouraging".

raffles-xingqi-zhu: Specifically it doesn't give me more intuition beyond what you already explained: the fact that the new penalty can be negative makes it overwhelm the objective, whereas the old penalty is bounded below by 0.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+
+To fix this, we need to change the plot above so that it looks that of the well-behaved penalty. That is, we must remap its range from $(-\infty, 0]$, where 0 corresponds to the values we want to avoid, to $[0, \infty$ (or $[0, x]$, for some finite $x$), where 0 corresponds to the values we want to encourage. Fortunately, the [exponential function](https://en.wikipedia.org/wiki/Exponential_function) will do this for us, remapping $(-\infty, 0]$ to $[0, 1]$:
```
raffles-xingqi-zhu:
```suggestion
To fix this, we need to change the plot above so that it looks that of the well-behaved penalty. That is, we must remap its range from $(-\infty, 0]$, where 0 corresponds to the values we want to avoid, to $[0, \infty)$ (or $[0, x]$, for some finite $x$), where 0 corresponds to the values we want to encourage. Fortunately, the [exponential function](https://en.wikipedia.org/wiki/Exponential_function) will do this for us, remapping $(-\infty, 0]$ to $[0, 1]$:
```

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+met = po.Metamer(img, model, penalty_function=custom_penalty)
+met.synthesize(MAX_ITER, stop_criterion=1e-16)
+
+create_metamer_figure(
+    met, "Encourage pixels outside [0.4, 0.6] -- good penalty, not quite a metamer"
+);
+```
+
+And now we can see that the model metamer pixel values mostly lie outside the range $[0.4, 0.6]$ and within $[0, 1]$. We can see a couple of other interesting results of our penalty function:
+- We have no pixels outside of $[0, 1]$, but we do have some pixels in $[0.4, 0.6]$. If we look at our penalty plot above, we can see that the magnitude of the penalty is much larger for values outside $[0, 1]$ than for those within $[0.4, 0.6]$, so it makes sense that we would see an asymmetry in the resulting metamer. If we do not wish this to be the case, we could multiply the components of the penalty (e.g., `penalty_two`) by some scalar to increase or decrease their relative importance.
```
raffles-xingqi-zhu: I find the asymmetry really curious. Since much higher penalty is given to values beyond 0 and 1 than values between 0.4 and 0.6, I would have expected the asymmetry to go the other way. Am I missing something here?

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+- More subtly, it looks like the pixel values are being pushed to a bimodal distribution, clumping near the values 0 and 1, instead of being uniformly distributed within $[0, 0.4]$ and $[0.6, 1]$. Again, returning to the penalty plot above can help explain this: we see that the penalty decreases very gradually as you move away from $0.4$ or $0.6$, and so it only approaches a value and a slope of zero as you near $0.2$ / $0.8$. If we do not wish to see this bimodal distribution, we could instead pick a penalty whose value decreases more rapidly within the allowed range (for example, using `plenoptic.regularize.penalize_range(image, allowed_range=(0.45, 0.65)` or a different function entirely, such as a narrow Gaussian centered at 0.5).
```
raffles-xingqi-zhu: Wouldn't using `plenoptic.regularize.penalize_range(image, allowed_range=(0.45, 0.65)` do the opposite of what we want (keeping pixels OUT of 0.4 and 0.6)? Am I missing something here?

docs/user_guide/intermediate_synthesis/how_to_penalty.md
```diff
@@ -0,0 +1,411 @@
...
+fig, axes = plt.subplots(1, 3, figsize=(15, 5), sharey=True)
+po.plot.synthesis_loss(met, plot_penalties=True, ax=axes[0])
+axes[0].set_title("Lambda just right")
+po.plot.synthesis_loss(met_small, plot_penalties=True, ax=axes[1])
+axes[1].set_title("Lambda too small")
+po.plot.synthesis_loss(met_large, plot_penalties=True, ax=axes[2])
+axes[2].set_title("Lambda too large");
+```
+
+Remember from the beginning of the notebook that our objective function balances the contribution of these components. Finding the right lambda value so that neither overwhelms the other is key!
```
raffles-xingqi-zhu:
```suggestion
Remember from the [beginning of the notebook](how-to-penalty) that our objective function balances the contribution of these components. Finding the right lambda value so that neither overwhelms the other is key!
```

raffles-xingqi-zhu (changes requested):
Thank you for addressing my previous points. Overall I find the notebook clear and well-presented. In addition to my inline comments, my main concern right now is the notebook is quite long (longest I've come across so far!). Also, as someone who had some experience playing with penalty in plenoptic, it required quite a bit of mental effort to follow the notebook. For this reason I'm not sure if this would still belong to "intermediate usage". Let me know what you think!

docs/user_guide/intermediate_synthesis/how_to_penalty.md
billbrod: MAD competition adversarial example or Metamer? Because the MAD one doesn't use `penalty_function`, but the metamer one does

docs/user_guide/intermediate_synthesis/how_to_penalty.md
billbrod: tried to explain what I meant a bit more: "constraining the pixel values to two disjoint ranges"

docs/user_guide/intermediate_synthesis/how_to_penalty.md
billbrod: true. if the second condition is violated, the optimization's solution won't have the property you want. tried to update the text to clarify

docs/user_guide/intermediate_synthesis/how_to_penalty.md
billbrod: tried to clarify the text here

docs/user_guide/intermediate_synthesis/how_to_penalty.md
billbrod: I'm trying to explain a subtle difference related to condition 2 at the beginning of this section: the desired property should correspond to the finite minimum. 

maybe that's unnecessary. I've removed this section entirely, let me know if it reads better without it

docs/user_guide/intermediate_synthesis/how_to_penalty.md
billbrod: Hmm. I put this in to try and address some of Edoardo's comments about foreshadowing the issue of bias and the possibility that you won't be able to find a good solution. I've tried to tighten and break up this paragraph, let me know if it's any better.

billbrod: maybe it would work better as an admonition?

billbrod:
> Thank you for addressing my previous points. Overall I find the notebook clear and well-presented. In addition to my inline comments, my main concern right now is the notebook is quite long (longest I've come across so far!). Also, as someone who had some experience playing with penalty in plenoptic, it required quite a bit of mental effort to follow the notebook. For this reason I'm not sure if this would still belong to "intermediate usage". Let me know what you think!

Thanks Raffles! I've tried to address your comments, leaving un-resolved the ones that I think need revisiting.

For me, the distinction between intermediate and advanced is more about how likely a user is to use its contents -- and I think many more users will use penalty functions than write custom plotting code or extending the synthesis objects (which are the focus of the advanced pages). Maybe intermediate/advanced is the wrong terminology then? Any other ideas?

In terms of "requiring quite a bit of mental effort", it is a reasonably involved notebook. There are three points being made in it:
- Technical requirements of `penalty_function` (single tensor in, single scalar out) and how to change it (set `penalty_function` arg during initialization)
- How to design a well-behaving penalty function
- How to pick `penalty_lambda`

Maybe the notebook can be split up into sub-notebooks, each of which makes a separate point? Similar to the PS model, which used to be a single giant notebook, and then was broken up into many smaller ones. I could have an index which gives everything before the first code cell, and then links to a separate notebook for each of the three points above. (Each of which would start with the figure showing the metamer with default penalty function.) that would hopefully reduce the mental load.

I'm now realizing there's also a fourth point hidden in the "how to design a well-behaving penalty function" section: how to combine penalties. That's relatively simple (sum them together), but may also be worth calling out separately. Thoughts?

docs/user_guide/intermediate_synthesis/how_to_penalty.md
raffles-xingqi-zhu: I like the way it reads now. I don't think the admonition is necessary.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
raffles-xingqi-zhu: metamer notebook, sorry.

docs/user_guide/intermediate_synthesis/how_to_penalty.md
raffles-xingqi-zhu: okay referring back helped.

raffles-xingqi-zhu:
Thank you for the changes. I think dividing by "intermediate" and "advanced" is fine after hearing more of your logic. For dividing up the notebook I think that would be helpful. Let's see what other people think.

BalzaniEdoardo:
This LGTM. I don't think you should expand on composing the penalties, because the penalization machinery is already so generic. People can pass custom functions therefore they have total freedom there. 

I don't know if it is worth mentioning but: the fact that you have some residual values in (0.4, 0.6) is most likely because the custom penalty is almost flat in the range, so the gradients are not pushing enough the pixel values. If you look at the large vs small lambda case: with `penalty_lambda=1e6` the pixel values are concentrated at the borders where the exponential is basically decayed numerically to zero (for pixel val in (0, 0.1) U (0.9,1) more or less). While in the`penalty_lambda=0.025` case, there are more values inside (0.4, 0.6) and a broader distribution.

BalzaniEdoardo (approved):
LGTM, the discussion is complicated but clear. The complexity is something that a user should appreciate before playing around with the penalties, and there is no way around it. I think the note does a good job to provide context and intuition.

billbrod:
Discussed briefly with Edoardo, and we decided that, instead of breaking the notebook into sub-notebooks, make it clear that there are three important points here, that all need to be understood before using penalty functions (with a little intro at the top of the notebook). And then make the headers for each section clear, with more info there as well. Basically, adding a bit more of the meta-info.

The risk of splitting it up into sub-notebooks it that people will only read the first one and think they're good to go, but they really do need to understand all of it.

PR #467, opened 2026-07-22
Mad adversarial example

raffles-xingqi-zhu:

This PR adds a notebook showing how to generate adversarial examples of Deep Neural Networks using MAD Competition.

Not that I'm aware of. 

- Is the notebook clear? Anywhere that needs more explanation/examples? 
- Should this notebook belong to "Reproduce examples from the literature"? Currently I don't think it would belong to any of other tabs in the User Guide. I'll work on another notebook showing how to generate adversarial examples using Metamer so these two would belong together as a pair. 
- Initially I explored L2 norm as the optimized metric but found the adversarial examples led to wrong but unconfident classification (not one-hot). To encourage one-hot behaviour, Billy and I decided to add a penalty to the metric. I initially wrote the documentation for this version of the metric but found the approach a bit convoluted and arbitrary (due to two additional hyper parameters of the penalty term). So I decided to try a closely related metric: the MSE. In the hyper parameter search it worked as well as the metric using L2 norm with the penalty term. Occam's razor made me go with this choice as the optimized metric. 

**Describe changes**

- Full example of working through using MAD competition to generate adversarial examples. 

**Checklist**


raffles-xingqi-zhu:
A reminder for Billy to check Jupytext config in VS code

billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/467

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+    format_name: myst
+    format_version: 0.13
+    jupytext_version: 1.19.3
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+<!-- #region -->
```
billbrod:
```suggestion
```

I think this must be from the VS code notebook editing you were doing, it's unnecessary

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+    axes[i, 1].set_title("Categories")
+    axes[i, 0].xaxis.set_visible(False)
+    axes[i, 0].yaxis.set_visible(False)
+    axes[i, 1].text(
+        1, 0.5, f"Likely categories:\n- {likely_cats}", transform=axes[i, 1].transAxes
+    )
+```
+
+This notebook demonstrates how to generate adversarial examples using the {class}`~plenoptic.MADCompetition` class. We encourage you to experiment with different images and hyperparameters to generate other adversarial examples yourself!
+<!-- #endregion -->
```
billbrod:
```suggestion
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+:::
+
+(adversarial_examples)=
+# Generate adversarial examples using MAD Competition
+
+:::{warning}
+This notebook requires the optional dependency `torchvision`, which can be installed with `pip`.
+:::
+
+In this notebook we demonstrate how we can use the {class}`~plenoptic.MADCompetition` class to synthesize adversarial examples. Adversarial examples are tiny perturbations to an image that causes Deep Neural Networks to misclasify ([Goodfellow et al., 2015](https://arxiv.org/abs/1412.6572)). MAD competition is used to falsify metrics/models of human perception by generating a pair of images that have the same value for the reference metric but extremal values (highest and lowest) for the optimized metric and testing human sensitivity to this pair of images. In this notebook we demonstrate a different use of the {class}`~plenoptic.MADCompetition` class and show how its underlying machinery can be readily used to generate adversarial examples of Deep Neural Networks. At a high level, this is achieved by defining a reference metric in pixel space (the value of which we want to be low) an optimized metric in model representation space (the value of which we want to be high).
```
billbrod:
```suggestion
In this notebook we demonstrate how we can use the {class}`~plenoptic.MADCompetition` class to synthesize adversarial examples. Adversarial examples are tiny perturbations to an image that causes Deep Neural Networks to misclassify ([Goodfellow et al., 2015](https://arxiv.org/abs/1412.6572)). MAD competition was developed to compare image quality metrics by generating a pair of images that have the same value for the reference metric but extremal values (highest and lowest) for the optimized metric {cite:alp}`Wang2008-maxim-differ`. These images were then used as stimuli in psychophysics experiments, to demonstrate which metric best aligned with human perception.

Here, we demonstrate a different use of the {class}`~plenoptic.MADCompetition` class and show how its underlying machinery can be readily used to generate adversarial examples of Deep Neural Networks. At a high level, this is achieved by defining a reference metric in pixel space (the value of which we want to be low) and an optimized metric in model representation space (the value of which we want to be high).
```

Do we want "model representation space" here to be something like "distance in model classification" instead?

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+# so that relative sizes of axes created by po.plot.imshow and others look right
+plt.rcParams["figure.dpi"] = 72
+
+# set seed for reproducibility
+po.set_seed(0)
+```
+
+## Prepare model and image for synthesis
+
+In this section, we initialize a plenoptic-compatible model using the weights from {external+torchvision:ref}`TorchVision <models>`. We use the standard ImageNet-trained ResNet50 as our network to attack. You may be also interested in checking out [](deep_nets) for details of choosing layer and preprocessing of the input image, and using models from {external+timm:doc}`timm <models>`.
```
billbrod:
```suggestion
In this section, we initialize a plenoptic-compatible model using the weights from {external+torchvision:ref}`TorchVision <models>`. We use the standard ImageNet-trained ResNet50 as our network to attack. 

:::{seealso} Deep Nets in plenoptic
Check out [](deep_nets) for more information, including how to choose different layers and preprocessing details, as well as how to use models from {external+timm:doc}`timm <models>`.
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+```
+
+## Visualizing classification of the clean image
+First let us extract all the ImageNet categories:
+
+```{code-cell} ipython3
+imagenet_categories = np.asarray(weights.meta["categories"])
+```
+
+Let us define two helper functions. `convert_logits_to_probs` converts logits to probabilities that sum to 1. For an input image, `get_category` returns the vector containing the category probabilities and name of the category with the highest probability.
```
billbrod:
```suggestion
Let us define two helper functions:
- `convert_to_probs` converts the activation in the final fully-connected layer to probabilities that sum to 1. It gets used in the `get_category`, below.
- `get_category` accepts a single image and returns both a vector containing the category probabilities and name of the category with the highest probability.
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+
+```{code-cell} ipython3
+def convert_logits_to_probs(logits):
```
billbrod:
```suggestion
def convert_to_probs(logits):
```

don't think the fact that this is logits originally is important (and we don't explain what that means)

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+    return torch.nn.functional.softmax(logits, dim=1).squeeze()
+
+
+def get_category(image):
+    category_probs = convert_logits_to_probs(deepnet(norm(image))).detach().cpu()
```
billbrod:
```suggestion
    category_probs = convert_to_probs(deepnet(norm(image))).detach().cpu()
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+    category = imagenet_categories[category_probs.argmax()]
+    return category_probs, category
+```
+
+ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The category, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is also highly confident in its classification (probability of ~0.8)
```
billbrod:
```suggestion
ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The category of our initial image, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is also highly confident in its classification (probability of ~0.8)
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+
+```{code-cell} ipython3
+category_probs, category = get_category(img)
+po.plot.stem_plot(category_probs, title=category);
+```
+
+## Define optimized and reference metric
+
+To qualify as an adversarial example, the image must satisfy two requirements: (1) the perturbation in image space is small and (2) the model outputting an incorrect classification with high confidence ([Goodfellow et al., 2015](https://arxiv.org/abs/1412.6572)). Conveniently, we already have these two ingredients built into the MAD competition framework. More concretely, if we define the reference metric in pixel space and ask it to not change or minimally change, and define the optimized metric in representation space, in which we make the image representation as far away from the original as possible, we would be able to meet the two requirements and accomplish our goal. For the reference metric, we use the the simple mean-squared-error (MSE).
```
billbrod:
```suggestion
To qualify as an adversarial example, the image must satisfy two requirements: (1) the perturbation in image space is small and (2) the model outputs an incorrect classification with high confidence ([Goodfellow et al., 2015](https://arxiv.org/abs/1412.6572)). Conveniently, we already have these two ingredients built into the MAD competition framework. More concretely, we should be able to meet these two requirements if we define:

1. The reference metric as a distance in pixel space and set it to a low value.
2. The optimized metric as a distance in representation space, so that we synthesize an image whose representation is as far away from the original as possible.

For the reference metric, we use mean-squared-error (MSE):
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+
+```{code-cell} ipython3
+def reference_metric(x, y):
+    return po.metric.mse(x, y).mean()  # .mean() averages across the RGB channels
+```
+
+For the optimized metric, we use the MSE of output logits in the last layer of the network.
```
billbrod:
```suggestion
For the optimized metric, we use the MSE on the output of the last layer of the network.
```

billbrod: We could add an admonition around here saying something like "Why not compute MSE on the probabilities?" with the explanation of what we came up. It's a little in the weeds, so it should be a dropdown. what do you think?

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+```{code-cell} ipython3
+mad = po.MADCompetition(
+    img, optimized_metric, reference_metric, "max", metric_tradeoff_lambda=1e10
+)
+```
+
+:::{admonition} How does {attr}`~plenoptic.MADCompetition.metric_tradeoff_lambda` affect the adversarial image
+:class: dropdown hint
+
+We conducted a hyperparameter search and found increasing {attr}`~plenoptic.MADCompetition.metric_tradeoff_lambda` led to decrease in both {attr}`~plenoptic.MADCompetition.optimized_metric_loss` and {attr}`~plenoptic.MADCompetition.reference_metric_loss`. This makes intuitive sense as a harsher penalty to keep reference metric loss the same will inevitably come at the cost of the optimized metric loss not increasing (the opposite of what we want). In practise, it would require experimenting with various choices to find the appropriate {attr}`~plenoptic.MADCompetition.metric_tradeoff_lambda`
```
billbrod:
```suggestion
We conducted a hyperparameter search and found increasing {attr}`~plenoptic.MADCompetition.metric_tradeoff_lambda` led to decrease in both {attr}`~plenoptic.MADCompetition.optimized_metric_loss` and {attr}`~plenoptic.MADCompetition.reference_metric_loss`. This makes intuitive sense as a larger penalty value increases the importance of keeping the reference metric constant, which inevitably prevents the optimized metric from increasing. 

If you are applying this procedure to new images or new image classification models, you will almost certainly need to experiment to find the appropriate {attr}`~plenoptic.MADCompetition.metric_tradeoff_lambda`
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+
+:::
+
+We set the initial noise to be a small value so the initial image is closer to the solution that we want: an image with small deviations in the pixel values from the original but produces large changes in the representation. We also set a custom learning rate value of 0.001 even though you should generally obtain good results with higher or lower learning rates provided the synthesis runs for long enough.
```
billbrod:
```suggestion
We set the initial noise to be a small value so that our synthesized image also has small deviations in the pixel values from the original. We also decreased the learning rate from the default, as this resulted in better solutions in our experiments.
```

That's not necessarily true -- high enough learning rates can get stuck in weird places. (if you start high and then use a scheduler to gradually decrease the learning rate, you can often still find good solutions.)

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+
+```{code-cell} ipython3
+mad.setup(initial_noise=0.001, optimizer_kwargs={"lr": 0.001})
+```
+
+Running the synthesis generates an image that looks just like the original image but we see the optimized metric loss has increased significantly. Even though the reference metric loss measured in pixel space has also increased a little bit, it is not nearly as big as the change in the representation space.
```
billbrod:
```suggestion
Running the synthesis generates an image that looks just like the original image but we see the optimized metric loss has increased significantly. Even though the reference metric loss measured in pixel space has also increased, it remains a small value.
```

because in relative terms, they look roughly equivalent. and the relevant thing is that it's still small in absolute terms.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+
+```{code-cell} ipython3
+imgs = [img, mad.initial_image, mad.mad_image]
+mse = [po.metric.mse(img, i) for i in imgs]
+titles = ["Original image", "Initial image", "Adversarial image"]
+diffs = [(i + 1) / 2 for i in [img - img, mad.initial_image - img, mad.mad_image - img]]
+titles.extend([f"MSE={m.mean().item():.2e}" for m in mse])
+imgs.extend(diffs)
+po.plot.imshow(imgs, as_rgb=True, title=titles, col_wrap=3);
+```
```
billbrod, line 262: I would rearrange this so that the original image is in the bottom left corner (and there's nothing in the top left), so that the bottom middle and bottom right show the difference between the row and column (and then update the text above)

billbrod: let me know if that explanation isn't clear

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+Running the synthesis generates an image that looks just like the original image but we see the optimized metric loss has increased significantly. Even though the reference metric loss measured in pixel space has also increased a little bit, it is not nearly as big as the change in the representation space.
+
+```{code-cell} ipython3
+mad.synthesize(1000)
+po.plot.synthesis_status(mad);
+```
+
+## Visualizing the adversarial image
+
+Let us compare how the synthesized image compares to the original and initial images. In the bottow row the original image is subtracted to visualize the changes in pixels. There is no change in the left image. In the middle image we see some hardly noticeable noise. In the last image we see a low amount of pixel noise.
```
billbrod:
```suggestion
Let us compare how the synthesized image compares to the original and initial images. In the bottow row the original image is subtracted to visualize the changes in pixels. There is no change in the left image. In the middle image we see some hardly noticeable noise. In the last image we see a small amount of random pixel noise, distributed across the image.
```

I can barely see any of it -- what if we reduce the range? and then have the text above point out the limited range

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+imgs = [img, mad.initial_image, mad.mad_image]
+mse = [po.metric.mse(img, i) for i in imgs]
+titles = ["Original image", "Initial image", "Adversarial image"]
+diffs = [(i + 1) / 2 for i in [img - img, mad.initial_image - img, mad.mad_image - img]]
+titles.extend([f"MSE={m.mean().item():.2e}" for m in mse])
+imgs.extend(diffs)
+po.plot.imshow(imgs, as_rgb=True, title=titles, col_wrap=3);
+```
+
+We can also visualize the difference in each color channel for the initial image (top row) and adversarial image (bottom row).
```
billbrod: use titles to show which row is which, something like: "Initial - Original (Red channel)", etc.

billbrod: and for these, maybe make them all have the same vrange? use `vrange='auto0'`, see what that looks like

docs/user_guide/reproduce/adversarial_examples_mad.md
billbrod: I also think we should foreshadow that the category has changed, something like: "the fact that the optimized metric has significantly increased shows that the model thinks this is a very different category than our initial image, which we'll show in a later section."

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+    po.plot.stem_plot(category_probs, ax=axes[i, 1], ylim=False)
+    axes[i, 1].set_title("Categories")
+    axes[i, 0].xaxis.set_visible(False)
+    axes[i, 0].yaxis.set_visible(False)
+    axes[i, 1].text(
+        1, 0.5, f"Likely categories:\n- {likely_cats}", transform=axes[i, 1].transAxes
+    )
+```
+
+This notebook demonstrates how to generate adversarial examples using the {class}`~plenoptic.MADCompetition` class. We encourage you to experiment with different images and hyperparameters to generate other adversarial examples yourself!
```
billbrod: and image classifier networks!

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+```
+
+We can also visualize the difference in each color channel for the initial image (top row) and adversarial image (bottom row).
+
+```{code-cell} ipython3
+channelwise_diffs = [mad.initial_image - img, mad.mad_image - img]
+po.plot.imshow(channelwise_diffs, col_wrap=3);
+```
+
+Finally let us visualize the category probabilities of the original, initial, and advesarial images using stem plots. We see that the network thinks the synthesized image is a burrito with almost 100% certainty! We have successfully generated an adversarial example of the network.
```
billbrod: the model thinks it's a dog. (I assume it did burrito when you tried this on the cluster, but Jenkins will give slightly different results). Try setting a different seed

The surefire way to make sure that the category you write in the text matches that of the model is to use [glue](https://myst-nb.readthedocs.io/en/latest/render/glue.html). I haven't done this yet, but it might be worth it.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+
+```{code-cell} ipython3
+fig, axes = plt.subplots(3, 2, figsize=(10, 20))
```
billbrod: change figsize so that the height is closer to that of the individual images -- we've got some wasted white space on the heights.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,241 @@
...
+for i, img in enumerate([mad.image, mad.initial_image, mad.mad_image]):
+    category_probs, category = get_category(img)
+    likely_cats = "\n- ".join(list(imagenet_categories[category_probs > 0.05]))
+    most_likely_cat = imagenet_categories[category_probs.argmax()]
+    po.plot.imshow(img, ax=axes[i, 0], as_rgb=True, title=most_likely_cat)
+    po.plot.stem_plot(category_probs, ax=axes[i, 1], ylim=False)
+    axes[i, 1].set_title("Categories")
+    axes[i, 0].xaxis.set_visible(False)
+    axes[i, 0].yaxis.set_visible(False)
+    axes[i, 1].text(
```
billbrod, line 211: put probabilities next to category names?

billbrod (changes requested):
Thanks for this Raffles! I've got a lot of tweaks to wording throughout, and a couple of bigger feedback, but this looks great! 
 
- We handle citations using bibtex and [sphinxcontrib-bibtex](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/). Look at other notebooks for how this works (e.g., `docs/index.md`) but basically: add the bibtex entry for any papers to `docs/references.bib` and then cite the with 
    ```
    {cite:alp}`name-of-key`
    ```
- I think a lot of the bits under "Prepare model and image" can be combined into a single code block, with minimum explanation, pointing people to the "how to use deep nets" notebook if they want to know more (like in the replicating Feather et al notebook).
    - The only thing that should be really called out here is the choice of layer.

docs/user_guide/reproduce/adversarial_examples_mad.md
raffles-xingqi-zhu: I set vrange to "auto0". Let me know if this is better.

raffles-xingqi-zhu:
Thank you for all the edits and comments. I appreciate all the tweaks to wording and the notebook is much clearer as a result. 

What I still want to do is changing the seed to hopefully make the network classify the image to a funnier category :) 

There is something I noticed when I was playing with `vrange` parameter of po.plot.imshow() (when `as_rgb=False`). The color map changed from blue-red to greyscale when I changed from `auto0` to `auto1`. That was really weird.

billbrod:
> There is something I noticed when I was playing with `vrange` parameter of po.plot.imshow() (when `as_rgb=False`). The color map changed from blue-red to greyscale when I changed from `auto0` to `auto1`. That was really weird.

Yes, using `auto0` (or `indep0`) uses a diverging color map by default, whereas the others all use grayscale. You can override this by explicitly setting `cmap`. (That behavior is documented, if you look at `cmap` in the docstring, but the docstring for imshow is pretty long...)

pyproject.toml
```diff
@@ -81,6 +81,7 @@ nb = [
     "nbclient>=0.5.5",
     "torchvision>=0.3",
     "jupytext",
+    "myst-nb"
```
billbrod:
```suggestion
    "myst-nb",
```

adding the comma means that this line doesn't have to be changed if we add additional requirements below it.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -28,12 +27,15 @@ Run it in your browser: **{binder}`adversarial_examples_mad.ipynb`**!
...
+In this notebook we demonstrate how we can use the {class}`~plenoptic.MADCompetition` class to synthesize adversarial examples. Adversarial examples are tiny perturbations to an image that causes Deep Neural Networks to misclassify ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). MAD competition was developed to compare image quality metrics by generating a pair of images that have the same value for the reference metric but extremal values (highest and lowest) for the optimized metric {cite:alp}`Wang2008-maxim-differ`. These images were then used as stimuli in psychophysics experiments, to demonstrate which metric best aligned with human perception.
```
billbrod:
```suggestion
In this notebook we demonstrate how we can use the {class}`~plenoptic.MADCompetition` class to synthesize adversarial examples. Adversarial examples are tiny perturbations to an image that causes Deep Neural Networks to misclassify ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). MAD competition was developed to compare image quality metrics by generating a pair of images that have the same value for the reference metric but extremal values (highest and lowest) for the optimized metric ({cite:alp}`Wang2008-maxim-differ`). These images were then used as stimuli in psychophysics experiments, to demonstrate which metric best aligned with human perception.
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -28,12 +27,15 @@ Run it in your browser: **{binder}`adversarial_examples_mad.ipynb`**!
...
+
+Here, we demonstrate a different use of the {class}`~plenoptic.MADCompetition` class and show how its underlying machinery can be readily used to generate adversarial examples of Deep Neural Networks. At a high level, this is achieved by defining a reference metric in pixel space (the value of which we want to be low) and an optimized metric on distance of model classification (the value of which we want to be high).
```
billbrod:
```suggestion
Here, we demonstrate a different use of the {class}`~plenoptic.MADCompetition` class and show how its underlying machinery can be readily used to generate adversarial examples of Deep Neural Networks. At a high level, this is achieved by defining a reference metric in pixel space (the value of which we want to be low) and an optimized metric on distance in model classification space (the value of which we want to be high).
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -55,34 +57,21 @@ DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
...
 ## Prepare model and image for synthesis
 
-In this section, we initialize a plenoptic-compatible model using the weights from {external+torchvision:ref}`TorchVision <models>`. We use the standard ImageNet-trained ResNet50 as our network to attack. You may be also interested in checking out [](deep_nets) for details of choosing layer and preprocessing of the input image, and using models from {external+timm:doc}`timm <models>`.
-
-### Initialize deep neural network and pre-trained weights
+In this section, we initialize a plenoptic-compatible model using the weights from {external+torchvision:ref}`TorchVision <models>`. We use the standard ImageNet-trained ResNet50 as our network to attack.
 
-First, we download the model weights for ResNet50 trained on [ImageNet-1K](https://en.wikipedia.org/wiki/ImageNet#ImageNet-1K) and initialize the `torchvision` model.
+:::{seealso} Deep Nets in plenoptic
+Check out [](deep_nets) for more information, including how to choose different layers and preprocessing details, as well as how to use models from {external+timm:doc}`timm <models>`.
```
billbrod:
```suggestion
:::{admonition} Deep Nets in plenoptic
:class: seealso
Check out [](deep_nets) for more information, including how to choose different layers and preprocessing details, as well as how to use models from {external+timm:doc}`timm <models>`.
```

this didn't render correctly. I think the above is correct, but double-check it locally

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -132,20 +121,22 @@ First let us extract all the ImageNet categories:
 imagenet_categories = np.asarray(weights.meta["categories"])
 ```
 
-Let us define two helper functions. `convert_logits_to_probs` converts logits to probabilities that sum to 1. For an input image, `get_category` returns the vector containing the category probabilities and name of the category with the highest probability.
+Let us define two helper functions:
+- `convert_to_probs` converts the activation in the final fully-connected layer to probabilities that sum to 1. It gets used in the `get_category`, below.
```
billbrod:
```suggestion
- `convert_to_probs` converts the activation in the final fully-connected layer to probabilities that sum to 1. It gets used in `get_category`, below.
```

(or "in the `get_category` function")

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -132,20 +121,22 @@ First let us extract all the ImageNet categories:
...
+    category_probs = convert_to_probs(deepnet(norm(image))).detach().cpu()
     category = imagenet_categories[category_probs.argmax()]
     return category_probs, category
 ```
 
-ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The category, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is also highly confident in its classification (probability of ~0.8)
+ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The category of our initial image, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is also highly confident in its classification (probability of ~0.8)
```
billbrod: I thinks this can be split up around the figure. Keep the first sentence above the paragraph and add something like: "The following plot shows the classification probabilities for our initial image as a stem plot. Each of the 1000 categories is represented by a line, whose y-value gives the model's probability that the image belongs to the corresponding category (the x-value is arbitrary). The title shows the label of the most likely category."

And then put everything starting at "The category of our initial image..." after the figure. And change the last line to be like "Notice the model is highly confident in its classification, with a probability of about 0.8 and no other category exceeding a probability of 0.1."

billbrod: it may be worth printing out the other categories as well? it would be natural to wonder what the other non-zero categories are (they're other primates, right?)

billbrod: by print out, I mean put on the figure in some way

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -154,14 +145,25 @@ po.plot.stem_plot(category_probs, title=category);
...
+To qualify as an adversarial example, the image must satisfy two requirements: (1) the perturbation in image space is small and (2) the model outputs an incorrect classification with high confidence ({cite:alp}`goodfellow_explaining_2015`). Conveniently, we already have these two ingredients built into the MAD competition framework. More concretely, we should be able to meet these two requirements if we define:
+
+1. The reference metric as a distance in pixel space and set it to a low value.
+2. The optimized metric as a distance in representation space, so that we synthesize an image whose representation is as far away from the original as possible.
+
+For the reference metric, we use mean-squared-error (MSE):
 
 ```{code-cell} ipython3
 def reference_metric(x, y):
     return po.metric.mse(x, y).mean()  # .mean() averages across the RGB channels
```
billbrod:
```suggestion
    # compute MSE independently per RGB channel and then average.
    return po.metric.mse(x, y).mean()
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -154,14 +145,25 @@ po.plot.stem_plot(category_probs, title=category);
...
 ```
 
-For the optimized metric, we use the MSE of output logits in the last layer of the network.
+For the optimized metric, we use the MSE on the output of the last layer of the network.
+
+:::{admonition} Why not define the optimized metric on probabilities?
+:class: dropdown hint
+
+While a vector of probabilities is the more natural way to think about the output of the network, logits provide smooth and differentiable gradients necessary for training neural networks via gradient descent.
```
billbrod:
```suggestion
While a vector of probabilities (`convert_to_probs(model(x))` instead of `model(x)`) is the more natural way to think about the output of the network, [logits](https://en.wikipedia.org/wiki/Logit) (the representation before calling `softmax`) provide the smooth and differentiable gradients necessary for well-behaved optimization.
```

we're not training neural networks here, and users aren't necessarily familiar with logits.

billbrod: This is true, but vague. Could add a sentence saying something like: "In particular, since many of the initial image's classification probabilities are exactly zero, optimizing on the probabilities means the resulting adversarial example will never be classified as one of those classes (since the gradients are zero)."

billbrod: and put this dropdown after the box defining `optimized_metric`

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -200,42 +204,81 @@ po.plot.synthesis_status(mad);
 
 ## Visualizing the adversarial image
 
-Let us compare how the synthesized image compares to the original and initial images. In the bottow row the original image is subtracted to visualize the changes in pixels. There is no change in the left image. In the middle image we see some hardly noticeable noise. In the last image we see a low amount of pixel noise.
+Let us compare how the synthesized image compares to the original and initial images. In the bottow row the original image is subtracted to visualize the changes in pixels. There is no change in the left image. In the middle image we see some hardly noticeable noise. In the last image we see a small amount of random pixel noise, distributed across the image.
```
billbrod: needs to be updated to reflect the new figure layout

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -200,42 +204,81 @@ po.plot.synthesis_status(mad);
...
+
+for ax in axes.flat:
+    ax.xaxis.set_visible(False)
+    ax.yaxis.set_visible(False)
+
+fig.tight_layout();
 ```
 
-We can also visualize the difference in each color channel for the initial image (top row) and adversarial image (bottom row).
+The noise is too faint to discern any structures. Let us visualize the difference in each color channel separately (RGB, or red, green, and blue) for the initial image (top row) and adversarial image (bottom row). Note the pixel value range is different between the top and bottom rows.
```
billbrod:
```suggestion
In the above plot, the noise is too faint to discern any structures. Let us visualize the difference in each color channel (red, green, and blue) separately for the initial image (top row) and adversarial image (bottom row). Note the pixel value range is different between the top and bottom rows.
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -200,42 +204,81 @@ po.plot.synthesis_status(mad);
...
 
 ```{code-cell} ipython3
-channelwise_diffs = [mad.initial_image - img, mad.mad_image - img]
-po.plot.imshow(channelwise_diffs, col_wrap=3);
+channelwise_diffs_initial = mad.initial_image - img
+titles = ["Initial (R)", "Initial (G)", "Initial (B)"]
```
billbrod:
```suggestion
titles = ["Initial - Original (R)", "Initial - Original (G)", "Initial - Original (B)"]
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -200,42 +204,81 @@ po.plot.synthesis_status(mad);
...
+po.plot.imshow(channelwise_diffs_initial, col_wrap=3, title=titles, vrange="auto0")
+channelwise_diffs_mad = mad.mad_image - img
+titles = ["Adversarial (R)", "Adversarial (G)", "Adversarial (B)"]
```
billbrod:
```suggestion
titles = ["Adversarial - Original (R)", "Adversarial - Original (G)", "Adversarial - Original (B)"]
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -200,42 +204,81 @@ po.plot.synthesis_status(mad);
...
+po.plot.imshow(channelwise_diffs_mad, col_wrap=3, title=titles, vrange="auto0");
 ```
 
-Finally let us visualize the category probabilities of the original, initial, and advesarial images using stem plots. We see that the network thinks the synthesized image is a burrito with almost 100% certainty! We have successfully generated an adversarial example of the network.
+Finally let us visualize the category probabilities of the original, initial, and advesarial images using stem plots.
```
billbrod: Tell people what they should see in the above plots, something like: "In the top row, we can see that the difference between the initial and original images is noise, randomly distributed across the image, with no structure. Given that the initial image is the original plus normally-distributed noise, that must be true. In the bottom row, the pattern has shifted, with the differences grouping together somewhat. However, this difference does not look like a $CATEGORY, or similar."

where $CATEGORY is the output of the glue, the class of the adversarial example

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -200,42 +204,81 @@ po.plot.synthesis_status(mad);
...
 
 ```{code-cell} ipython3
```
billbrod: I think this plot should probably come first in this section. This is the evidence that synthesis succeeded, the other plots are trying to see whether we can understand what's happening in the difference

billbrod (changes requested):
Thanks Raffles, this is coming along!

In addition the comments below, I still think the first "Prepare model and image..." section can be combined more. See what I did in the "Reproducing Feather" notebook: put them all in a single code block and add text above that says something like: "In the following block, we create a DeepNetFeatures model matching the output of the final fully connected layer of ResNet50. ADD SOMETHING ABOUT THIS LAYER. We then prepare the image. Finally, we ensure that the model and image have the proper device and dtype, and remove the gradient from all model parameters. To learn more about any of these steps and why we take them, read `[Using Deep Neural Networks with plenoptic]()`."

pyproject.toml
raffles-xingqi-zhu: That's really cool, I always wondered why people do that as it looks extra. But it now makes complete sense.

raffles-xingqi-zhu:
Thank you for all the suggestions! I've made changes in response and please let me know what you think. 

One thing I'm not sure about is the wording of this sentence: 

"At a high level, this is achieved by defining a reference metric in pixel space (the value of which we want to be low) and an optimized metric on distance in model classification space (the value of which we want to be high)."

I feel "an optimized metric on distance in model classification space" is a bit confusing. Any thoughts on making this clearer as this is one of the first things they'll read?

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -57,57 +57,26 @@ DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
...
 ```
 
 ## Prepare model and image for synthesis
 
-In this section, we initialize a plenoptic-compatible model using the weights from {external+torchvision:ref}`TorchVision <models>`. We use the standard ImageNet-trained ResNet50 as our network to attack.
-
-:::{seealso} Deep Nets in plenoptic
-Check out [](deep_nets) for more information, including how to choose different layers and preprocessing details, as well as how to use models from {external+timm:doc}`timm <models>`.
-:::
+In the following block, we create a {class}`~plenoptic.models.DeepNetFeatures` model matching the output of the final fully connected layer of ResNet50. We chose the final layer because the goal of adversarial attack is misclassification and the final layer contains probabilities for the 1000 categories. After creating the model, we then prepare the image. Finally, we ensure that the model and image have the proper device and dtype, and remove the gradient from all model parameters. To learn more about any of these steps and why we take them, read [](deep_nets).
```
billbrod:
```suggestion
In the following block, we create a {class}`~plenoptic.models.DeepNetFeatures` model matching the output of the final fully connected layer of ResNet50. We chose the final layer because the goal of adversarial attack is misclassification and the final layer corresponds to the probabilities for the 1000 categories. After creating the model, we then prepare the image. Finally, we ensure that the model and image have the proper device and dtype, and remove the gradient from all model parameters. 

To learn more about any of these steps and why we take them, read [](deep_nets).
```

without the transformation, the final layer isn't probabilities. I wanted to say "is proportional to the probabilities..." but that's not right. is "corresponds" better? or something more general, like "is related to" or "represents the model's confidence that the image belongs to each of the 1000 categories" (though I don't like anthropomorphizing language...

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -136,13 +105,19 @@ def get_category(image):
...
+ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The following plot shows the classification probabilities for our initial image as a stem plot. Each of the 1000 categories is represented by a line, whose y-value gives the model's probability that the image belongs to the corresponding category (the x-value is arbitrary). The title shows the label of the most likely category.
 
 ```{code-cell} ipython3
 category_probs, category = get_category(img)
-po.plot.stem_plot(category_probs, title=category);
+po.plot.stem_plot(category_probs, title=category)
+likely_cats = "\n- ".join(list(imagenet_categories[category_probs > 0.01]))
+plt.text(700, 0.5, f"Likely categories:\n- {likely_cats}")
```
billbrod:
```suggestion
plt.text(700, 0.5, f"Likely categories:\n- {likely_cats}");
```

to suppress the plotting related output. this might run afoul of one of my linters, in which case you can assign it to somethign to suppress as well: `_ = plt.text(...`

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -136,13 +105,19 @@ def get_category(image):
...
 ```
 
+The category of our initial image, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is highly confident in its classification, with a probability of about 0.8 and no other category exceeding a probability of 0.1. We also show any categories the model predicts with a probability higher than 0.01 and they are all non-human primates.
```
billbrod:
```suggestion
The category of our initial image, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is highly confident in its classification, with a probability of about 0.8 and no other category exceeding a probability of 0.1. The other predicted categories are all [Old World monkeys](https://en.wikipedia.org/wiki/Old_World_monkey).
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -136,13 +105,19 @@ def get_category(image):
     return category_probs, category
 ```
 
-ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The category of our initial image, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is also highly confident in its classification (probability of ~0.8)
+ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The following plot shows the classification probabilities for our initial image as a stem plot. Each of the 1000 categories is represented by a line, whose y-value gives the model's probability that the image belongs to the corresponding category (the x-value is arbitrary). The title shows the label of the most likely category.
```
billbrod:
```suggestion
ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The following plot shows the classification probabilities for our initial image as a stem plot. Each of the 1000 categories is represented by a line, whose y-value gives the model's probability that the image belongs to the corresponding category (the x-value is arbitrary). The title shows the label of the most likely category, and the text on the plot shows the other categories with probability higher than 0.01.
```

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -204,10 +182,39 @@ po.plot.synthesis_status(mad);
...
+    )
+category_probs, category = get_category(mad.mad_image)
+glue("category_name", str(category), display=False)
+```
+
+We see our initial noise has not changed the original category. At the end of synthesis the network is highly confident that the synthesized image is a {glue}`category_name`!
+
++++
+
+While the synthesized image is visually similar to the original image (also note the low MSE reference metric loss), let us verify more rigorously. In the bottow row,  the original image is subtracted from the initial image and synthesized image to visualize the changes in pixel values. Between initial and original images, the difference is hardly noticeable (middle panel, bottom row). The difference between synthesized and original images looks like random pixel noise distributed across the image (right panel, bottom row). However, the noise is too faint for us to reliably discern any structures in both cases.
```
billbrod: MSE reference metric not shown explicitly in the figure before this. Can we add MSE to the plots above? as text or title somewhere

billbrod:
```suggestion
While the synthesized image is visually similar to the original image (also note the low MSE reference metric loss), let us verify more rigorously. In the bottow row,  the original image is subtracted from the initial image and synthesized image to visualize the changes in pixel values. 

Between initial and original images, the difference is hardly noticeable (middle panel, bottom row). The difference between synthesized and original images looks like random pixel noise distributed across the image (right panel, bottom row). However, the noise is too faint for us to tell if there's any structure.
```

I would move the second paragraph here to after the figure

billbrod (changes requested):
This is getting close! Just some small tweaks from me here.

> "At a high level, this is achieved by defining a reference metric in pixel space (the value of which we want to be low) and an optimized metric on distance in model classification space (the value of which we want to be high)."
>
> I feel "an optimized metric on distance in model classification space" is a bit confusing. Any thoughts on making this clearer as this is one of the first things they'll read?

Hmm you're right. Maybe: "To do so, we will set up the problem to fix a distance in pixel space at some low value while maximizing the distance in model output space."

Or "we will try to fix" instead of "we will setup the problem". something like that?

docs/user_guide/reproduce/adversarial_examples_mad.md
raffles-xingqi-zhu: "correspond" makes sense to me. Resolving this comment.

raffles-xingqi-zhu:
Thank you for the suggestions. I've accepted/implemented changes accordingly.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+This notebook requires the optional dependency `torchvision`, which can be installed with `pip`.
+:::
+
+In this notebook we demonstrate how we can use the {class}`~plenoptic.MADCompetition` class to synthesize adversarial examples. Adversarial examples are tiny perturbations to an image that causes Deep Neural Networks to misclassify ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). MAD competition was developed to compare image quality metrics by generating a pair of images that have the same value for the reference metric but extremal values (highest and lowest) for the optimized metric ({cite:alp}`Wang2008-maxim-differ`). These images were then used as stimuli in psychophysics experiments, to demonstrate which metric best aligned with human perception.
+
+Here, we demonstrate a different use of the {class}`~plenoptic.MADCompetition` class and show how its underlying machinery can be readily used to generate adversarial examples of Deep Neural Networks. At a high level, this is achieved by defining a reference metric in pixel space (the value of which we want to be low) and an optimized metric in model output space (the value of which we want to be high).
```
BalzaniEdoardo: I would stop at "Deep Neural Networks." The high-level description doesn't help my understanding of the framework, I think I know too little at this point to map the sentence to the adversarial example problem.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+# so that relative sizes of axes created by po.plot.imshow and others look right
+plt.rcParams["figure.dpi"] = 72
+
+# set seed for reproducibility
+po.set_seed(4)
+```
+
+## Prepare model and image for synthesis
+
+In the following block, we create a {class}`~plenoptic.models.DeepNetFeatures` model matching the output of the final fully connected layer of ResNet50. We chose the final layer because the goal of adversarial attack is misclassification and the final layer corresponds to the probabilities for the 1000 categories. After creating the model, we then prepare the image. Finally, we ensure that the model and image have the proper device and dtype, and remove the gradient from all model parameters.
```
BalzaniEdoardo: I would add a link to resources on ResNet50, either a model architecture, a wiki page or something like that. I would also be explicit and say:
"on a classic image recognition network, ResNet50." or something like that

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+img = po.process.center_crop(img, transform.crop_size[0])
+po.plot.imshow(img, as_rgb=True)
+img = img.to(DEVICE).to(torch.float64)
+model.to(DEVICE).to(torch.float64)
+deepnet.to(DEVICE).to(torch.float64)
+po.remove_grad(model)
+```
+
+## Visualizing classification of the clean image
+First let us extract all the ImageNet categories:
```
BalzaniEdoardo: Also link to ImageNet

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+
++++
+
+## Define optimized and reference metric
+
+To qualify as an adversarial example, the image must satisfy two requirements: (1) the perturbation in image space is small and (2) the model outputs an incorrect classification with high confidence ({cite:alp}`goodfellow_explaining_2015`). Conveniently, we already have these two ingredients built into the MAD competition framework. More concretely, we should be able to meet these two requirements if we define:
+
+1. The reference metric as a distance in pixel space and set it to a low value.
+2. The optimized metric as a distance in representation space, so that we synthesize an image whose representation is as far away from the original as possible.
```
BalzaniEdoardo, line 129: This sentence fully clarifies the high-level description in the intro. This confirms that I would drop the intro sentence, which was confusing at that point since once the reader arrives here, the framework becomes clear

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+:::
+
++++
+
+## Synthesize the adversarial image
+
+We want to maximize {attr}`~plenoptic.MADCompetition.optimized_metric` while holding {attr}`~plenoptic.MADCompetition.reference_metric` fixed. Therefore we set {attr}`~plenoptic.MADCompetition.minmax` to "max". {attr}`~plenoptic.MADCompetition.metric_tradeoff_lambda` controls the relative weight of {attr}`~plenoptic.MADCompetition.optimized_metric_loss` and {attr}`~plenoptic.MADCompetition.reference_metric_loss` in the objective function. We found a {attr}`~plenoptic.MADCompetition.metric_tradeoff_lambda` value of 1e10 generally produced good adversarial examples for the monkey image.
```
BalzaniEdoardo: fixed or keeping it small?

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+
+```{code-cell} ipython3
+mad = po.MADCompetition(
+    img, optimized_metric, reference_metric, "max", metric_tradeoff_lambda=1e10
```
BalzaniEdoardo: is minmax the name of the keyword argument? If so
 
```suggestion
    img, optimized_metric, reference_metric, minmax="max", metric_tradeoff_lambda=1e10
```

Otherwise the  the description (setting the minmax) is less obvious

BalzaniEdoardo: Also, long code line, consider a multiline

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+)
+```
+
+:::{admonition} How does {attr}`~plenoptic.MADCompetition.metric_tradeoff_lambda` affect the adversarial image
```
BalzaniEdoardo: I think two admonitions or an admonition and a link would work here.

One is this admonition, explaining what the tradeoff does, and another should advise on how to find a good one. 

Both what and how could be in the same admonition, I think the how part should be a little bit more specific than the current version.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+
+:::
+
+We set the initial noise to be a small value so that our synthesized image also has small deviations in the pixel values from the original. We also decreased the learning rate from the default, as this resulted in better solutions in our experiments.
+
+```{code-cell} ipython3
+mad.setup(initial_noise=0.001, optimizer_kwargs={"lr": 0.001})
+```
+
+Running the synthesis generates an image that looks just like the original image but we see the optimized metric loss has increased significantly. Even though the reference metric loss measured in pixel space has also increased, it remains a small value. The fact that the optimized metric has significantly increased shows that the model thinks this is a very different category than our initial image, which we'll show in the next section.
```
BalzaniEdoardo: Q: are the two losses on the same scale to begin with?

I not, plotting the relative increment from the initial value would be more informative. It looks like the optimized metric increased of 2 order of magnitude, the reference of just one. 

If you can normalize the plot to a percent increase, you could share the y-axis in the plot and make the differential effect of loss-changes more evident.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+    )
+    po.plot.stem_plot(category_probs, ax=axes[i, 1], ylim=False)
+    axes[i, 1].set_title("Categories")
+    axes[i, 0].xaxis.set_visible(False)
+    axes[i, 0].yaxis.set_visible(False)
+    axes[i, 1].text(
+        1, 0.5, f"Likely categories:\n- {likely_cats}", transform=axes[i, 1].transAxes
+    )
+category_probs, category = get_category(mad.mad_image)
+glue("category_name", str(category), display=False)
```
BalzaniEdoardo, line 215: Are the images on the same vmim vmax? It would be worth making explicit.

BalzaniEdoardo (commented):
This example is very fun, and clearly explained. I had a few minor comments worth considering

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+:::
+
+(adversarial_examples)=
+# Generate adversarial examples using MAD Competition
+
+:::{warning}
+This notebook requires the optional dependency `torchvision`, which can be installed with `pip`.
+:::
+
+In this notebook we demonstrate how we can use the {class}`~plenoptic.MADCompetition` class to synthesize adversarial examples. Adversarial examples are tiny perturbations to an image that causes Deep Neural Networks to misclassify ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). MAD competition was developed to compare image quality metrics by generating a pair of images that have the same value for the reference metric but extremal values (highest and lowest) for the optimized metric ({cite:alp}`Wang2008-maxim-differ`). These images were then used as stimuli in psychophysics experiments, to demonstrate which metric best aligned with human perception.
```
emeyer121:
```suggestion
In this notebook we demonstrate how we can use the {class}`~plenoptic.MADCompetition` class to synthesize adversarial examples. Adversarial examples are tiny perturbations to an image that cause Deep Neural Networks to misclassify stimuli ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). MAD competition was developed to compare image quality metrics by generating a pair of images that have the same value for a reference metric but extremal values (highest and lowest) for an optimized metric ({cite:alp}`Wang2008-maxim-differ`). These images were then used as stimuli in psychophysics experiments, to demonstrate which metric best aligned with human perception.
```

docs/user_guide/reproduce/adversarial_examples_mad.md
emeyer121: agreed, or something clarifying that you are trying to define the reference metric and optimized metric (if that is what you are referring to), using language from previous paragraph

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+
+img = po.data.macaque()
+img = po.process.blur_downsample(img, 2)[..., :-59, :]
+img = po.process.center_crop(img, transform.crop_size[0])
+po.plot.imshow(img, as_rgb=True)
+img = img.to(DEVICE).to(torch.float64)
+model.to(DEVICE).to(torch.float64)
+deepnet.to(DEVICE).to(torch.float64)
+po.remove_grad(model)
+```
```
emeyer121, line 87: not necessary, but wouldn't be opposed to adding a few comments within this block of code with a general summary of each step just because it's a lot to take in

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+    return torch.nn.functional.softmax(logits, dim=1).squeeze()
+
+
+def get_category(image):
+    category_probs = convert_to_probs(deepnet(norm(image))).detach().cpu()
+    category = imagenet_categories[category_probs.argmax()]
+    return category_probs, category
+```
+
+ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The following plot shows the classification probabilities for our initial image as a stem plot. Each of the 1000 categories is represented by a line, whose y-value gives the model's probability that the image belongs to the corresponding category (the x-value is arbitrary). The title shows the label of the most likely category, and the text on the plot shows the other categories with probability higher than 0.01.
```
emeyer121: you also mention the 1000 categories in a previous paragraph. don't think it's necessary to say twice but I'd add the 1000 category link above if you remove this.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+
+For the reference metric, we use mean-squared-error (MSE):
+
+```{code-cell} ipython3
+def reference_metric(x, y):
+    # compute MSE independently per RGB channel and then average.
+    return po.metric.mse(x, y).mean()
+```
+
+For the optimized metric, we use MSE on the output of the last layer of the network.
```
emeyer121: you say this you're using MSE on the output of the last layer but also saying that the probabilities come from the final fully-connected layer which may be confused as the same thing. I would clarify more what the actual difference is in what you used previously vs what you're using now.

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+Running the synthesis generates an image that looks just like the original image but we see the optimized metric loss has increased significantly. Even though the reference metric loss measured in pixel space has also increased, it remains a small value. The fact that the optimized metric has significantly increased shows that the model thinks this is a very different category than our initial image, which we'll show in the next section.
+
+```{code-cell} ipython3
+mad.synthesize(1000)
+po.plot.synthesis_status(mad);
+```
+
+## Visualizing the adversarial image
+
+First let us visualize the category probabilities of the original, initial, and advesarial images using stem plots.
```
emeyer121:
```suggestion
First let us visualize the category probabilities of the original, initial, and adversarial images using stem plots.
```

emeyer121: also maybe clarify what the difference between the original and initial images are

docs/user_guide/reproduce/adversarial_examples_mad.md
```diff
@@ -0,0 +1,276 @@
...
+    ax.set_title(ax.get_title(), fontsize=14)
+
+for ax in axes.flat:
+    ax.xaxis.set_visible(False)
+    ax.yaxis.set_visible(False)
+
+fig.tight_layout();
+```
+
+Between initial and original images, the difference is hardly noticeable (middle panel, bottom row). The difference between synthesized and original images looks like random pixel noise distributed across the image (right panel, bottom row). However, the noise is too faint for us to tell if there's any structure. Instead, let us try visualizing the difference in each color channel (red, green, and blue) separately for the initial image (top row) and adversarial image (bottom row). Note the pixel value range is different between the top and bottom rows.
```
emeyer121: not a huge deal, but before seeing the next figures, I thought the top row and bottom row were referring to the previous figures and I was a little confused here. Maybe just remove the parentheticals here because you explain them more after the figure.

emeyer121:
Looks great! This is a really interesting use case of MAD competition. Added a few minor comments about places to improve clarity but overall sounds good and is understandable.

docs/user_guide/reproduce/adversarial_examples_mad.md
raffles-xingqi-zhu: This is a good point. Let me know whether what I added addresses the confusion.

docs/user_guide/reproduce/adversarial_examples_mad.md
raffles-xingqi-zhu: If I understand your question correctly, they are not. Optimized metric is defined in model output space whereas reference metric is defined in pixel space. The two metrics' values are not directly comparable. 

What we are looking for, as you mentioned, is the relative change in each metric from the initial value. The percentage idea is interesting and would likely require modifying plenoptic.plot.synthesis_loss() to take another argument. I will discuss this possibility with Billy.

docs/user_guide/reproduce/adversarial_examples_mad.md
raffles-xingqi-zhu: Could you clarify on which line is unnecessary? Thanks!

emeyer121: I just meant it's not a necessary change to make but it might be helpful to give a bit more context within the code block about what is happening

raffles-xingqi-zhu:
Thank you both for your comments and edits. They revealed some blindspots I had and I made changes to fix those. I left two of Edoardo's comments unresolved for now because I'm waiting to hear what you think/discuss with Billy.

docs/user_guide/reproduce/adversarial_examples_mad.md
raffles-xingqi-zhu: I have had a chat with Billy and we agreed from the perspective of synthesis and user understanding what is going on under the hood, it is better to show the unaltered metric values. In fact, I think comparing the order of magnitude change for reference and optimized metric can be misleading (1 vs 2) because they are not on the same scale. I'll explain this in greater details in the text.

billbrod (approved):
This looks great! Thanks Raffles for all your work, and Emily and Edoardo for your feedback!

PR #472, opened 2026-07-31
Fix flaky zenodo url checker

### LINKED ISSUE #471: Fix flaky url checker
Our URL checker action regularly (almost every time) fails because of an issue resolving the zenodo doi. Look into the right way to fix that.

billbrod:

Our url checker action regularly fails because of the zenodo badge svg, but every time I click on the link to manually check it, it's fine.

~Here, we try simply increasing the timeout and retries (as [suggested](https://github.com/urlstechie/urlchecker-action/issues/93)). If that doesn't work, I'll exclude it from the check (which isn't ideal, but I think if there's a problem with zenodo one of the other urls will also go down).~

That approach did not work. Instead, inspired by https://github.com/GenericMappingTools/pygmt/issues/4665, I swapped out the zenodo badges with a static one from https://shields.io with the same content.

Closes #471

**Checklist**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,7 +184,7 @@ jobs:
       with:
         file_types: .md,.py,.rst
         print_all: false
-        timeout: 15
+        timeout: 25
         retry_count: 5
         exclude_urls: https://github.com/github-username/plenoptic.git
   validate_switcher_json:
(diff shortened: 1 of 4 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/472

PR #476, opened 2026-08-11
Metamer adversarial example

raffles-xingqi-zhu:

> This PR shows how to generate adversarial examples using the `metamer` class.

> It would be great if you could check the clarity of explanation. 

> This notebook shares some redundancy with the other notebook using MAD competition to generate adversarial examples. My philosophy for now is keeping the redundancy because merging them into one would be too much. Thoughts on this? 

**Describe changes**

> Full working example showing how to generate adversarial examples with the `metamer` class. 

**Checklist**


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/476

docs/user_guide/reproduce/feather2023.md
```diff
@@ -39,7 +39,7 @@ Run it in your browser: **{binder}`feather2023.ipynb`**!
 
 :::
 
-(feather2023)=
+(feather_2023)=
```
billbrod:
```suggestion
(feather2023-resnet50)=
```

I'd prefer something that's more obviously different from the file name and that gives a little more info about the content.

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+
+(adversarial_examples_metamer)=
+# Generate adversarial examples using Metamer
+
+:::{warning}
+This notebook requires the optional dependency `torchvision`, which can be installed with `pip`.
+:::
+
+In this notebook we demonstrate how we can use the {class}`~plenoptic.Metamer` class to synthesize adversarial examples. Adversarial examples are images with subtle, imperceptible perturbations designed to deceive a Deep Neural Networks into making an incorrect classification ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). In a non-strict sense, an adversarial example is a metamer of the new (misclassified) class because the model has the same classification behaviour for the adversarial image and other images in the class. On this basis, we can use {class}`~plenoptic.Metamer` to generate adversarial examples by matching the model output representation between the synthesized image and another image in that class. For an in-depth dive of creating metamers of Deep Neural Networks, read [](feather_2023) and [](deep_nets).
+
```
billbrod, line 31: Add a new line here saying something like, "For an alternate approach to synthesizing adversarial examples, see [](adversarial_examples_mad)." (and add a similar line in that notebook)

docs/user_guide/reproduce/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+
+# set seed for reproducibility
+po.set_seed(4)
+```
+
+## Prepare model and images for synthesis
+
+In the following block, we create a {class}`~plenoptic.models.DeepNetFeatures` model matching the output of "layer4" of a classic image recognition network, [ResNet50](https://en.wikipedia.org/wiki/Residual_neural_network), trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/).. After creating the model, we then prepare the original image and an image of a different class whose represetnation we try to match. For this example let us fool the network into thinking a macaque is a cheeseburger (image taken from the Caltech256 dataset {cite:alp}`griffin_caltech_2022`)! Finally, we ensure that the model and images have the proper device and dtype, and remove the gradient from all model parameters.
+
+To learn more about any of these steps and why we take them, read [](deep_nets).
```
billbrod:
```suggestion
In the following block, we create a {class}`~plenoptic.models.DeepNetFeatures` model matching the output of `"layer4"` of a classic image recognition network, [ResNet50](https://en.wikipedia.org/wiki/Residual_neural_network), trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/).. After creating the model, we then prepare the original image and an image of a different class whose represetnation we try to match. Finally, we ensure that the model and images have the proper device and dtype, and remove the gradient from all model parameters.

To learn more about any of these steps and why we take them, read [](deep_nets).

As part of this procedure, we must choose an image whose `layer4` representation (and thus, categorization) we would like to match. For this example let us fool the network into thinking a macaque is a cheeseburger (image taken from the Caltech256 dataset {cite:alp}`griffin_caltech_2022`).
```
something like that

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+# set seed for reproducibility
+po.set_seed(4)
+```
+
+## Prepare model and images for synthesis
+
+In the following block, we create a {class}`~plenoptic.models.DeepNetFeatures` model matching the output of "layer4" of a classic image recognition network, [ResNet50](https://en.wikipedia.org/wiki/Residual_neural_network), trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/).. After creating the model, we then prepare the original image and an image of a different class whose represetnation we try to match. For this example let us fool the network into thinking a macaque is a cheeseburger (image taken from the Caltech256 dataset {cite:alp}`griffin_caltech_2022`)! Finally, we ensure that the model and images have the proper device and dtype, and remove the gradient from all model parameters.
+
+To learn more about any of these steps and why we take them, read [](deep_nets).
+
```
billbrod, line 68: why layer4? should probably say something about that, maybe as an admonition

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+img = po.process.blur_downsample(img, 2)[..., :-59, :]
+img = po.process.center_crop(img, transform.crop_size[0])
+po.plot.imshow(img, as_rgb=True)
+
+target_img = po.load_images(po.data.fetch_data("caltech256_burger.jpg"), as_gray=False)
+target_img = torchvision.transforms.functional.resize(
+    target_img, [img.shape[-2], img.shape[-1]], antialias=True
+)
+po.plot.imshow(target_img, as_rgb=True)
+
```
billbrod, line 88: instead of the separate imshows, do a single `po.plot.imshow([img, target_img], titles=["Initial image", "Target image"], as_rgb=True)`

docs/user_guide/reproduce/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+
+img = po.data.macaque()
+img = po.process.blur_downsample(img, 2)[..., :-59, :]
+img = po.process.center_crop(img, transform.crop_size[0])
+po.plot.imshow(img, as_rgb=True)
+
+target_img = po.load_images(po.data.fetch_data("caltech256_burger.jpg"), as_gray=False)
+target_img = torchvision.transforms.functional.resize(
+    target_img, [img.shape[-2], img.shape[-1]], antialias=True
+)
```
billbrod: This distorts the image. Instead, let's just crop out the center 224 pixels:
```suggestion
target_img = po.process.center_crop(target_img, img.shape[-1])
```
(if this changes the behavior we can come up with a different way of doing it)

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+plt.text(700, 0.5, f"Likely categories:\n- {likely_cats}");
+```
+
+The category of our initial image, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is highly confident in its classification, with a probability of about 0.8 and no other category exceeding a probability of 0.1. The other predicted categories are all [Old World monkeys](https://en.wikipedia.org/wiki/Old_World_monkey). Now let us make the same plot for the target image.
+
+```{code-cell} ipython3
+category_probs, category = get_category(target_img)
+po.plot.stem_plot(category_probs, title=category)
+likely_cats = "\n- ".join(list(imagenet_categories[category_probs > 0.01]))
+plt.text(700, 0.5, f"Likely categories:\n- {likely_cats}");
```
billbrod: change location of this, it's overlapping with the stem

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+```
```
billbrod, line 148: the categories printed on the plot are shown in the order they are present in the class list, which is arbitrary. can we instead sort them by probability? because that's what I thought it was doing anyway (This would apply to the other examples of this shown in this notebook and the other adversarial example one.)

docs/user_guide/reproduce/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+plt.text(700, 0.5, f"Likely categories:\n- {likely_cats}");
+```
+
+The category of the target image is cheeseburger. This is the desired category of our adversarial example.
+
++++
+
+## Synthesize the adversarial image
+
+To qualify as an adversarial example, the image must satisfy two requirements: (1) the perturbation in image space is small and (2) the model outputs an incorrect classification with high confidence ({cite:alp}`goodfellow_explaining_2015`). By starting with the original image and changing pixel values until the model output representation is matched between the synthesized and target images, the model should misclasify the synthesized image. However, we also need to constrain the synthesized image to be close to the original image in pixel space. To do this we define a penalty function that calculates the Mean Squared Error (MSE) between the original and synthesized images. In the penalty function we also add another term that penalizes the range of the synthesized image pixel values being outside 0 and 1.
```
billbrod:
```suggestion
To qualify as an adversarial example, the image must satisfy two requirements ({cite:alp}`goodfellow_explaining_2015`): 
1. The perturbation in image space is small.
2. The model outputs an incorrect classification with high confidence.

By starting with the original image and changing pixel values until the model output representation is matched between the synthesized and target images, the model should misclassify the synthesized image. However, we also need to constrain the synthesized image to be close to the original image in pixel space. To do this we define a penalty function that calculates the Mean Squared Error (MSE) between the original and synthesized images. In the penalty function we also add another term that penalizes pixels in the synthesized image whose values values outside $[0, 1]$.
```

docs/user_guide/reproduce/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+
+The relative weight of the penalty in the objective function is controlled by the {attr}`~plenoptic.Metamer.penalty_lambda`. We found a value of 1000 generally worked well in our experiments.
+
+```{code-cell} ipython3
+met = po.Metamer(
+    target_img, model, penalty_function=custom_penalty, penalty_lambda=1000
+)
+```
+
+:::{admonition} How does {attr}`~plenoptic.Metamer.penalty_lambda` affect the adversarial image
```
billbrod:
```suggestion
:::{admonition} How does {attr}`~plenoptic.Metamer.penalty_lambda` affect the adversarial image?
```

docs/user_guide/reproduce/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+:class: dropdown hint
+
+The objective function for metamer synthesis is made of two parts: the synthesis loss that measures the difference in representation between synthesized and target images, and the penalty. As {attr}`~plenoptic.Metamer.penalty_lambda` increases, the relative weight of the penalty in the objevtive increases. This helps ensure the synthesized and original images are close to each other in pixel space, at the cost of making representation matching more difficult. If you are applying this procedure to new images or new image classification models, you will almost certainly need to experiment to find the appropriate {attr}`~plenoptic.Metamer.penalty_lambda`.
```
billbrod:
```suggestion
The objective function for metamer synthesis is made of two parts: the synthesis loss that measures the difference in representation between synthesized and target images, and the penalty. As {attr}`~plenoptic.Metamer.penalty_lambda` increases, the relative weight of the penalty in the objective increases. This helps ensure the synthesized and original images are close to each other in pixel space, at the cost of making representation matching more difficult. If you are applying this procedure to new images or new image classification models, you will almost certainly need to experiment to find the appropriate {attr}`~plenoptic.Metamer.penalty_lambda`.
```

billbrod: do you have any specific examples from the experiments you did? e.g., "We found that at a value of 1, the penalty was essentially ignored, so the resulting image was very distant in pixel space" or the like?

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+:::{admonition} How does {attr}`~plenoptic.Metamer.penalty_lambda` affect the adversarial image
+:class: dropdown hint
+
+The objective function for metamer synthesis is made of two parts: the synthesis loss that measures the difference in representation between synthesized and target images, and the penalty. As {attr}`~plenoptic.Metamer.penalty_lambda` increases, the relative weight of the penalty in the objevtive increases. This helps ensure the synthesized and original images are close to each other in pixel space, at the cost of making representation matching more difficult. If you are applying this procedure to new images or new image classification models, you will almost certainly need to experiment to find the appropriate {attr}`~plenoptic.Metamer.penalty_lambda`.
+
+:::
+
++++
+
+We also decreased the learning rate from the default, as this resulted in better solutions in our experiments.
```
billbrod, line 189: The improtant thing for making this an adversarial example is the setting of `initial_image`, so should mention that too

docs/user_guide/reproduce/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+
+While the synthesized image is visually similar to the original image (also note the low MSE reference metric loss), let us verify more rigorously. To do this we subtract the the original image from the synthesized image to visualize the changes in pixel values.
+
+```{code-cell} ipython3
+mse = po.metric.mse(img, met.metamer)
+title = f"Adversarial - Original \nMSE={mse_val:.2e}"
+diff = (
+    met.metamer - img + 1
+) / 2  # convert the range from [-1,1] to [0,1] for RGB images
+po.plot.imshow(diff, as_rgb=True, title=title, col_wrap=2, vrange="auto0");
```
billbrod:
```suggestion
po.plot.imshow(diff, as_rgb=True, title=title);
```

billbrod: if you've remapped the values to lie between 0 and 1, shouldn't use `0` vranges (since there are no negative values)

docs/user_guide/reproduce/adversarial_examples_metamer.md
```diff
@@ -0,0 +1,245 @@
...
+We see the network is highly confident that the synthesized image is a {glue}`category_name`!
+
++++
+
+While the synthesized image is visually similar to the original image (also note the low MSE reference metric loss), let us verify more rigorously. To do this we subtract the the original image from the synthesized image to visualize the changes in pixel values.
+
+```{code-cell} ipython3
+mse = po.metric.mse(img, met.metamer)
+title = f"Adversarial - Original \nMSE={mse_val:.2e}"
+diff = (
```
billbrod: use `po.process.rescale(met.metamer - img)` for this.

billbrod: That function will actually remap the min/max to 0, 1 which is what the text says it's doing, even thought he code is not. if you want to remap to a different range, you can set its `a`/`b` arguments

billbrod: also, might as well assign `met.metamer - img` to a variable, since you use it again in the next block

billbrod (changes requested):
This is great, thanks Raffles! That's a good looking cheeseburger you've found.

I'd like to put both the adversarial example notebooks into a single folder (like the Portilla-Simoncelli ones). So can you create a folder `docs/user_guide/reproduce/adversarial_examples/` and move the two notebooks there, then create an `index.md` page in it, which gives a high-level overview of what adversarial examples are and what the two notebooks are doing (again, similar to the PS ones)?

billbrod:
Oh and for repetition, that's fine, because we want the two notebooks to be stand alone. When you add the index page, you can say a little bit about how some of the sections at the top are repeated, if you like

docs/user_guide/reproduce/adversarial_examples_metamer.md
raffles-xingqi-zhu: I'm not sure if I exactly follow but please check what I did.

raffles-xingqi-zhu: Here is what I think the difference is between the original and the new rescaling. In the original I was scaling the range of _possible_ values to be between 0 and 1 (assuming original range was between -1 and 1). In the new approach you suggested, we are scaling the min and max to be 0 and 1. We see a lot more structure of the noise now. I'm undecided as to what one is the correct method. Thoughts?

raffles-xingqi-zhu:
Thank you for your comments! I've addressed and resolved all except one on the rescaling. After we decide on what to do, I will need to ensure the mad competition notebook is consistent.

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_index.md
```diff
@@ -0,0 +1,13 @@
+# Synthesizing Adversarial Examples
+
+The following notebooks demonstrate how to synthesize adversarial examples using Plenoptic. Adversarial examples are images with subtle, imperceptible perturbations designed to deceive a Deep Neural Networks into making an incorrect classification ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). We show in these notebooks how adversarial examples can be conceptualized and generated as metamers of a different class or images that maximize an optimized metric in representational space, while minimizing changes in pixel space.
```
billbrod:
```suggestion
The following notebooks demonstrate how to synthesize adversarial examples using Plenoptic. Adversarial examples are images with subtle, imperceptible perturbations designed to deceive a Deep Neural Networks into making an incorrect classification ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). In these notebooks, we show two ways to synthesize adversarial examples using plenoptic's machinery:
1. Using {class}`~plenoptic.MADCompetition`, we can synthesize an adversarial example by maximizing the distance in the model's representation while keeping the distance in pixel space small.
2. Using {class}`~plenoptic.Metamer`, we can synthesize an adversarial example by starting with one image and turning it into metamers of a different class.
```
something like that

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -20,14 +20,14 @@ Run it in your browser: **{binder}`adversarial_examples_metamer.ipynb`**!
...
+In this notebook we demonstrate how we can use the {class}`~plenoptic.Metamer` class to synthesize adversarial examples (see [](adversarial-examples-mad) for an alternate approach to synthesizing adversarial examples). Adversarial examples are images with subtle, imperceptible perturbations designed to deceive a Deep Neural Networks into making an incorrect classification ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). In a non-strict sense, an adversarial example is a metamer of the new (misclassified) class because the model has the same classification behaviour for the adversarial image and other images in the class. On this basis, we can use {class}`~plenoptic.Metamer` to generate adversarial examples by matching the model output representation between the synthesized image and another image in that class. For an in-depth dive of creating metamers of Deep Neural Networks, read [](feather2023-resnet50) and [](deep_nets).
```
billbrod: I meant this as a separate line, like:
```suggestion
In this notebook we demonstrate how we can use the {class}`~plenoptic.Metamer` class to synthesize adversarial examples. Adversarial examples are images with subtle, imperceptible perturbations designed to deceive a Deep Neural Networks into making an incorrect classification ({cite:alp}`Szegedy2013`, {cite:alp}`goodfellow_explaining_2015`). In a non-strict sense, an adversarial example is a metamer of the new (misclassified) class because the model has the same classification behaviour for the adversarial image and other images in the class. On this basis, we can use {class}`~plenoptic.Metamer` to generate adversarial examples by matching the model output representation between the synthesized image and another image in that class. For an in-depth dive of creating metamers of Deep Neural Networks, read [](feather2023-resnet50) and [](deep_nets).

See [](adversarial-examples-mad) for an alternate approach to synthesizing adversarial examples using {class}`~plenoptic.MADCompetition`.
```

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -60,10 +60,20 @@ po.set_seed(4)
...
+In the following block, we create a {class}`~plenoptic.models.DeepNetFeatures` model matching the output of `"layer4"` of a classic image recognition network, [ResNet50](https://en.wikipedia.org/wiki/Residual_neural_network), trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/).. After creating the model, we then prepare the original image and an image of a different class whose represetnation we try to match. Finally, we ensure that the model and images have the proper device and dtype, and remove the gradient from all model parameters.
 
 To learn more about any of these steps and why we take them, read [](deep_nets).
 
+As part of this procedure, we must choose an image whose `layer4` representation (and thus, categorization) we would like to match. For this example let us fool the network into thinking a macaque is a cheeseburger (image taken from the Caltech256 dataset {cite:alp}`griffin_caltech_2022`).
```
billbrod:
```suggestion
As part of this procedure, we must choose an image whose `"layer4"` representation (and thus, categorization) we would like to match. For this example let us fool the network into thinking a macaque is a cheeseburger (image taken from the Caltech256 dataset {cite:alp}`griffin_caltech_2022`).
```

should be consistent in how we refer to layer4, and I think 
```
`"layer4"`
```
is the right way

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -60,10 +60,20 @@ po.set_seed(4)
...
+
++++
+
+:::{admonition} Why choose "layer4"?
```
billbrod:
```suggestion
:::{admonition} Why choose `"layer4"`?
```

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -60,10 +60,20 @@ po.set_seed(4)
...
+:class: dropdown hint
+
+Ideally, metamers of an early layer should also be metamers of all subsequent layers. However this is not the case in practise because of the additional constraint on the metamer image we put in pixel space (explained in the following sections). Even when the loss has converged, there will still be some residual error between the original representation and the synthesized representation at the matched layer that can propagate through the network and cause later representation to diverge. By choosing the last convolutional layer of the network ("layer 4"), we minimize this divergence in reprensetation, which gives the network more consistent classification behaviour of the metamer image.
```
billbrod: actually to me the question is -- why not a later layer? in the MADCompetition example, we match the final output, why don't we do that here?

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -113,14 +122,21 @@ def get_category(image):
...
 ```
 
 ResNet50 is trained to classify images into one of [1000 categories](https://deeplearning.cms.waikato.ac.nz/user-guide/class-maps/IMAGENET/). The following plot shows the classification probabilities for the original image as a stem plot. Each of the 1000 categories is represented by a line, whose y-value gives the model's probability that the image belongs to the corresponding category (the x-value is arbitrary). The title shows the label of the most likely category, and the text on the plot shows the other categories with probability higher than 0.01.
 
 ```{code-cell} ipython3
 category_probs, category = get_category(img)
 po.plot.stem_plot(category_probs, title=category)
-likely_cats = "\n- ".join(list(imagenet_categories[category_probs > 0.01]))
+likely_cats = get_likely_categories(category_probs)
 plt.text(700, 0.5, f"Likely categories:\n- {likely_cats}");
```
billbrod: move this text to the right, it's overlapping with the stem

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
billbrod:
```suggestion
The category of our initial image, [guenon](https://en.wikipedia.org/wiki/Guenon), is an Old World monkey. Though it isn't the actual species of the monkey in question (a [Celebes crested macaque](https://en.wikipedia.org/wiki/Celebes_crested_macaque)), it's a reasonable category for it. Notice the model is highly confident in its classification, with a probability of about 0.8 and no other category exceeding a probability of 0.1. The other predicted categories are all [Old World monkeys](https://en.wikipedia.org/wiki/Old_World_monkey). 

Now let us make the same plot for the target image.
```

billbrod:
```suggestion
Putting it all together, the goal of the synthesis procedure in this notebook is to create an image that looks like our original macaque image while being classified as a cheeseburger.
```

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -157,10 +177,10 @@ met = po.Metamer(
...
+The objective function for metamer synthesis is made of two parts: the synthesis loss that measures the difference in representation between synthesized and target images, and the penalty. As {attr}`~plenoptic.Metamer.penalty_lambda` increases, the relative weight of the penalty in the objevtive increases. This helps ensure the synthesized and original images are close to each other in pixel space, at the cost of making representation matching more difficult. From our experiments, if you set {attr}`~plenoptic.Metamer.penalty_lambda` to a small value (e.g. 1), the penalty was essentially ignored and the resulting image was very distant from the original in pixel space, breaking the first requirement of an adversarial example. If you are applying this procedure to new images or new image classification models, you will almost certainly need to experiment to find the appropriate {attr}`~plenoptic.Metamer.penalty_lambda`.
```
billbrod:
```suggestion
The objective function for metamer synthesis is made of two parts: the synthesis loss that measures the difference in representation between synthesized and target images, and the penalty. As {attr}`~plenoptic.Metamer.penalty_lambda` increases, the relative weight of the penalty in the objective increases. This helps ensure the synthesized and original images are close to each other in pixel space, but it makes matching the target image's representation more difficult. From our experiments, if you set {attr}`~plenoptic.Metamer.penalty_lambda` to a small value (e.g. 1), the penalty was essentially ignored and the resulting image was very distant from the original in pixel space, breaking the first requirement of an adversarial example. 

If you are applying this procedure to new images or new image classification models, you will almost certainly need to experiment to find the appropriate {attr}`~plenoptic.Metamer.penalty_lambda`.
```

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
billbrod:
```suggestion
met.synthesize(store_progress=True, max_iter=1000)
```

given that it only runs for ~600 iterations

billbrod:
```suggestion
Everything has been set up, so now let's synthesize the adversarial example!
```

billbrod, line 200: should describe what's in the plot:
- image looks like the initial one
- loss has decreased and looks like it has converged
- as I said in the deep net metamer one: doesn't look like there are any hotspots in the error, either spatially or in a specific channel

and then say something like: "however, we can't tell what the classification is from this plot. we'll investigate that in the next section."

billbrod:
```suggestion
First let us visualize the category probabilities of the original, target, and adversarial images using the same stem plots of categorization probability that we used above.
```

billbrod: describe what's happening in the plots in more detail: top row shows ..., middle row shows ..., etc. title of the images show ..., stem plots show ...., the text to the right is...

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -220,22 +240,21 @@ While the synthesized image is visually similar to the original image (also note
...
-    met.metamer - img + 1
-) / 2  # convert the range from [-1,1] to [0,1] for RGB images
-po.plot.imshow(diff, as_rgb=True, title=title, col_wrap=2, vrange="auto0");
+diff = met.metamer - img
+diff_rescaled = po.process.rescale(diff)  # rescale to values between 0 and 1
+po.plot.imshow(diff_rescaled, as_rgb=True, title=title);
 ```
 
-The difference between synthesized and original images looks like random pixel noise distributed across the image. However, the noise is too faint for us to tell if there's any structure. Instead, let us try visualizing the difference in each color channel (red, green, and blue) separately.
+The difference between synthesized and original images looks like random pixel noise distributed across the image. Let us try visualizing the difference in each color channel (red, green, and blue) separately.
```
billbrod:
```suggestion
The difference between synthesized and original images looks like random pixel noise distributed across the image. Let us try visualizing the difference in each color channel (red, green, and blue) separately, to see if there's any structure hiding there.
```

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -220,22 +240,21 @@ While the synthesized image is visually similar to the original image (also note
...
 titles = [
     "Adversarial - Original (R)",
     "Adversarial - Original (G)",
     "Adversarial - Original (B)",
 ]
-po.plot.imshow(channelwise_diffs_met, col_wrap=3, title=titles, vrange="auto0");
+po.plot.imshow(diff, col_wrap=3, title=titles, vrange="auto0");
 ```
 
 We see the difference does not look like a {glue}`category_name`, or similar. At this point we can safely conclude the synthesized image is an adversarial example of the network.
```
billbrod:
```suggestion
We see the difference does not look like a {glue}`category_name` or similar, either in any of the individual channels are when combined. At this point we can safely conclude the synthesized image is an adversarial example of the network.
```

docs/user_guide/reproduce/adversarial_examples_metamer.md
billbrod: Ah I see, that makes more sense now.

We discussed this in person and agreed that you should rescale the image as you were originally doing (that is, remap the domain of possible values to 0, 1) and then call `po.plot.imshow` with the default vrange (`indep1`) so that the *pattern* of the distortion is as visible as it is in the current figure, but that the *magnitude* is displayed in the title (both in the printed vrange and the MSE, as you do now). Then, add an admonition underneath to describe what's going on with the remapping and how to undo it, to make it clear that you can't just add that figure back to `img` to recover the adversarial image, you need to do `img + 2*diff - 1` or `img + 2 * (diff - 0.5)`)

billbrod (changes requested):
Out of curiosity, what happens in this synthesis without the penalty? What does it look like?

Otherwise, just some tweaking to wording here. This is almost done!

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
raffles-xingqi-zhu: Thanks for bringing this up! I almost forgot to mention this. It's because I was getting an error from `po.plot.synthesis_status` when I used the last fully connected layer. "UnboundLocalError: cannot access local variable 'height_ratios' where it is not associated with a value" is what I got when I ran it just now. I was debugging it for a while and couldn't figure out at the time so I used the last conv layer. We should look into this though.

billbrod: Oh that sounds like  a bug and shouldn't happen. If you can send me a code snippet to reproduce it, I'll look into and fix it.

But I'm surprised, because isn't the last fully connected layer what we use in the mad competition adversarial example? and you were able to use `po.plot.synthesis_status` for that one?

billbrod: Still want to figure this out, I'd prefer to have this comment address "why not a later layer" rather than "why not an earlier layer"

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -192,16 +198,20 @@ We also decreased the learning rate from the default, as this resulted in better
...
-We let the optimization run until loss converges by setting `max_iter` to a large value.
+Everything has been set up, so now let's synthesize the adversarial example!
 
 ```{code-cell} ipython3
-met.synthesize(store_progress=True, max_iter=10000)
+met.synthesize(store_progress=True, max_iter=1000)
 po.plot.synthesis_status(met);
 ```
 
+It looks like the representation error has converged and yet the synthesized metamer still looks similar to the original image. It doesn't look like there are any hotspots in the error, either spatially or in a specific channel. However, we can't tell what the classification is from this plot. We will investigate that in the next section.
```
billbrod:
```suggestion
From the middle axis in the above plot, we can see that the loss has converged. The right-most plot shows the difference in the model's `"layer4"` representation between the metamer and the target cheeseburger image; it doesn't look like there are any hotspots in the error, either spatially or in a specific channel. The synthesized metamer, shown on the left, still looks like our initial macaque image. However, we can't tell what the classification is from this plot. We will investigate that in the next section.
```

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -231,21 +241,21 @@ category_probs, category = get_category(met.metamer)
 glue("category_name", str(category), display=False)
 ```
 
-We see the network is highly confident that the synthesized image is a {glue}`category_name`!
+In the top row, we have the original image on the left and the stem plot on the right. The middle and bottom rows correspond to the target image and the adversarial example, togeher with their stem plots. On top of each image, we show the MSE from the original image. The texts next to the stem plots show the most likely categories. We see the network is highly confident that the synthesized image is a {glue}`category_name`! We also see the MSE of the synthesized image is much smaller than the MSE of the target image, despite both being classified as a {glue}`category_name`.
```
billbrod:
```suggestion
In the top row, we have the original image on the left and the stem plot on the right. The middle and bottom rows correspond to the target image and the adversarial example, together with their stem plots. On top of each image, we show the MSE from the original image. The texts next to the stem plots show the most likely categories. We see the network is highly confident that the synthesized image is a {glue}`category_name`! We also see the MSE between the original and adversarial images is quite small, much smaller than that of the target image, despite both being classified as a {glue}`category_name`.
```

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -231,21 +241,21 @@ category_probs, category = get_category(met.metamer)
...
 
 +++
 
-While the synthesized image is visually similar to the original image (also note the low MSE reference metric loss), let us verify more rigorously. To do this we subtract the the original image from the synthesized image to visualize the changes in pixel values.
+While the synthesized image is visually similar to the original image (note the low MSE with the original), let us subtract the the original image from the synthesized image to visualize the changes in pixel values. Since individual synthesized image pixel values could have either increased or decreased, the difference image contains negative values. To better visualize the difference image, we therefore rescale the pixel values to be between 0 and 1, which correspond to the min and max of the original difference image.
```
billbrod:
```suggestion
While the synthesized image is visually similar to the original image (note the low MSE with the original), let us subtract the the original image from the synthesized image to visualize the changes in pixel values. 

:::{admonition} Why are we rescaling the difference image?
:class: dropdown question
Since the individual image pixel values could have either increased or decreased over synthesis, the difference image contains both positive and negative values. However, visualizing RGB images requires all pixel values to lie between 0 and 1 (see {external+matplotlib:func}`matplotlib.pyplot.imshow` for more details). Therefore, we must rescale the pixel values' to lie between 0 and 1; we additionally use {func}`~plenoptic.process.rescale` to remap the minimum and maximum to be 0 and 1, respectively, to make the structure of the difference image more visible.
:::
```

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -257,7 +267,7 @@ titles = [
 po.plot.imshow(diff, col_wrap=3, title=titles, vrange="auto0");
 ```
 
-We see the difference does not look like a {glue}`category_name`, or similar. At this point we can safely conclude the synthesized image is an adversarial example of the network.
+We see the difference does not look like a {glue}`category_name` or similar, either in any of the individual channels are when combined. At this point we can safely conclude the synthesized image is an adversarial example of the network.
```
billbrod:
```suggestion
We see the difference does not look like a {glue}`category_name` or similar, either in any of the individual channels or when combined. At this point we can safely conclude the synthesized image is an adversarial example of the network.
```

billbrod (commented):
Couple more small changes noted here, also: 
- Make the same change to the first paragraph (separating out "See Metamer notebook for an alternate approach...")  in the adversarial example MAD notebook
- Out of curiosity, what happens in this synthesis without the penalty? What does it look like?

billbrod:
I accidentally made the above a "Comment", but it should be a "Request changes".

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
raffles-xingqi-zhu: To reproduce the bug, simply set `target_layer` to `"fc"`. And you are right, there was no bug when I used `po.plot.synthesis_status` for that one. Let me know if you are able to reproduce the bug.

raffles-xingqi-zhu:
Thanks Billy. Responses to your requests/questions: 

1. In addition to the cross reference statement, I also harmonized the rescaling and the corresponding descriptions in MAD competition notebook to match the metamer notebook.
2. When I changed `penalty_lambda` to 0, the synthesized image looks visibly different form the original with MSE error jumping two orders of magnitude compared to `penalty_lambda=1000` . I also tried the default synthesis set up with only penalty on the pixel range and got the same result. I remember doing a hyper parameter search using lower `penalty_lambda` values and reached the same conclusion. These tell me the custom penalty imposing the MSE to be small in pixel space is necessary for successful adversarial example synthesis. 
3. For deciding whether the admonition should address why not earlier or later layers, I think we need to figure out the bug first (see my response to the original comment for you to reproduce it).

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
billbrod: Fixed this in #480 , let me know what it looks like when you give it a try.

billbrod (changes requested):
The text here all looks good to me! Hopefully the fix in #480 works for you -- can you try doing this while matching the `"fc"` layer and see what that looks like? If it works, then let's go with that.

Also: the MAD adversarial example category has changed, to `malinois` (a dog), rather than cheeseburger. Why would it change?
    - Also, the name got broken up in that plot, so there's an issue with the output of `get_likely_categories` if it only returns a single category

billbrod:
Oh one last thing -- you're setting `store_progress=True` in this example, but we don't do anything with that, so remove it.

raffles-xingqi-zhu:
Thank you for the suggestions. I made changes. I have no idea why the MAD competition notebook gives a different classification now. That's weird as I didn't make changes to the code, only the explanation.

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_mad.md
```diff
@@ -114,8 +114,8 @@ def get_category(image):
 def get_likely_categories(category_probs):
     likely_idx = torch.where(category_probs > 0.01)[0]
     likely_idx = likely_idx[torch.argsort(category_probs[likely_idx], descending=True)]
-    likely_cats = "\n- ".join(imagenet_categories[likely_idx].tolist())
-    return likely_cats
+    likely_cats = imagenet_categories[likely_idx].tolist()
+    return "\n".join(f"- {cat}" for cat in likely_cats)
```
billbrod, line 120: the problem here is with the iteration. `likely_cats` here (with or without `tolist()`) can be either a list of strings or a single string. If it's a single string, then your iteration through it will pull apart the individual characters in the string (as we're seeing now). You could check by using `if isinstance(likely_cats, str)` and the converting it to a list (`likely_cats = [likely_cats]`). Explicitly checking the typing isn't great in python (as opposed to duck-typing), but it's hard to disentangle a single string and a list of strings (both are iterables that may have any non-negative length).

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_metamer.md
```diff
@@ -62,18 +62,18 @@ po.set_seed(4)
...
 
 To learn more about any of these steps and why we take them, read [](deep_nets).
 
-As part of this procedure, we must choose an image whose `"layer4"` representation (and thus, categorization) we would like to match. For this example let us fool the network into thinking a macaque is a cheeseburger (image taken from the Caltech256 dataset {cite:alp}`griffin_caltech_2022`).
+As part of this procedure, we must choose an image whose `"fc"` representation (and thus, categorization) we would like to match. For this example let us fool the network into thinking a macaque is a cheeseburger (image taken from the Caltech256 dataset {cite:alp}`griffin_caltech_2022`).
 
 +++
 
-:::{admonition} Why choose `"layer4"`?
+:::{admonition} Would matching other layers also work?
```
billbrod: I think we can remove this block now -- for the purposes of generating adversarial examples, focusing on the fully connected layer is the intuitive thing to do, and so I don't think we need to explain more. (if you're targeting a different layer, it no longer is as obviously an adversarial example to me, but I think that's beyond the scope of this tutorial)

billbrod (changes requested):
I figured out the issue with the MAD Competition changing categories: it's because we need to set `torch.use_deterministic_algorithms` at the top (see my discussion [here](https://docs.plenoptic.org/docs/pulls/476/reference/reproducibility.html#reprodubility)); convolutions are, by default, non-deterministic on GPUs. I can do that in a follow-up PR, since it's orthogonal to this discussion here (I will also update the PR template to add a reminder here, since I should've remembered this, it came up in #460 earlier this summer).

I'm glad that the `"fc"` layer works well now! Couple of comments below and then we can merge this

docs/user_guide/reproduce/synthesizing_adversarial_examples/adversarial_examples_mad.md
billbrod: And ensure whatever change you make here works before pushing it

raffles-xingqi-zhu:
Both are great points. I fixed those. 

Should I try using `plenoptic.set_seed` function? Would that make a difference?

billbrod (approved):
This is good to go, thanks for all your work Raffles! I'll go ahead and merge this in.

> Should I try using plenoptic.set_seed function? Would that make a difference?

You already are using that function. That just sets the seed of the RNG process (for both torch and numpy). The main thing this affects is the random noise we initialize the synthesis process with. It unfortunately doesn't seem to have any effect on these non-deterministic algorithms (see [pytorch docs](https://docs.pytorch.org/docs/2.13/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms) for details)

PR #481, opened 2026-08-18
Adds Tutorials+Exercises, `MAX_ITER`, fix run notebooks in deploy action

billbrod:

After my presentation at Cold Spring Harbor this summer, realized that I should probably include the content from my [workshops](https://workshops.plenoptic.org/) in the documentation, since they may be helpful to people

This PR ports over the "introductory tutorial" (which I normally live-code or talk people through) and the exercises to the documentation. I excluded the Nemos GLM exercise, as that requires a bunch of extra dependencies (jax, nemos, pynapple).

Additionally, this PR adds the use of a `MAX_ITER` environmental variable to control the duration of some of the notebooks (as well as an explanation in contributing). In particular the new `introductory_tutorial` notebook took 2 hours on the CPU if run to completion, so this drastically shortens that.

- ~I have placed them in a new section of the docs called, "Tutorials + Exercises". I have linked to this from the getting started section as well. Does that location make sense?~
    - After thinking about this some more, I moved them instead to "Getting started", since I think that's where they belong. Added more links to them from other places in the docs.
- I think people will get the most out of these notebooks if they go through them themselves, either by opening a new notebook and following along, or downloading the existing notebook. Is the way I discussed this (~both in `tutorials/index.md` and~ at the top of each notebook) clear?
- The introductory tutorial has some of the `render` tags/classes that we use for generating different versions of the notebook for workshops. These aren't used here, but they remain in the document so I can start from this version when creating the material for a new workshop. I added a little description of this at the top of the notebook (within one of those hidden render-user divs). I don't think there's a better way to do this, right?
- Is my usage of `MAX_ITER` reasonable? In other long-running notebooks, I have cached and retrieved the results of synthesis, but the `introductory_tutorial` notebook here takes very little time on the GPU, so that didn't seem as necessary there.
    - I could also have just made the number of iterations depend on whether a GPU is available or not, but this feels a little cleaner to me. In particular, the deep net notebooks run in ~10 minutes on the CPU, which is fine for someone to download and run themselves, but there's no reason not to speed it up for the CI.

**Describe changes**

- ~Adds new "Tutorials + Exercises" section to docs, containing the ported notebooks.~ Adds these pages to the "Getting Started" section of the docs. 
- Makes sure those notebooks are compliant with all our sanity checks.
- Increases the duration of synthesis in the introductory notebook to make metamers look better. 
- Changes to `linting/check_tutorials.py` to allow for these notebooks to have a different "download" admonition at the top.
- ~Small changes to layout to look better with this new tab: replace search field with search button.~ May bring this back at one point, but no longer needed.
- Realized that deploy action had a section that was "running notebook" which was going through `docs/tutorials` (which no longer exists, holdover from #410). updated it to match the corresponding block from `ci.yml`
- Added a custom dictionary for `codespell` which I've added common mispellings of plenoptic, will keep updating as I go.
    - Added a section to PR template about typos.

**Checklist**


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/481

docs/tutorials/index.md
```diff
@@ -0,0 +1,34 @@
+(tutorials-exercises)=
+# Tutorial + Exercises
```
sjvenditto: Do you want "Tutorials + Exercises" or "Tutorial + Exercises"? In the PR description you use the former, but the PR title and here use the latter. I like the first one better, especially if you add more tutorials in the future?

sjvenditto (approved):
- The section and placement all make sense to me, and it's clear in each notebook how you expect the user to do the exercises themselves.
- I think leaving in the div tags is also fine, and the hidden description is clear on why they are there
- I also like the use of MAX_ITER since it allows you to change it selectively and leaves the default for anyone who downloads it.

My only main comment is whether you wanted to use "Tutorial + Exercises" or "Tutorials + Exercises" ? I left a little blurb in the code

BalzaniEdoardo:
> **Describe the change in this PR at a high-level**
> 
> After my presentation at Cold Spring Harbor this summer, realized that I should probably include the content from my [workshops](https://workshops.plenoptic.org/) in the documentation, since they may be helpful to people
> 
> This PR ports over the "introductory tutorial" (which I normally live-code or talk people through) and the exercises to the documentation. I excluded the Nemos GLM exercise, as that requires a bunch of extra dependencies (jax, nemos, pynapple).
> 
> Additionally, this PR adds the use of a `MAX_ITER` environmental variable to control the duration of some of the notebooks (as well as an explanation in contributing). In particular the new `introductory_tutorial` notebook took 2 hours on the CPU if run to completion, so this drastically shortens that.
> 
> **Outstanding questions / particular feedback**
> 
>     * I have placed them in a new section of the docs called, "Tutorials + Exercises". I have linked to this from the getting started section as well. Does that location make sense?

I think this makes sense.

>     * I think people will get the most out of these notebooks if they go through them themselves, either by opening a new notebook and following along, or downloading the existing notebook. Is the way I discussed this (both in `tutorials/index.md` and at the top of each notebook) clear?

yes, I think that's the best you can do.

>     * The introductory tutorial has some of the `render` tags/classes that we use for generating different versions of the notebook for workshops. These aren't used here, but they remain in the document so I can start from this version when creating the material for a new workshop. I added a little description of this at the top of the notebook (within one of those hidden render-user divs). I don't think there's a better way to do this, right?

I think that's fine since it is not rendered.

>     * Is my usage of `MAX_ITER` reasonable? In other long-running notebooks, I have cached and retrieved the results of synthesis, but the `introductory_tutorial` notebook here takes very little time on the GPU, so that didn't seem as necessary there.

That looks good to me.

>       * I could also have just made the number of iterations depend on whether a GPU is available or not, but this feels a little cleaner to me. In particular, the deep net notebooks run in ~10 minutes on the CPU, which is fine for someone to download and run themselves, but there's no reason not to speed it up for the CI.
> 

The CPU/GPU  gating seems like an overkill

billbrod:
Thanks! I decided to move the Tutorials+Exercises to the Getting started page, because I think that's where they belong (and added in some cross-links elsewhere in the docs). What do you think?

One of the reasons I'm moving it is because I'm thinking about adding a top-level section for "How-To Guides", brief examples of how to do things (like we recently discussed for nemos). That should definitely be top-level, but I don't want to have too many top-level items. So I decided to demote this

sjvenditto:
I like the content better in the Getting Started section as you have it now. I have a few outstanding comments:
- I wonder if the exercises should be moved after the installation portion? I don't feel strongly about this, it more comes down to which piece makes more sense being surfaced first
- In the Quickstart you mention "See the other tutorials for more details." Do you want to link specifically to the introductory tutorial? 
- I noticed that the "Getting Started" page doesn't have any textual reference to the installation pages. Do you want to add a reference and link after the quick install to point to more detailed installation instructions?

billbrod:
> I like the content better in the Getting Started section as you have it now. I have a few outstanding comments:
> 
>     * I wonder if the exercises should be moved after the installation portion? I don't feel strongly about this, it more comes down to which piece makes more sense being surfaced first
> 
>     * In the Quickstart you mention "See the other tutorials for more details." Do you want to link specifically to the introductory tutorial?
> 
>     * I noticed that the "Getting Started" page doesn't have any textual reference to the installation pages. Do you want to add a reference and link after the quick install to point to more detailed installation instructions?

I agree with all these suggestions and have made them, thanks!

sjvenditto (approved):

PR #482, opened 2026-08-18
Adds Color image `PortillaSimoncelli` model

dherrera1911:

This PR adds the color Portilla-Simoncelli model into the existing `PortillaSimoncelli()` class. It follows [Vacher & Briand, Image Processing On Line (2021)](https://www.ipol.im/pub/art/2021/324/) appendix B2, which offers a description of the original color Portilla-Simoncelli model.

The color model has three main components compared to the standard model:
1. A color-space transform that maps RGB images to a different color space
2. Extension of existing statistics families to include cross-channel correlations
3. New statistics families not present in the original Portilla-Simoncelli model

The `PortillaSimoncelli()` class now takes two new parameters, `color_statistics: bool` to indicate whether to use the extended set of color statistics, and `transform: Callable` containing the color-space transform.

A minimal example of the new API:
```python
import torch
import plenoptic as po
import matplotlib.pyplot as plt

img_color = po.process.center_crop(po.data.color_wheel(as_gray=False)[:,:,:512,:512], 128)

pca = po.process.PCA(img_color) # Generate color-space PCA transform
model_color = po.models.PortillaSimoncelli(
  img_color.shape[-2:],
  color_statistics=True,  # Flag to use color model
  transform=pca,          # Color space transform
)
met_color = po.Metamer(img_color, model_color)
met_color.synthesize(max_iter=50)

po.plot.imshow(
  [img_color, met_color.metamer], title=["Target", "PCA color metamer"], as_rgb=True
)
plt.show()
```
I'll add more specific examples in the comments below.

Towards #46.

- [We were wondering](https://github.com/plenoptic-org/plenoptic/issues/46#issuecomment-5217918925) whether to have a separate class for the color Portilla-Simoncelli. I think that I managed to fit it into the existing class tidily, so I favor the single class as implemented here.
- I set the OPC color transform discussed in #46 as the default value of `transform`. When `color_statistics = False` the transform is ignored, maintaining default behavior. I anticipate @billbrod might not like this default, and prefer having no transform as default. My argument for it is that there is value in providing reasonable defaults for the user, even if they could set them themselves.
- I haven't checked yet whether `portilla_simoncelli_loss_factory` works with the color model, it might need tweaking.
- Plotting the color model representation is not implemented. I would rather leave for future PR.
- The new model is not yet added to the docs. I propose that: 1) We review the implementation, 2) We delineate what the new docs about color Portilla-Simoncelli would include, 3) I implement the new docs/tutorial. This might also require adding a color texture image to `plenoptic`.
- Similarly, I need to add docstring examples.
- Do we want to change the name of `po.process.PCA` to `po.process.ColorPCA` or such?
- Do we need more checks for the color transform inside `PortillaSimoncelli()`?
- Initialization with an image that has similar RGB pixel statistics to the original is important. Should we add helpers for this?
- We currently require that the inputs to color `PortillaSimoncelli()` have 3 color channels. We could accept arbitrary channels, but maybe I would leave that for a future PR.
- Should we not ignore `transform` when `color_statistics=False`? This would only make sense if we change the default to the identity transform. It would enable using the independent-channel model with a color transform. But not sure this is important, the user can transform the image outside `PortillaSimoncelli()` if they wanted to do this.

**Describe changes**

- Added the module `process.color` with the color-space transforms `OPC()` and `PCA()` that can be passed to `PortillaSimoncelli()`
- Added explicit color model path to `PortillaSimoncelli()` via flag `color_statistics`
- Added `transform` argument to `PortillaSimoncelli()`, that transforms the input image color space (ignored when `color_statistics=False`)
- The same attributes `self._necessary_stats_dict` and `self._representation_scales` are used by the independent-channel and the color model. The color model extends them as needed.
- The representation tensor for the independent-channel model has shape `(batch, n_channels, n_stats)`, but the color model has `(batch, 1, n_stats)`, because it no longer makes sense to separate stats by channel.
- As described in Vacher 2021 appendix B2 bullet point 2, cross correlations families `"cross_orientation_correlation_magnitude", `"cross_scale_correlation_magnitude"`, `"cross_scale_correlation_real"` are extended to include correlations across (transformed) channels.
- As described in Vacher 2021 extra statistics are added to the model: (vii) color covariance in input color space, (viii) auto-correlation of transformed image channels,  (ix) skewness and kurtosis of each transformed channel, (x) cross-correlation of real coeffs across channel-orientation for each scale, (xi) cross-correlation of low pass residual, across image shifts and channels, (xii) cross-correlation of real coeffs and low pass residuals across different shifts and channels.
- Extended several existing tests to include the color model
- Added tests checking that some of the new statistics have the expected shapes.
- The `test_scales_shapes()` was improperly handling the case with `n_scales=1`, where some statistics are empty. Can't remember the details 100%, but I think it was zipping dicts of different length, and it passed out of luck. This PR also fixes that. 

**Checklist**

- [ ] I have added any necessary documentation. This includes docstrings, updates to existing files found in `docs/`, or (for large changes) adding new files to the `docs/` folder.
**Daniel:** I added docstrings (see if some need improvement), but not documentation yet, as described above.

**AI disclaimer**

I used coding agents for this contribution. The major design decisions (API and internals) are almost entirely my own. I either wrote docstrings myself or edited almost all AI-generated docstrings, and the ones I didn't modify I read and didn't find anything requiring editing. I reviewed all AI generated code, and often edited for better design, readability or correctness.

### DIFF
--- a/docs/api/process.rst
+++ b/docs/api/process.rst
@@ -110,6 +110,19 @@ The following functions return a modified version of their input.
    ~center_crop
    ~modulate_phase
 
+.. rubric:: Color transformations
+   :heading-level: 3
+
+These functions transform the color space of RGB images.
+
+.. autosummary::
+   :signatures: none
+   :toctree: generated
+   :template: torch_module.rst.jinja
+
+   ~OPC
+   ~PCA
+
 .. rubric:: Image statistics
    :heading-level: 3
 
--- a/src/plenoptic/_api_change.py
+++ b/src/plenoptic/_api_change.py
@@ -193,6 +193,8 @@ NEW = [
     "plenoptic.io.LoadWarning",
     "plenoptic.data.macaque",
     "plenoptic.models.DeepNetFeatures",
+    "plenoptic.process.OPC",
+    "plenoptic.process.PCA",
 ]
 
 UNCHANGED = [
--- a/src/plenoptic/models/portilla_simoncelli.py
+++ b/src/plenoptic/models/portilla_simoncelli.py
@@ -451,6 +451,11 @@ class PortillaSimoncelli(nn.Module):
             "residual_lowpass",
             dtype=object,
         )
+        color_shape_dict["cross_correlation_coarsest_scale_lowpass"] = np.full(
+            (N_RGB_CHANNELS * self.n_orientations, n_lowpass_signals),
+            self.n_scales - 1,
+            dtype=int,
+        )
         return color_shape_dict
 
     def _create_necessary_stats_dict(
@@ -617,6 +622,9 @@ class PortillaSimoncelli(nn.Module):
                     lowpass_mask[row, column] = True
                     seen_correlations.add(correlation)
         color_mask_dict["cross_correlation_lowpass"] = lowpass_mask
+        color_mask_dict["cross_correlation_coarsest_scale_lowpass"] = torch.ones(
+            n_signals, n_lowpass_signals, dtype=torch.bool
+        )
         return color_mask_dict
 
     @staticmethod
@@ -798,6 +806,11 @@ class PortillaSimoncelli(nn.Module):
             cross_corr_lowpass = self._compute_joint_cross_correlation(
                 [lowpass_signals], [lowpass_signals]
             ).squeeze(-1)
+            cross_corr_coarsest_scale_lowpass = (
+                self._compute_joint_cross_correlation(
+                    [real_pyr_coeffs[-1]], [lowpass_signals]
+                ).squeeze(-1)
+            )
         else:
             cross_ori_corr_mags = self._compute_cross_correlation(
                 mag_pyr_coeffs, mag_pyr_coeffs, mags_var, mags_var
@@ -862,6 +875,7 @@ class PortillaSimoncelli(nn.Module):
                 kurtosis_transformed,
                 cross_ori_corr_real,
                 cross_corr_lowpass,
+                cross_corr_coarsest_scale_lowpass,
             ]
         # And then pack them into a 3d tensor
         if self.color_statistics:
--- a/src/plenoptic/models/portilla_simoncelli.py
+++ b/src/plenoptic/models/portilla_simoncelli.py
@@ -620,7 +620,8 @@ class PortillaSimoncelli(nn.Module):
             ):
                 # Compute relative shift, and disambiguate
                 relative_shift = (
-                  column_shift[0] - row_shift[0], column_shift[1] - row_shift[1]
+                    column_shift[0] - row_shift[0],
+                    column_shift[1] - row_shift[1],
                 )
                 if row_channel == column_channel:
                     opposite_shift = tuple(-offset for offset in relative_shift)
@@ -822,11 +823,9 @@ class PortillaSimoncelli(nn.Module):
             cross_corr_lowpass = self._compute_joint_cross_correlation(
                 [lowpass_signals], [lowpass_signals]
             ).squeeze(-1)
-            cross_corr_coarsest_scale_lowpass = (
-                self._compute_joint_cross_correlation(
-                    [real_pyr_coeffs[-1]], [lowpass_signals]
-                ).squeeze(-1)
-            )
+            cross_corr_coarsest_scale_lowpass = self._compute_joint_cross_correlation(
+                [real_pyr_coeffs[-1]], [lowpass_signals]
+            ).squeeze(-1)
         else:
             cross_ori_corr_mags = self._compute_cross_correlation(
                 mag_pyr_coeffs, mag_pyr_coeffs, mags_var, mags_var
--- a/src/plenoptic/models/portilla_simoncelli.py
+++ b/src/plenoptic/models/portilla_simoncelli.py
@@ -41,7 +41,7 @@ from ..tensors import to_numpy
 
 SCALES_TYPE = Literal["pixel_statistics"] | PYR_SCALES_TYPE
 N_RGB_CHANNELS = 3
-LOWPASS_SHIFTS = ((0, 0), (2, 0), (-2, 0), (0, 2), (0, -2))
+LOWPASS_SHIFTS = ((0, 0), (0, 2), (0, -2), (2, 0), (-2, 0))
 
 __all__ = [
     "PortillaSimoncelli",
--- a/src/plenoptic/models/portilla_simoncelli.py
+++ b/src/plenoptic/models/portilla_simoncelli.py
@@ -410,6 +410,7 @@ class PortillaSimoncelli(nn.Module):
             if key in [
                 "cross_orientation_correlation_magnitude",
                 "cross_scale_correlation_magnitude",
+                "cross_scale_correlation_real",
             ]:
                 # Add the joint-channel statistics scales
                 color_shape_dict[key] = np.tile(
@@ -537,7 +538,10 @@ class PortillaSimoncelli(nn.Module):
                 )
                 mask[triu_inds[0], triu_inds[1]] = False
                 color_mask_dict[key] = mask
-            elif key == "cross_scale_correlation_magnitude":
+            elif key in [
+                "cross_scale_correlation_magnitude",
+                "cross_scale_correlation_real",
+            ]:
                 # Cross-scale correlations are not symmetric, so keep all entries.
                 color_mask_dict[key] = value.repeat(N_RGB_CHANNELS, N_RGB_CHANNELS, 1)
             else:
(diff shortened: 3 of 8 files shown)


dherrera1911:
Below I report two experiments with the new model.

First, I compare the metamers for an image of green and red apples, synthesized by the color model with PCA transform, the original model with independent channels, and the color model with OPC transform:

```python
import matplotlib.pyplot as plt
import torch
import plenoptic as po

DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
DTYPE = torch.float32
MAX_ITER = 300

optimizer_kwargs = {
    "max_iter": 2,
    "max_eval": 4,
    "history_size": 10,
    "line_search_fn": "strong_wolfe",
    "lr": 1.0,
}
loss = po.loss.l2_norm

# Load the image
img = po.load_images("./Apples_128.png", as_gray=False).to(device=DEVICE, dtype=DTYPE)

##### NEW COLOR MODEL ########
po.set_seed(0)

pca = po.process.PCA(img)

color_ps = po.models.PortillaSimoncelli(
  img.shape[-2:], color_statistics=True, transform=pca
)
color_met = po.Metamer(
  img, color_ps, loss_function=loss
)
color_met.setup(
  optimizer=torch.optim.LBFGS,
  optimizer_kwargs=optimizer_kwargs
)
color_met.synthesize(max_iter=MAX_ITER)

##### INDEPENDENT CHANNELS (OLD) MODEL ########
po.set_seed(0)

independent_ps = po.models.PortillaSimoncelli(
  img.shape[-2:]
)
indep_met = po.Metamer(
  img, independent_ps, loss_function=loss
)
indep_met.setup(
  optimizer=torch.optim.LBFGS, optimizer_kwargs=optimizer_kwargs
)
indep_met.synthesize(
  max_iter=MAX_ITER
)

##### OPC MODEL ########
po.set_seed(0)

# OPC color transform is the default when `color_statistics=True`.
opc_ps = po.models.PortillaSimoncelli(
  img.shape[-2:], color_statistics=True
)
opc_met = po.Metamer(
  img, opc_ps, loss_function=loss
)
opc_met.setup(
  optimizer=torch.optim.LBFGS, optimizer_kwargs=optimizer_kwargs
)
opc_met.synthesize(
  max_iter=MAX_ITER
)

po.plot.imshow(
  [img, color_met.metamer, indep_met.metamer, opc_met.metamer],
  title=["Target", "PCA color metamer", "Independent channel metamer", "OPC color metamer"],
  as_rgb=True
)
plt.show()
```
apples_metamers

Results are pretty good for the color model, and poor for the independent-channel model. Both the PCA and OPC transforms perform well. Both have some high-frequency noise. I feel that has less extreme pixels, but more noise overall, while OPC has some smooth regions and some very off pixels. In other synthesis examples the OPC was noticeably worse.

The second example is for an image of beans with one single dominant color. This example surfaced a problem when optimizing the color model with PCA. This type of image has very low variability in one axis of color space, so PCA whitening strongly scales that (usually low-information) axis. For a metamer's random initialization, this whitening transform will highly scale the low-information color axis and make it dominate optimization. One way to help this is to initialize synthesis with an image with similar color statistics to the target. In the example below, I initialize the metamer to a version of the target with pixel locations scrambled (but keeping each pixel color). I also use a color PCA regularization that limits the maximum upscaling that whitening can apply:

```python
import matplotlib.pyplot as plt
import torch
import plenoptic as po

DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
DTYPE = torch.float32
MAX_ITER = 400

optimizer_kwargs = {
    "max_iter": 2,
    "max_eval": 4,
    "history_size": 10,
    "line_search_fn": "strong_wolfe",
    "lr": 1.0,
}
loss = po.loss.l2_norm

# Load the image
img = po.load_images("./beans_128.png", as_gray=False)

##### COLOR MODEL, RANDOM INIT, NO CAP ########
po.set_seed(0)

pca = po.process.PCA(img)

color_ps = po.models.PortillaSimoncelli(
  img.shape[-2:], color_statistics=True, transform=pca
)
color_met = po.Metamer(
  img, color_ps, loss_function=loss
)
color_met.setup(
  optimizer=torch.optim.LBFGS,
  optimizer_kwargs=optimizer_kwargs
)
color_met.synthesize(
  max_iter=MAX_ITER
)


##### COLOR MODEL, COLOR INIT, NO CAP ########

# Randomize target image pixels
n_pixels = img.shape[-2] * img.shape[-1]
perm = torch.randperm(n_pixels, device=img.device)
initial = img.flatten(2)[:, :, perm].reshape_as(img)

color_init_met = po.Metamer(
  img, color_ps, loss_function=loss # Use same PS model as above
)
color_init_met.setup(
  initial_image=initial, # Add color init image to Metamer
  optimizer=torch.optim.LBFGS,
  optimizer_kwargs=optimizer_kwargs
)
color_init_met.synthesize(max_iter=MAX_ITER)


##### COLOR MODEL, COLOR INIT, PCA CAP ########

pca_cap = po.process.PCA(img, max_relative_scaling=10)

color_ps_cap = po.models.PortillaSimoncelli(img.shape[-2:], color_statistics=True, transform=pca_cap)
color_cap_met = po.Metamer(img, color_ps_cap, loss_function=loss)
color_cap_met.setup(
  initial_image=initial, # Add color init image to Metamer
  optimizer=torch.optim.LBFGS,
  optimizer_kwargs=optimizer_kwargs
)
color_cap_met.synthesize(max_iter=MAX_ITER)


##### COLOR MODEL, RANDOM INIT, OPC ########

color_ps_opc = po.models.PortillaSimoncelli(img.shape[-2:], color_statistics=True)
color_opc_met = po.Metamer(img, color_ps_opc, loss_function=loss)
color_opc_met.setup(
  optimizer=torch.optim.LBFGS,
  optimizer_kwargs=optimizer_kwargs
)
color_opc_met.synthesize(max_iter=MAX_ITER)


po.plot.imshow(
  [img, color_met.metamer, color_init_met.metamer, color_cap_met.metamer, color_opc_met.metamer],
  title=["Target", "Random Init", "Color Init", "Color Init + PCA Cap", "OPC"],
  as_rgb=True
)
plt.show()
```

bean_metamers

The metamer with random initialization and PCA transform looks very poor (second image). Color-matched initialization greatly improved the outcome for PCA transform (third image). Capping the maximum PCA scaling further improved the metamer, reducing high-frequency noise. The OPC metamer looks pretty good without the need of color-matched initialization.

**Conclusions:**

- The experiments above show that the color model works pretty well.
- Reducing the high-frequency residual seems like an important thing to try next.
- In some examples before I saw PCA perform much better than OPC, but not sure if it was a fluke. Maybe we need to do a bit of experimentation to see if PCA is actually better than OPC or not. This should be relevant to what version of the model we include in the docs as suggest that the users use.

dherrera1911:
Also, in some informal analyses I observed that `MetamerCTF` performed pretty poorly with this color synthesis. I didn't look into this issue further.

billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/482

dherrera1911:
I think that the failing tests are because `transform` defaults to `OPC()`, which is an `nn.Module` that makes the `PortillaSimoncelli` object print as
```python
PortillaSimoncelli(
    (transform): OPC()
    (_pyr): SteerablePyramidFreq()
  )
```
even for the independent-channel model, which messes up a regex call in the tests.

One solution is to change the default of `transform`. We could make it `None` for the case when `color_statistics=False`, and `OPC()` when it is `color_statistics=True`, or some other default.

I thought of making `OPC()` a function instead of an `nn.Module`, but we would lose the torch benefits of easily changing the OPC transform matrix's device and dtype. 

Let me know your thoughts @billbrod

billbrod:
Hey Daniel, thanks for your work on this!

In order to merge the color version of the PS model, we need to demonstrate that it is able to successfully synthesize color textures. The example you show  of the apple in your comment shows that's not currently the case -- the hue of the greens and reds do not match the original image in any of the examples. This matches my recollection of @hmd101's results -- the colors just didn't look right. 

For what I mean by "right", you can look at the examples on [Eero's webiste](https://www.cns.nyu.edu/~lcv/texture/color/), figures 6 or 7 of the Vacher and Briand 2021 paper, or figures 13 and 15 of the [Brown et al. 2023](https://dl.acm.org/doi/10.1145/3564605) paper.

I think, unfortunately, we're going to have to do a similar process to what I did in #365. I think the way to do handle this is:
- Generate some successful color metamers from other implementations. The color matlab implementation is available on [Eero's website](https://www.cns.nyu.edu/~lcv/texture/), unsure if we can find a runnable version of the Vacher and Brown synthesis procedures, but if we can, we should generate some from all the independent procedures.
- For each of the successful metamers and some of our current (unsuccessful) metamers, look at the synthesis error in a per-stat-family manner (similar to what I did in #365) using your current model. That will help us figure out if the problem is "we have the right stats and aren't matching them closely enough" or "we're missing some important stats" (if our unsuccessful metamers have comparable error in all stats to the successful ones, then we're missing something.)
    - In particular, Vasha DuTell (second author on the Brown paper) told me that she remembers color coming together last, so it might just be an optimization issue.
    - I find `plot_representation` very helpful for this, and am happy to take a crack at implementing this in your branch. We would just need to add an additional subplot per stats family and summarize them in a similar way (e.g., l2-norm over each band separately).
- Based on the above results, we'll be able to determine whether we should play around with optimization hyper-parameters or need to think more deeply about the stats themselves. Right now, I believe all three of the existing implementations use a different set of stats (and the original matlab one doesn't add any beyond the initial color-channel mixing and correlations between the RGB channels), so one of the things I will want from this process is to determine which set of stats are actually needed.

We should open up a new issue to track progress on this, similar to what I did in #365. (We could reuse #46, but I think this is specific and different enough that it would benefit from its own location.)

I can find some additional color textures and send them to you.

An additional thing we can try: I believe the matlab implementation performs the PCA transform of the RGB image, optimizes *in that space* and then undoes it. I think it is unlikely that including the transform in the model would cause us any problems (and, similar to our conversation in #460, I think it makes the most sense to include it in the model), but explicitly checking it will allow us to rule it out.

Once we get some successful synthesis, then we can spend more time reviewing the implementation and discussing tests, docs, etc. (I do think the high frequencies you're noting are due to the fact that you're not using the `portilla_simoncelli_loss_factory` -- all that function does is remove the min/max and upweight the variance of the highpass residuals in the loss, and it resulted in a great improvement, which you can see in #365).

PR #483, opened 2026-08-20
Use deterministic algorithms for deep nets

billbrod:

In #476 , realized that the category for the MAD Adversarial Example had changed from "cheeseburger" to "malinois". After poking around a bit, realized this was because I need to call [torch.use_deterministic_algorithms(True)](https://docs.pytorch.org/docs/2.12/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms) because, among other things, convolutions are non-deterministic on GPUs by default. This adds that line to three docs pages: the two adversarial examples ones and the "How to use Deep Nets" one (it is unnecessary for reproducing Feather et al., 2023, since that notebook just loads in cached results, and the test producing those results already has this option).

Originally encountered this problem in #460, should've realized this was an issue earlier. Thus, I've now updated the PR template to include a question about this.

**Describe changes**

- Adds `torch.use_deterministic_algorithms(True)` with comment to top of relevant notebooks, right after setting seed.
- Updates PR template to include a question about reproducibility practices

**Checklist**

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -37,4 +37,5 @@ Affirm that you have done the following:
 - [ ] I have described the changes in this PR, following the template above.
 - [ ] I have added any necessary tests.
 - [ ] I have added any necessary documentation. This includes docstrings, updates to existing files found in `docs/`, or (for large changes) adding new files to the `docs/` folder.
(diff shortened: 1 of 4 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/483

PR #485, opened 2026-08-27
Bugfix: Align `penalties`, `reference_metric_loss`, `optimized_metric_loss` with `losses` for LBFGS

billbrod:

While working on datasaurus example for #428, realized that computing the penalty after synthesis (by calling `torch.func.vmap(met.penalty_function)(met.saved_metamer)`) did not give the same values as `met.penalties` when using `LBFGS` as the optimizer. However, this was not the case for the objective function (`torch.func.vmap(met.objective_function)(met.saved_metamer)` does equal `met.losses`). (I have a test that should've caught this, but the penalties value was so small that the error was within tolerance)

LBFGS (unlike Adam) needs to evaluate the loss function multiple times per synthesis iteration, which we handle by defining `_closure`, which evaluates and returns the loss function. It can only return a single value, and so to avoid re-computing the penalty, we were caching the penalty computed in `_closure` and retrieving it after `optimizer.step(_closure)`.  Turns out, the loss value actually returned by `optimizer.step(_closure)` corresponds to the **first** time it was called on each iteration, whereas our caching procedure meant we were grabbing the penalty value corresponding to the **last**time it was called. The value we cached for the objective function corresponded to the value that can be computed post-hoc from `saved_metamer`, so updated penalty to correspond to that value as well.

The same problem affects `penalties` for MADCompetition, so that is also updated here.

While working on this, realized that the same issue affects `reference_metric_loss` and `optimized_metric_loss` for MADCompetition. They are similarly updated. In so doing, realized they should be updated so they work similarly to `losses` / `penalties`: cache the value corresponding to the current metamer at end of synthesis, and compute that if needed when property `reference_metric_loss` / `optimized_metric_loss` is accessed (which happens if `synthesize` has not been called or was interrupted).

That last change results in a small change to the MADCompetition API (adding `_current_ref_penalty` / `_current_opt_penalty`, created at initialization), so needed to add `FutureWarning` on load for these attributes, similar to how we handled `_current_loss` / `_current_penalty` in previous versions. Updates compatibility docs about this.

None of this affects Adam (the default) or the majority of other optimizers, which only call `_closure` once. I cannot find a list on the torch site of the optimizers which call `_closure` multiple times, only [this note](https://docs.pytorch.org/docs/2.13/optim.html#optimizer-step-closure), which says "Some optimization algorithms such as Conjugate Gradient and LBFGS need to reevaluate the function multiple times". So unclear what other optimizers beyond those two are affected.

**Describe changes**

- Updates how penalty, reference metric loss, optimized metric loss are all updated over synthesis to align behavior with objective function (for Metamer and MADCompetition).
- Adds tests to explicitly check that the post-hoc computation matches the stored values.
- Updated existing test so it would've failed.
- Makes losses, penalties, reference metric loss, and optimized metric loss return tensors in correct dtype (matching that of self.image)
    - This requires making `self.image` an `abstractproperty` of `OptimizedSynthesis` (that is, required for all sub-classes). That's fine for now, may need to update for #434
- Adds FutureWarning to load
- Updates compatibility guide with notes on these changes

**Checklist**


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/485

PR #486, opened 2026-08-31
Updates create_model call to match new syntax for `timm>=1.0.29`

billbrod:

Starting with `timm==1.0.29`, `timm.create_model` raises an error for strings like `"timm/resnet50.tv_in1k"`: 

> Model name 'timm/resnet50.tv_in1k' has no source prefix but looks like a Hugging Face Hub repo id or a local path. Use 'hf-hub:timm/resnet50.tv_in1k' to load from the Hub or 'local-dir:timm/resnet50.tv_in1k' to load from a local folder.

This just adds `hf-hub` as a prefix to each of our invocations of `create_model` to avoid these warnings.

Required for timm>=1.0.29, also works with 1.0.28 (and hopefully older)

Example [action failure](https://github.com/plenoptic-org/plenoptic/actions/runs/33289417483/job/99198297091). 

I can't find a mention of this in the release notes, and [the docs about `create_model`](https://huggingface.co/docs/timm/en/reference/models#timm.create_model.model_name) show passing it strings like `"resnet50.tv_in1k"`, so the problem appears to be the `timm/`, which could be a directory.

**Checklist**

### DIFF
--- a/docs/user_guide/models_and_metrics/deep_nets.md
+++ b/docs/user_guide/models_and_metrics/deep_nets.md
@@ -194,7 +194,7 @@ import plenoptic as po
 
 # since these two models are identical, they have the same layer names
 target_layer = "layer3"
-deepnet = timm.create_model("timm/resnet50.tv_in1k", pretrained=True)
+deepnet = timm.create_model("hf-hub:timm/resnet50.tv_in1k", pretrained=True)
 deepnet.eval()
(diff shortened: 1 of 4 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/486

PR #487, opened 2026-09-02
Updates imshow/animshow docstring to warn about vrange for RGB images

### LINKED ISSUE #479: How to handle plotting RGB difference images
As @raffles-xingqi-zhu was trying to make some plots in #476, he discovered that changing `vrange` for `po.plot.imshow` seemed to have no effect on the colormap of an RGB image. We double-checked [pyplot.imshow](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html) and yep: colormap, vmin, vmax (and other args) are all ignored for RGB(A) images.

We should either raise a warning if someone tries to set `vrange` when `as_rgb=True` or do the remapping for them. In either case, does that behavior belong in plenoptic's imshow or pyrtools?

billbrod:

This updates the docstrings of `imshow` and `animhow` to provide more information about how vrange is handled for RGB images. It relies on https://github.com/LabForComputationalVision/pyrtools/pull/51 (release v1.0.11) to update the titles and raise the warnings, so this is largely about alerting plenoptic users.

Closes #479 

- How is the wording in the docstring? I had a bit of trouble writing something I was happy with.
- How is the example in the doctest? I ended up adding several sections -- are they clear? Are they all necessary?

**Describe changes**

- Adds admonition to vrange parameter
- Updates as_rgb parameter description as well
- Adds longish doctest trying to show how this works.
- Adds info about UserWarning (raised by pyrtools) that happens when vrange is changed
- Increases min pyrtools version so that it will get triggered

**Checklist**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ keywords = ['neuroscience', 'pytorch', 'visual information processing', 'machine
 dependencies = [
     "numpy>=1.1",
     "torch>=1.8,!=1.12.0",
-    "pyrtools>=1.0.9",
+    "pyrtools>=1.0.11",
     "scipy>=1.0",
     "matplotlib>=3.3",
     "tqdm>=4.29",
--- a/src/plenoptic/plot/display.py
+++ b/src/plenoptic/plot/display.py
@@ -386,7 +386,7 @@ def imshow(
       ...     as_rgb=True,
       ...     zoom=0.5,
       ... )
-      <PyrFigure size ... with 1 Axes>
+      <PyrFigure size ... with 2 Axes>
       >>> po.plot.imshow(
       ...     [color_wheel, 2 * color_wheel],
       ...     title="vrange=auto1",
@@ -394,7 +394,7 @@ def imshow(
       ...     as_rgb=True,
       ...     zoom=0.5,
       ... )
-      <PyrFigure size ... with 1 Axes>
+      <PyrFigure size ... with 2 Axes>
 
     Notice that:
 
@@ -413,11 +413,11 @@ def imshow(
 
       >>> gray_whl = color_wheel.mean(1, keepdim=True)
       >>> po.plot.imshow([gray_whl, 2 * gray_whl], title="vrange=indep1", zoom=0.5)
-      <PyrFigure size ... with 1 Axes>
+      <PyrFigure size ... with 2 Axes>
       >>> po.plot.imshow(
       ...     [gray_whl, 2 * gray_whl], title="vrange=auto1", vrange="auto1", zoom=0.5
       ... )
-      <PyrFigure size ... with 1 Axes>
+      <PyrFigure size ... with 2 Axes>
 
     To avoid this problem, one can rescale the input before plotting using a function
     like :func:`plenoptic.process.rescale`:
@@ -432,7 +432,7 @@ def imshow(
       >>> imgs += [po.process.rescale(im) for im in imgs]
       >>> titles += ["rescaled", "rescaled"]
       >>> po.plot.imshow(imgs, as_rgb=True, zoom=0.5, col_wrap=2, title=titles)
-      <PyrFigure size ... with 1 Axes>
+      <PyrFigure size ... with 4 Axes>
 
     If ``as_rgb=False``, images with multiple channels will have each channel plotted
     as a separate grayscale image:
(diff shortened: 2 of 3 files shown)


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/487

sjvenditto (approved):
> How is the wording in the docstring? I had a bit of trouble writing something I was happy with.

I think the wording in the admonition is clear, however I would also just point to the specific marplotlib arguments in the parenthetical statement as well since it's slightly different from plenoptic, e.g. "(see `vmin, vmax` in [matplotlib.pyplot.imshow](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html#matplotlib.pyplot.imshow))."

> How is the example in the doctest? I ended up adding several sections -- are they clear? Are they all necessary?

I like the examples, especially since you give an example on how to rescale RGB outside of the function. One piece that could be removed is the comparison of the RGB wheel to the grayscale counterpart, i.e. the code block after the text "Compare this to the results of the same syntax with grayscale images:", mostly because it's redundant with the grayscale examples that come before

These comments are pretty minor though, otherwise I think this is good to go!

raffles-xingqi-zhu:
I find both the admonition and examples overall clear. Here are my thoughts: 

I agree with Sarah that the greyscale counterpart example is redundant. Moreover, it disrupts the flow because by the time I got to "To avoid this problem...", I already forgot what the problem is, which itself could use a bit more clarification (e.g. something like "to help visualize 3-channel images that aren't (defined) in the range of [0,1] or [0,255] as RGB images". 

It was also curious to me in the greyscale counterpart code block how `gray_whl = color_wheel.mean(1, keepdim=True)` led to the averaged colorwheel having vmin=0.1. I would've expected vmin=0. 

I want to bring up something separate that I noticed while reading the doc for `vrange`. Are we sure this statement for "indep0" is correct (and similar statement for "auto0")? 

"indep0": Each image has an independent vmin/vmax, which have the same absolute value, which comes from either their minimum or maximum value, whichever has the larger absolute value.

This statement makes it sound like vmin and max have the same absolute value, which is clearly not the case. Wouldn't it make more sense to say something like: 

"indep0": Each image has an independent vmin/vmax, which come from the minimum or maximum values of the image.

billbrod:
Thanks both! I have pushed responses to your comments.

> It was also curious to me in the greyscale counterpart code block how `gray_whl = color_wheel.mean(1, keepdim=True)` led to the averaged colorwheel having vmin=0.1. I would've expected vmin=0.

The values are bounded below by 0, so for `gray_whl` to have a minimum of 0, then each channel would need to have a minimum of exactly zero -- that's not what we have: `color_wheel.min(-1).values.min(-1)` (to take the minimum across the height and width) gives `[0.0039, 0.0039, 0]`. Put another way, we don't have any actual true black in the image.

Also I realized as I was doing this -- this isn't the proper way to convert an RGB to a grayscale image (see [here](https://scikit-image.org/docs/stable/auto_examples/color_exposure/plot_rgb_to_gray.html)), which might confuse people, so maybe it's for the best not to show this example anyway.

> I want to bring up something separate that I noticed while reading the doc for `vrange`. Are we sure this statement for "indep0" is correct (and similar statement for "auto0")?
> 
> "indep0": Each image has an independent vmin/vmax, which have the same absolute value, which comes from either their minimum or maximum value, whichever has the larger absolute value.
> 
> This statement makes it sound like vmin and max have the same absolute value, which is clearly not the case. Wouldn't it make more sense to say something like:
> 
> "indep0": Each image has an independent vmin/vmax, which come from the minimum or maximum values of the image.

I think you're confused because `indep0` is not the default -- `indep1` is. If the user specifically sets `vrange="indep0"`, then it has the documented behavior:

```python
import plenoptic as po
img = po.data.color_wheel()
po.plot.imshow(img, vrange="indep0", as_rgb=False)
```

tmp

I don't actually have any examples of `indep0` / `auto0` in the docstring. Would it be helpful to include one?

raffles-xingqi-zhu:
You are absolutely right about me misreading the argument. In my head I assumed the first argument is the default and got confused. I don't think it's necessary to add the example as it is self-explanatory in my opinion. Changing the order could be helpful so the default becomes the first argument to be introduced but this is not necessary.  The PR otherwise looks good and is ready to be merged!

PR #491, opened 2026-09-18
Adds doctests to Eigendistortion methods

billbrod:

Adds doctests to Eigendistortion object, mirroring those of Metamer. See [API docs](https://docs.plenoptic.org/docs/pulls/491/api/generated/plenoptic.Eigendistortion.html#plenoptic.Eigendistortion)

Makes progress on #237. Also opened #490 to track changes I want to make to Eigendistortion.

**Describe changes**

- Adds doctests to Eigendistortion public methods and the jacobian attribute
- Removes unnecessary lines from some Metamer and MADCompetition lines
- Removes unnecessary prints from doctests. only ones that still have print are those that are printing an f-formatted string (for `to_numpy`)
- Adds CPU version of example_eigendistortion.pt, which we use in most of the doctests.

**Checklist**

- [x] If I noticed any typos (my own or others), I have updated `plenoptic-dict.txt` to include those typos and suggested fixes.


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/491

src/plenoptic/_synthesize/eigendistortion.py
```diff
@@ -163,9 +182,76 @@ def synthesize(
         Warns
         -----
         UserWarning
-            If ``method == "power"`` but the Jacobian size is greater than 1e6 (which
+            If ``method == "exact"`` but the Jacobian size is greater than 1e6 (which
```
BalzaniEdoardo: I think you should have the word raises here

src/plenoptic/_synthesize/eigendistortion.py
```diff
@@ -163,9 +182,76 @@ def synthesize(
...
+        >>> eig.eigenvalues
+        tensor([0.9320, 0.1514])
+
+        Set ``method`` to use different algorithms for finding the eigendistortions.
+        Note that, while ``"power"`` synthesizes the top and bottom ``k``
+        eigendistortions, ``"randomized_svd"`` only synthesizes the top ``k``.
+
+        >>> eig = po.Eigendistortion(img, model)
+        >>> # randomized_svd doesn't take a max_iter argument
+        >>> eig.synthesize("randomized_svd")
```
BalzaniEdoardo: is the `max_iter` not needed here?

src/plenoptic/_synthesize/eigendistortion.py
```diff
@@ -163,9 +182,76 @@ def synthesize(
...
+        >>> # eigenvalue of top eigendistortion
+        >>> eig.eigenvalues
+        tensor(0.8777)
```
BalzaniEdoardo: is it expected that the top value is  so different across methods? is that due to the fact that the synthesis run for a different duration?

src/plenoptic/_synthesize/eigendistortion.py
```diff
@@ -163,9 +182,76 @@ def synthesize(
...
+
+        And ``"exact"`` explicitly computes the Jacobian, retrieving all
+        eigendistortions, and thus uses a lot of memory, and so is only recommended
+        for small images or models.
```
BalzaniEdoardo: Is that due to the size of the jacobian itself, or for the checkpointing in the forward pass?

If the checkpoints are the main issue, in jax, there is a way to not save intermediates in the forward pass, that makes the backward slower, but it doesn't blow up the memory. 

https://docs.jax.dev/en/latest/jep/11830-new-remat-checkpoint.html

there is probably an equivalent in torch

src/plenoptic/_synthesize/eigendistortion.py
```diff
@@ -642,15 +756,85 @@ def load(
...
+        >>> img = po.data.einstein().to(torch.float64)
+        >>> lg = po.models.LuminanceGainControl(
+        ...     (31, 31), pad_mode="circular", pretrained=True, cache_filt=True
+        ... ).eval()
+        >>> lg.to(torch.float64)
+        LuminanceGainControl(...)
+        >>> po.remove_grad(lg)
+        >>> eig = po.Eigendistortion(img, lg)
+        >>> eig.eigendistortions
+        >>> eig.load(po.data.fetch_data("example_eigendistortion.pt"))
```
BalzaniEdoardo: is the data in the loaded example required to be float64 if lg is float64? how does this works?

src/plenoptic/_synthesize/eigendistortion.py
```diff
@@ -642,15 +756,85 @@ def load(
...
+
+        If the loading ``Eigendistortion`` object was not initialized with same values
+        as the saved object, an error will be raised:
+
+        >>> eig = po.Eigendistortion(torch.rand_like(img), lg)
+        >>> eig.load(po.data.fetch_data("example_eigendistortion.pt"))
+        Traceback (most recent call last):
+        ValueError: Saved and initialized attribute image have different values...
+
+        If the loading ``Eigendistortion`` object has a different data type than the
```
BalzaniEdoardo: ok, that is the answer to my previous question

BalzaniEdoardo (approved):
This looks good,  I have a few questions here and there. 

In the load example, I would mention that the dtype as to match right away, and then write: see below.

src/plenoptic/_synthesize/eigendistortion.py
billbrod: Size of the Jacobian itself, which is the number of elements in the model output by the number of elements in the image. So there's not a way around it.

src/plenoptic/_synthesize/eigendistortion.py
billbrod: For `"power"`, it's probably a sign that it needs to be run for more iterations. I'm not sure about randomized SVD, I can play around with it a bit to see if changing the relevant parameters changes its outcome.

PR #492, opened 2026-09-22
Adds Datasaurus dozen example to demonstrate penalty usage

billbrod:

This PR adds a new series of notebooks discussing the [datasaurus dozen](https://en.wikipedia.org/wiki/Datasaurus_dozen) and showing how to synthesize a similar dataset with plenoptic by using Metamer with some custom penalty functions.

The motivation for this is that I realized the Datasaurus dozen are all metamers -- they are sets of 142 x,y points with different values but identical summary statistics. This is useful because it's a nice non-perceptual example of the concept (and some people will already be familiar with it) and also because synthesizing similar examples requires a wide variety of penalty functions. So this hopefully serves as an example for what can be done with penalty functions.

Makes progress on #428.

- How are the docs, are they understandable?
- I tried to have the landing page give an overview of the dataset and how we approach it, and then the sub-pages are all relatively small (and very similar), focusing on the penalty function and describing what comes up. Does that work well? The goal of the sub-pages is to explain the penalty function (and thus show many examples of how they can be used in plenoptic) -- do they accomplish that goal?
- how is the discussion of the model's redundant stats?
- Currently, there's a decent amount of duplicated code in each of the sub-notebooks for the model and animating code. If I want them to be downloadable and independently runnable, not sure I can avoid that (without including them in package, which I don't want to do). I could've created the pages using some sort of template, but by the time I realized that it was too far along. That's fine, right?
- I cannot find any source for the star equation other than the linked stack overflow post. I think that's fine, but is frustrating.

**Describe changes**

- Change `show_nav_level` to 1, so that we don't automatically expand the headers for this in the user guide. In an ideal world, I'd have that be the case only for the datasaurus notebooks (since there's so many of them) but the theme doesn't let that option be set in a page-specific manner
- Adds new landing page describing the dataset and giving an overview, then a separate small page per example.
- Adds synthesis to `test_uploaded_files.py`
- Adds input and output to fetch

**Checklist**

- [x] If I noticed any typos (my own or others), I have updated `plenoptic-dict.txt` to include those typos and suggested fixes.


billbrod:
Documentation built by flatiron-jenkins at http://docs.plenoptic.org/docs//pulls/492

